PR #1, opened 2019-11-21
Change development status from Alpha to Production/Stable

zhujun98:
Should we change the development status?


takluyver:
:+1: thanks

PR #5, opened 2019-12-16
run info duration

### LINKED ISSUE #4: Spurious precision in info 'Duration'
I've just noticed that the output of `run.info()` includes the duration like this:

```
Duration:       0:04:02.500000
```

As we're only calculating this to 0.1 seconds, we shouldn't show it with microsecond precision. Maybe we should also hide the hours for runs that last less than 1 hour.

tmichela:
Show run duration to time precision (100ms). Fix time delta calculation.

```bash
In [3]: run.info()                                                              
# of trains:    5738
Duration:       0:09:33.8
First train ID: 296112239
Last train ID:  296117976
```

closes #4

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1047,8 +1047,8 @@ class DataCollection:
         first_train = self.train_ids[0]
         last_train = self.train_ids[-1]
         train_count = len(self.train_ids)
-        span_sec = (last_train - first_train) / 10
-        span_txt = str(datetime.timedelta(seconds=span_sec))
+        seconds, deciseconds = divmod((last_train - first_train + 1), 10)
+        span_txt = f'{datetime.timedelta(seconds=seconds)}.{int(deciseconds)}'
 
         detector_modules = {}
         for source in self.detector_sources:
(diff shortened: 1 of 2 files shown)


takluyver:
Someone mentioned that nbval appears to be incompatible with the recently released coverage v5: https://github.com/computationalmodelling/nbval/issues/129

takluyver:
Other than that, though, I think this change looks good.

takluyver:
Do you want to pin coverage <5, as I have in https://github.com/European-XFEL/EXtra-geom/pull/4 , to avoid this issue?

tmichela:
arf, do we still want to support Python `3.5`? it does not like f-strings :'D

takluyver:
I think it's probably safe to require Python 3.6 now. The main (DESY) Anaconda installation on Maxwell is 3.6. Unless you know of any environment where people might still be stuck on 3.5?

tmichela:
> you know of any environment where people might still be stuck on 3.5?

Not that I know...

I changed to the old formatting style for now, I suppose we should mention it to the group before dropping support...

takluyver:
OK, thanks

Issue #3, opened 2019-12-12
Find new data for Dask example


takluyver:
I think the DSSC data we're using in the Dask example notebook has been migrated to dCache, and now reading it is slow. We should use some XMPL data which will stay on GPFS.

PR #7, opened 2019-12-17
Use XMPL data for Dask example notebook

takluyver:
This data should stay on GPFS, so it should be relatively quick to access.

Closes gh-3


tmichela:
LGTM

Issue #2, opened 2019-12-11
Open File Limit Error


RobertRosca:
Open file limit per process may be hit for very large runs, leading to this error:

```
✘  roscar@max-exfl015 /gpfs/exfel/exp/MID/201931/p900091/raw/r0491  extra-data-validate .
Checking run directory: .
Traceback (most recent call last):
 File "/software/anaconda3/5.2/bin/extra-data-validate", line 10, in <module>
 File "/software/anaconda3/5.2/lib/python3.6/site-packages/extra_data/validation.py", line 222, in main
 File "/software/anaconda3/5.2/lib/python3.6/site-packages/extra_data/validation.py", line 163, in validate
 File "/software/anaconda3/5.2/lib/python3.6/site-packages/extra_data/validation.py", line 171, in run_checks
 File "/software/anaconda3/5.2/lib/python3.6/site-packages/extra_data/validation.py", line 191, in check_files_map
 File "/software/anaconda3/5.2/lib/python3.6/site-packages/extra_data/run_files_map.py", line 52, in __init__
 File "/software/anaconda3/5.2/lib/python3.6/site-packages/extra_data/run_files_map.py", line 86, in load
OSError: [Errno 24] Too many open files: './karabo_data_map.json
```

https://github.com/European-XFEL/EXtra-data/blob/e1f2b48e39f4627484bebc02045f554823f663ee/extra_data/run_files_map.py#L84-L90

Files may not be getting closed after they are looked at.

takluyver:
Making a note for myself here: this is hopefully easy enough to fix for validate, but I want to check what happens with accessing the HDF5 files themselves - does HDF5 do anything to work around the issue, or do we need to be smarter with opening and closing files?

tmichela:
That's strange. How the cache file was generated in the first place? (I tried to use `open_run` without using the cache and ran into the same issue...)

tmichela:
got it: @ebadkamil just told me that he opened in the past this run per sequence (so not the full run). I guess then the cache file is used and did not run into this issue until running validate...

takluyver:
HDF5 does not appear to do anything smart with this, so we'll have to work around it ourselves.

Do we have any examples of runs with that many files which are still on GPFS? The raw data from that MID proposal was migrated to dCache already, so testing with it can be extra slow.

tmichela:
do we close this?

egorsobolev:
The issue is still unsolved.
[...]

PR #8, opened 2019-12-17
Crude mitigation for hitting file limit

takluyver:
This is a simple way to address the most likely cause of #2. I was able to `lsxfel` and `extra-data-validate` the same run mentioned there with these changes.

The mitigation is to close the files after getting metadata from them, but leave them open if we want anything else from them. I don't consider it a fix, but as we've only hit this once so far, and you rarely want to access every file in a run, it will probably avoid the problem for quite a while.

I also found that 1024 is the 'soft' limit on open files - a process can raise it up to the 'hard' limit (configured at 4096 on Maxwell) without any special privileges. From Python, the [resource module](https://docs.python.org/3/library/resource.html) provides the necessary API for this. So this is another way to dodge the problem further.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -74,6 +74,7 @@ class FileAccess:
     """
     _file = None
     _format_version = None
+    metadata_fstat = None
 
     def __init__(self, filename, _cache_info=None):
         self.filename = filename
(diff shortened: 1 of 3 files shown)


RobertRosca:
LGTM, tested with run that has over 1024 files.

PR #10, opened 2020-01-08
Reorganise documentation

takluyver:
This prioritises the examples over the reference docs, and prioritises the simpler 'load this data into memory' interfaces over the interfaces for dealing with larger data. This is in line with how I now tend to present EXtra-data.

The old 'Demo' notebook is gone, with the content that it covered split up over several places (some existing content, and some new examples). As EXtra-data has grown, showing all the functionality in one example notebook makes less sense.


takluyver:
Most of the content there hasn't changed - it's still essentially the same examples, just in a different arrangement.

philsmt:
LGTM.

A small details I wondered though is whether we should include a small, more explicit warning to iterating trains directly. While it is conceptually very helpful for the first steps in looking at train data, it is hardly useful in all but a few scenarios of actual use. I feel a bit like promoting a highly inefficient anti-pattern in Python, which in particular novice programmers might fall prey to.

As a matter of fact I'm speaking from personal experience, as both SQS staff and users approached me to fix code that "never finishes running". Iterating directly, doing math with Python primitives per train and aggregating the result in dicts and lists was often the reason, even for medium-sized runs (< 1h). Now, we can hardly teach proper use of numpy and EXtra-data at the same time, but this interface seems more susceptible than, e.g., get_array/get_dataframe/etc. If total memory is a concern, one will need to use a different approaches anyway (like dask).

takluyver:
Part of this reorganisation is already to present things like `get_array` and `get_dataframe` as the first port of call, and iteration as a fallback option for larger data. But the iteration approach has one big advantage: it lets you easily use similar code for online processing with Karabo bridge and offline processing from files.

I don't think the overhead of Python loops itself is a problem for iterating over trains - even though Python loops are 'slow', they're not that slow. 12 hours continuous recording would net you 432k trains, and the overhead of a loop running that many times is about 10ms on my laptop. (Obviously if you iterate through pixels or something, that's another story, but we're not encouraging that.)

Where I think it does get slow is reading a chunk at a time from HDF5 files. I've been doing some work in h5py which should significantly reduce the overhead on typical reads. Possibly we can also be smarter in EXtra-data, e.g. caching references to open datasets, or pre-emptively reading smaller data into memory. Another approach we could take is to reduce the overhead by letting people iterate over larger chunks of data, rather than strictly one train per step.

I really like Dask, but I think in many cases it's much easier to think about code that visibly iterates over the data to do something than it is about code which relies on a whole lot of cleverness behind the scenes. I've also seen cases where Dask has run out of memory and ground to a halt - from what I understand (from e.g. [this issue](https://github.com/dask/distributed/issues/2602)), the scheduling doesn't know which tasks will reduce the data size, so it can't arrange things to release memory before using more. So I'm reluctant to go all-in on that approach.

philsmt:
I did not mean to imply a technical reasoning, either in EXtra-data nor Python or h5py. It is an educational issue, which I encountered a lot with scientists attempting their first steps writing analysis code. I meant to add as many cues as possibly to point into the right direction. As much we'd hope a good written documentation will be well read, understood and applied, this has not been my experience in the "field" too often.

I think it is vital to provide users of any expertise level with tools to write code as quickly as possible (see psana). We are still in this weird transition phase, where writing software is not completely a natural part of science. Hence we're also part of that educational pipeline to get there.

takluyver:
Sure, that makes sense. But I don't think we need to discourage iterating over trains in general. Iterating over individual pixels, for instance, is definitely an anti-pattern, but I think iterating over trains is a reasonable approach in many cases.

In any case, I believe this PR downplays iteration compared to the current state of the docs, by putting the 'load it into memory' options first in the examples. So if you're otherwise happy with it, I propose we merge it, and continue to discuss whether the docs should further discourage iteration.

fangohr:
Thanks for reworking this and for the discussion.

Issue #15, opened 2020-02-06
Missing method `squeeze` for StackView


zhujun98:
For raw modular detector data (`image.data`) the shape is (memory cell, 1, modules, y, x), which has an extra axis. If `stack_detector_modules` returns a true numpy array, `np.squeeze` will be needed (at least in `EXtra-foam`) to remove the extra axis. So we need to add a `squeeze` method for StackView, otherwise the code will fail at shape checking when a pre-allocated output array is used. 

As far as I remember, In the early version, `karabo_data` can deal with raw and calibrated data at the same time.

takluyver:
Makes sense, and it turns out to be very easy (unless I've missed something). Have a look at #16.

PR #16, opened 2020-02-06
Add StackView.squeeze() method, like numpy.squeeze()

takluyver:
Closes gh-15

### DIFF
--- a/extra_data/stacking.py
+++ b/extra_data/stacking.py
@@ -220,9 +220,15 @@ class StackView:
             slices = [slice(None, None)] * self.ndim
 
             for ax in axis:
+                try:
+                    slices[ax] = 0
+                except IndexError:
+                    raise np.AxisError(
+                        "axis {} is out of bounds for array of dimension {}"
+                        .format(ax, self.ndim)
+                    )
                 if self.shape[ax] != 1:
                     raise ValueError("cannot squeeze out an axis with size != 1")
-                slices[ax] = 0
         else:
             raise TypeError("axis={!r} not supported".format(axis))
 
--- a/extra_data/stacking.py
+++ b/extra_data/stacking.py
@@ -208,3 +208,8 @@ class StackView:
         for modno, data in self._data.items():
             arr[modno] = data
         return np.moveaxis(arr, 0, self._stack_axis)
+
+    def squeeze(self):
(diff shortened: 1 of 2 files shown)


extra_data/stacking.py
```diff
@@ -208,3 +208,8 @@ def asarray(self):
         for modno, data in self._data.items():
             arr[modno] = data
         return np.moveaxis(arr, 0, self._stack_axis)
+
+    def squeeze(self):
+        """Drop any length-1 dimensions - see numpy.squeeze()"""
+        slices = [0 if d == 1 else slice(None, None) for d in self.shape]
```
zhujun98: Thanks for the quick response! Indeed there is a corner case, in which the data contains only one pulse. I tried to write
a mirror implementation of `np.squeeze` and it seems there are something more to consider. The following is my unfinished code

```py
from numpy import AxisError

def squeeze(self, axis=None):
    if axis is None:
         slices = [0 if d == 1 else slice(None, None) for d in self.shape] 
     
     if not isinstance(axis, int):
         raise TypeError("...")

     try:
         if self.shape[axis] != 1:
             raise ValueError(cannot select an axis to squeeze out which has size not equal to one)
     except IndexError as e:
          raise AxisError(str(e))

     ...
     # take care of negative axis     

     # fix the shape and ndim
```

takluyver: Oh, I'd missed that squeeze takes an axis. If I know which dim I want to drop, I tend to do it like `arr[:, 0]` . But it should be easy to add.

extra_data/stacking.py
```diff
@@ -208,3 +208,22 @@ def asarray(self):
...
+        if axis is None:
+            slices = [0 if d == 1 else slice(None, None) for d in self.shape]
+        elif isinstance(axis, (int, tuple)):
+            if isinstance(axis, int):
+                axis = (axis,)
+
+            slices = [slice(None, None)] * self.ndim
+
+            for ax in axis:
+                if self.shape[ax] != 1:
```
zhujun98: Great! LGTM except a nitpick, for out-of-bound axis this code raise `IndexError` while `np.squeeze` raises `AxisError`.

takluyver: True, should be fixed now.

zhujun98: Thank you!

Issue #13, opened 2020-01-31
Method to get single dask array for entire detector


takluyver:
This should be more convenient, rather than requiring the user to get each module manually and align them.

I might also experiment again with using Dask in xarray, to preserve the labels on the data.

PR #18, opened 2020-02-07
get_dask_array across detector modules

takluyver:
This enables working with Dask like this:

```python
from extra_data.components import AGIPD1M
agipd = AGIPD1M(run, min_modules=16)
arr = agipd.get_dask_array('image.data')
avg = arr.unstack('train_pulse').mean(dim='trainId', dtype=np.float32)
avg.compute()
```

See #13.

In theory it should be possible to leave out `min_modules=16` and rely on xarray's missing value handling. But when I tried this, I got a `KilledWorker` error from Dask. Including only trains where all modules have data seems to work around this.

Unfortunately, the warnings I saw when experimenting with Dask & xarray before are back when I try this, though they don't seem to stop it working:

```
distributed.utils_perf - WARNING - full garbage collections took 52% CPU time recently (threshold: 10%)
```

There are a few API differences between `agipd.get_dask_array()` and `agipd.get_array()` (apart from the obvious one of the return type):

- `get_array()` splits trains and pulses into two separate dimensions, whereas `get_dask_array()` leaves them in a single `train_pulse` dimension, with a multi-level index. The latter is more natural if we one day have different numbers of pulses stored per train.
- The index levels in `get_dask_array()` are called `trainId` and `pulseId`, whereas `get_array()` calls them `train` and `pulse`. 'trainId' is consistent with the labels `run.get_array()` uses for a single source.
- `get_array()` tries to guess labels for the other dimensions, and automatically drops the spurious extra dimension (the one that was meant to be for the train builder) if it's present and length 1. `get_dask_array()` currently doesn't try to do any of this.

On the first two, I guess that not many people use `detector.get_array()` - it implies loading a lot of data into memory, and we've never advertised it widely. So I aimed for consistency with the `run.get_array()` interface instead.

I'm still unsure about the last one - making educated guesses about the axes. It feels wrong to be guessing, but on the other hand, these formats are pretty consistent, and we can make common use patterns more convenient with a few small assumptions.

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -330,6 +330,45 @@ class MPxDetectorBase:
 
         return xarray.concat(arrays, pd.Index(modnos, name='module'))
 
+    def get_dask_array(self, key):
+        """Get a labelled Dask array of detector data
+
+        Dask does lazy, parallelised computing, and can work with large data
+        volumes. This method doesn't immediately load the data: that only
+        happens once you trigger a computation.
+
+        Parameters
+        ----------
+
+        key: str
+          The data to get, e.g. 'image.data' for pixel values.
+        """
+        arrays = []
+        modnos = []
+        for modno, source in sorted(self.modno_to_source.items()):
+            modnos.append(modno)
+            mod_arr = self.data.get_dask_array(source, key, labelled=True)
+
+            # At present, all the per-pulse data is stored in the 'image' key.
+            # If that changes, this check will need to change as well.
+            if key.startswith('image.'):
+                # Add pulse IDs to create multi-level index
+                pulse_id = self.data.get_array(source, 'image.pulseId')
+                # Raw files have a spurious extra dimension
+                if pulse_id.ndim >= 2 and pulse_id.shape[1] == 1:
+                    pulse_id = pulse_id[:, 0]
+
+                mod_arr = mod_arr.rename({'trainId': 'train_pulse'})
+
+                mod_arr.coords['train_pulse'] = pd.MultiIndex.from_arrays(
+                    [mod_arr.coords['train_pulse'], pulse_id],
+                    names=['trainId', 'pulseId']
+                )
+
+            arrays.append(mod_arr)
+
+        return xarray.concat(arrays, pd.Index(modnos, name='module'))
+
     def trains(self, pulses=by_index[:]):
         """Iterate over trains for detector data.
 
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -330,7 +330,7 @@ class MPxDetectorBase:
 
         return xarray.concat(arrays, pd.Index(modnos, name='module'))
 
-    def get_dask_array(self, key):
+    def get_dask_array(self, key, subtrain_index='pulseId'):
         """Get a labelled Dask array of detector data
 
         Dask does lazy, parallelised computing, and can work with large data
@@ -342,7 +342,15 @@ class MPxDetectorBase:
 
         key: str
           The data to get, e.g. 'image.data' for pixel values.
+        subtrain_index: str
+          Specify 'pulseId' (default) or 'cellId' to label the frames recorded
+          within each train. Pulse ID should allow this data to be matched with
+          other devices, but depends on how the detector was manually configured
+          when the data was taken. Cell ID refers to the memory cell used for
+          that frame in the detector hardware.
         """
+        if subtrain_index not in {'pulseId', 'cellId'}:
+            raise ValueError("subtrain_index must be 'pulseId' or 'cellId'")
         arrays = []
         modnos = []
         for modno, source in sorted(self.modno_to_source.items()):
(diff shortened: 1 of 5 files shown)


takluyver:
Upgrading xarray, dask and distributed to the latest versions (respectively 0.15.0, 2.10.1 and 2.10.0) has got rid of the warnings when using `min_modules=16`. I still get both warnings and the `KilledWorker` error without that. Looking at the graph, it looks like one task in the graph depends on a massive number of prior tasks - I guess this is somehow caused by xarray bringing the individual module arrays together and reindexing them.

extra_data/components.py
```diff
@@ -330,6 +330,45 @@ def get_array(self, key, pulses=by_index[:]):
...
+        modnos = []
+        for modno, source in sorted(self.modno_to_source.items()):
+            modnos.append(modno)
+            mod_arr = self.data.get_dask_array(source, key, labelled=True)
+
+            # At present, all the per-pulse data is stored in the 'image' key.
+            # If that changes, this check will need to change as well.
+            if key.startswith('image.'):
+                # Add pulse IDs to create multi-level index
+                pulse_id = self.data.get_array(source, 'image.pulseId')
```
tmichela: Hum, that's again tricky here :)

Usual message from Detector people is that the `pulseId` from the detector data is not a reliable information.

First, is relies on the fact that the configuration is correct (machine rate information relatively to the acquisition rate), which is not automatically set (human action).

Then, (I take AGIPD as example, because I know that one best): the first memory cell isn't working properly, and is then not used. The consequence is that it is configured to effectively store the first xray pulse to the second memory cell. Of course the `pulseId` information will not reflect that. So, let's say that the machine is operating at 4MHz and the detector at 1MHz: the actual first pulse will be labeled `4` instead of `0` (could be even worse if the configuration as mentioned in the first point is wrong).

Could we use `cellId` instead of `pulseId` here? It does not say anything with pulse correlation, but at least the information is correct. Right now, cell ids are always increasing from 0 to n for a train. I future it might have any kind of pattern I believe.

takluyver: Hmm, that's a pain.

IIRC, Steffen told me in the past that cell ID was just about detector hardware, and shouldn't be used to identify frames for scientific purposes. Pulse ID is the one that's *supposed* to be scientifically meaningful, even if there are operational problems with it.

If pulse ID is consistently wrong across a run, I think I favour 1 - i.e. if averaging across all frames with pulseId=4 makes sense, even though that's not actually pulse 4. Making it more visible would create pressure to make it more correct. Or we could let the user specify whether they want pulseId or cellId.

takluyver:
I've added a `subtrain_index` parameter, with which you can specify to use `pulseId` or `cellId` to label frames within each train. @tmichela pointed out that in some situations, such as detector characterisation, you do actually care about the cell IDs.

tmichela:
one last point: Unstacking the `train_pulse` dimension to the last axes:
- stacked
```python
<xarray.DataArray 'concatenate-bb368be6f6f06d0bf226beafab93fcc8' (module: 16, train_pulse: 118250, dim_0: 2, dim_1: 512, dim_2: 128)>
dask.array<shape=(16, 118250, 2, 512, 128), dtype=float64, chunksize=(1, 6718, 2, 512, 128)>
Coordinates:
  * train_pulse  (train_pulse) MultiIndex
  - trainId      (train_pulse) int64 507399279 507399279 ... 507399279 507399279
  - cellId       (train_pulse) int64 0 1 2 3 4 5 6 7 ... 22 23 24 25 26 27 28 29
  * module       (module) int64 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15
Dimensions without coordinates: dim_0, dim_1, dim_2
```

- unstacked
```python
<xarray.DataArray 'concatenate-bb368be6f6f06d0bf226beafab93fcc8' (module: 16, dim_0: 2, dim_1: 512, dim_2: 128, trainId: 473, cellId: 250)>
dask.array<shape=(16, 2, 512, 128, 473, 250), dtype=float64, chunksize=(1, 2, 512, 128, 26, 250)>
Coordinates:
  * module   (module) int64 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15
  * trainId  (trainId) int64 507399279 507399280 ... 507399750 507399751
  * cellId   (cellId) int64 0 1 2 3 4 5 6 7 ... 242 243 244 245 246 247 248 249
Dimensions without coordinates: dim_0, dim_1, dim_2
```

takluyver:
I know, that is kind of annoying. The [xarray docs](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.unstack.html) do appear to say it will happen:

> New dimensions will be added at the end.

In `.get_array()`, I unstack and then rearrange the dimensions to the 'correct' order:

https://github.com/European-XFEL/EXtra-data/blob/204fcbe86ae54223366230976c602a1f94970ecf/extra_data/components.py#L37-L40

That's inelegant, but I don't know of a better way. I'll open an xarray issue to ask about it - but I don't think it should hold this up.

tmichela:
Ok then, LGTM!

PR #20, opened 2020-02-19
Start writing release notes for next version

takluyver:
Tentatively calling it 0.9, but it could equally be 1.0.


takluyver:
@tmichela was in favour of calling the next version 1.0, so I've updated the header in these notes.

PR #23, opened 2020-02-24
Unit test for validation code

tmichela:
- add unit test for validation
- fix on validation: catch `first` / `count` mismatch (that probably never happened...)

### DIFF
--- a/extra_data/validation.py
+++ b/extra_data/validation.py
@@ -70,6 +70,21 @@ class FileValidator:
                     dataset=ds_path,
                 )
 
+    def _get_index(self, source, group):
+        """returns first and count dataset for specified source.
+        
+        This is slightly different to the same method in FileAccess as it does
+        cut the dataset up to the trainId's dataset length.
+        """
+        ix_group = self.file.file['/INDEX/{}/{}'.format(source, group)]
+        firsts = ix_group['first'][:]
+        if 'count' in ix_group:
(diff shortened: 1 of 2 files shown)


extra_data/validation.py
```diff
@@ -70,6 +70,21 @@ def check_trainids(self):
...
+        cut the dataset up to the trainId's dataset length.
+        """
+        ix_group = self.file.file['/INDEX/{}/{}'.format(source, group)]
+        firsts = ix_group['first'][:]
+        if 'count' in ix_group:
+            counts = ix_group['count'][:]
+        else:
+            status = ix_group['status'][:]
+            counts = np.uint64((ix_group['last'][:] - firsts + 1) * status)
+        return firsts, counts
```
tmichela: Copy of `FileAccess._get_index` but returning the full dataset not sliced by trainIds length (required to compare `count` and `first`)

extra_data/validation.py
```diff
@@ -106,6 +121,7 @@ def check_indices(self):
                         first_shape=first.shape,
                         count_shape=count.shape,
                     )
+                    continue
```
tmichela: we return here because code will break later if `count` and `first` shape mismatch

takluyver:
LGTM

Thanks, I'd been aware in the back of my mind that validation needed some tests for failing cases as well as passing ones.

PR #24, opened 2020-02-25
Check keys without scanning entire file

takluyver:
Checking for the existence of a single key was scanning all the keys of that source (though it was cached). I thought this would be a relatively small amount of overhead, but a user at MID ran into a case where it seems to be ridiculously slow (familiar story?).

Testing with this script, the `AGIPD1M()` call takes 100s of seconds with master, and <2 seconds with this branch.

```python
from extra_data import RunDirectory
from extra_data.components import AGIPD1M
from time import monotonic

t0 = monotonic()
run = RunDirectory('/gpfs/exfel/exp/MID/202030/p900139/proc/r0036')
t1 = monotonic()
print(f'Opened run in {t1 - t0} s')

agipd = AGIPD1M(run)
t2 = monotonic()
print(f'AGIPD1M() in {t2 - t1} s')
```

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -101,6 +101,8 @@ class FileAccess:
         self._index_cache = {}
         # {source: set(keys)}
         self._keys_cache = {}
+        # {source: set(keys)} - including incomplete sets
+        self._known_keys = defaultdict(set)
 
     @property
     def file(self):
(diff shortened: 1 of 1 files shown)


tmichela:
It can probably be even further sped up for control (slow) data with the new file format. the `RUN` group in each files contains a list of all sources and keys for the whole run (replicated in all files).

takluyver:
We'll need to find some way to provide access to the RUN section, but for now I think this should check if a key exists in the CONTROL/INSTRUMENT sections, because that's what we access. CONTROL will be a subset of RUN, if I've understood correctly.

tmichela:
> CONTROL will be a subset of RUN, if I've understood correctly.

yep

tmichela:
LGTM

takluyver:
Thanks :-)

PR #25, opened 2020-02-26
Use of file cache to keep the number of opened files under the limit

egorsobolev:
This is a try to address the cause of #2

I implemented the file cache. It stores opened files in the hashed priority queue. Hash allows finding opened files by name. The priority queue is used to track the order of access to files. It the queue is full the new file pops the file accessed the longest time ago.

The file cache stores opened files wrapped in the special proxy object which intermediate the access to the real file from FileAccess. If file closed all FileAccess instances referenced this file may know that via this proxy object.

The interface is not transparent but requires minimal changes in FileAccess

The changes pass tests and work for me at very populated runs. But the behaviour in delayed reading (e.g. dask array or during iterations over trains) may be affected by the closing of files in file cache between the taking of dataset instance and actual reading. From another side, it is possible if the cache is smaller than the required number of files. It means such cases fail anyway even without file cache.

### DIFF
--- a/extra_data/__init__.py
+++ b/extra_data/__init__.py
@@ -39,8 +39,6 @@ __version__ = "1.0.0"
 from .reader import *
 from .stacking import *
 from .utils import *
-from .filecache import get_global_filecache
 
 
-__all__ = reader.__all__ + utils.__all__ + stacking.__all__ + \
-          ['get_global_filecache', ]
+__all__ = reader.__all__ + utils.__all__ + stacking.__all__
--- a/extra_data/__init__.py
+++ b/extra_data/__init__.py
@@ -39,6 +39,8 @@ __version__ = "1.0.0"
 from .reader import *
 from .stacking import *
 from .utils import *
+from .filecache import get_global_filecache
 
 
-__all__ = reader.__all__ + utils.__all__ + stacking.__all__
+__all__ = reader.__all__ + utils.__all__ + stacking.__all__ + \
+          ['get_global_filecache', ]
--- a/extra_data/filecache.py
+++ b/extra_data/filecache.py
@@ -13,24 +13,21 @@ across entire process
 import h5py
 from collections import OrderedDict
 
-class H5FileDummy(object):
-    """
-    The fake `h5py.File` allows to unify interface for FileAccess object
-    """
-    class id:
-        valid = 0
-        
-        
+class FileRef(object):
+    __slots__ = 'nref', 'fh'
+
+
 class FileCache(object):
     """
     FileCache is a collection for opened h5 files. It keeps the number of opened files 
     under the given limit popping files accessed longest time ago.
     """
-    dummy = H5FileDummy
-    
     def __init__(self, maxfiles=128):
         self._maxfiles = maxfiles
         self._cache = OrderedDict()
+        
+    def __del__(self):
+        self.clean()
         
     @property
     def maxfiles(self):
@@ -43,22 +40,20 @@ class FileCache(object):
         """
         n = len(self._cache)
         while n > maxfiles:
-            rp, rf = self._cache.popitem(last=False)
-            rf.close()
+            p, f = self._cache.popitem(last=False)
+            f.fh.close()
             n -= 1
         self._maxfiles = maxfiles
-        
+
     def clean(self):
         """
         Closes all cached files
         """
-        self.cat = {}
-        self.counter = 0
         while len(self._cache):
-            rp, rf = self._cache.popitem(last=False)
-            rf.close()
+            p, f = self._cache.popitem(last=False)
+            f.fh.close()
 
-    def get_or_open(self, filename):
+    def open(self, filename):
         """
         Returns the opened `h5.File` instance from the cache.
         It opens new file only if the requested file is absent.
@@ -69,14 +64,17 @@ class FileCache(object):
         """
         try:
             f = self._cache[filename]
+            f.nref += 1
             self._cache.move_to_end(filename)
-        except:
+        except KeyError:
             if len(self._cache) >= self._maxfiles:
                 rp, rf = self._cache.popitem(last=False)
-                rf.close()
-            f = h5py.File(filename, 'r')
+                rf.fh.close()
+            f = FileRef()
+            f.fh = h5py.File(filename, 'r')
+            f.nref = 1
             self._cache[filename] = f
-        return f
+        return f.fh
     
     def touch(self, filename):
         """
@@ -86,27 +84,32 @@ class FileCache(object):
         it provides the underying instance of h5File for reading.
         """
         self._cache.move_to_end(filename)
-        
+
     def close(self, filename):
+        f = self._cache.get(filename, None)
+        if f is not None:
+            if f.nref <= 1:
+                f.fh.close()
+                del self._cache[filename]
+            else:
+                f.nref -= 1
+        
+    def force_close(self, filename):
         """
         Pops file from cache and closes it
         
         Useful, if it is necessary to reopen some file for writing.
         """
-        try:
-            f = self._cache[filename]
-            f.close()
-            del self._cache[filename]
-        except:
-            pass
-    
+        f = self._cache.pop(filename, None)
+        if f is not None:
+            f.fh.close()
+
         
 import resource
-import atexit
 
 def set_global_filecache():
     nofile_rlimits = resource.getrlimit(resource.RLIMIT_NOFILE)
-    maxfiles = nofile_rlimits[0] >> 1
+    maxfiles = nofile_rlimits[0] // 2
     global _extra_data_file_cache
     _extra_data_file_cache = FileCache(maxfiles)
 
@@ -114,10 +117,9 @@ def get_global_filecache():
     """
     Returns the global instance of FileCache
     """
-    return _extra_data_file_cache
+    try:
+        return _extra_data_file_cache
+    except NameError:
+        set_global_filecache()
+        return _extra_data_file_cache
 
-set_global_filecache()
-
-@atexit.register
-def clean_global_filecache():
-    _extra_data_file_cache.clean()
--- a/extra_data/filecache.py
+++ b/extra_data/filecache.py
@@ -16,6 +16,9 @@ from collections import OrderedDict
 class FileRef(object):
     __slots__ = 'nref', 'fh'
 
+    def __init__(self, fh):
+        self.fh = fh
+        self.nref = 1
 
 class FileCache(object):
     """
@@ -27,7 +30,7 @@ class FileCache(object):
         self._cache = OrderedDict()
         
     def __del__(self):
-        self.clean()
+        self.close_all()
         
     @property
     def maxfiles(self):
@@ -45,7 +48,7 @@ class FileCache(object):
             n -= 1
         self._maxfiles = maxfiles
 
-    def clean(self):
+    def close_all(self):
         """
         Closes all cached files
         """
@@ -55,12 +58,12 @@ class FileCache(object):
 
     def open(self, filename):
         """
-        Returns the opened `h5.File` instance from the cache.
+        Returns the opened `h5py.File` instance from the cache.
         It opens new file only if the requested file is absent.
         If new file exceeds the *maxfiles* limit it pops file accessed most far.
 
-        For use of the file cache, FileAccess shold use `get_or_open(filename)`
-        instead of direct opening file with h5py
+        For use of the file cache, FileAccess shold use `open(filename)`
+        instead of direct opening file with `h5py.File`
         """
         try:
             f = self._cache[filename]
@@ -70,9 +73,8 @@ class FileCache(object):
             if len(self._cache) >= self._maxfiles:
                 rp, rf = self._cache.popitem(last=False)
                 rf.fh.close()
-            f = FileRef()
-            f.fh = h5py.File(filename, 'r')
-            f.nref = 1
+            fh = h5py.File(filename, 'r')
+            f = FileRef(fh)
             self._cache[filename] = f
         return f.fh
     
@@ -81,11 +83,15 @@ class FileCache(object):
         Move the touched file to the end of the `cache`
         
         For use of the file cache, FileAccess should use `touch(filename)` every time 
-        it provides the underying instance of h5File for reading.
+        it provides the underying instance of `h5py.File` for reading.
         """
         self._cache.move_to_end(filename)
 
     def close(self, filename):
+        """
+        Closes the underlying file if called the same times as `open(filename)`
+        Otherwise, just decreases the counter.
+        """
         f = self._cache.get(filename, None)
         if f is not None:
             if f.nref <= 1:
@@ -107,19 +113,11 @@ class FileCache(object):
         
 import resource
 
-def set_global_filecache():
-    nofile_rlimits = resource.getrlimit(resource.RLIMIT_NOFILE)
-    maxfiles = nofile_rlimits[0] // 2
-    global _extra_data_file_cache
-    _extra_data_file_cache = FileCache(maxfiles)
+def init_extra_data_filecache():
+    nofile = resource.getrlimit(resource.RLIMIT_NOFILE)
+    resource.setrlimit(resource.RLIMIT_NOFILE, (nofile[1], nofile[1]))
+    maxfiles = nofile[1] // 2
+    global extra_data_filecache
+    extra_data_filecache = FileCache(maxfiles)
 
-def get_global_filecache():
-    """
-    Returns the global instance of FileCache
-    """
-    try:
-        return _extra_data_file_cache
-    except NameError:
-        set_global_filecache()
-        return _extra_data_file_cache
-
+init_extra_data_filecache()
--- a/extra_data/filecache.py
+++ b/extra_data/filecache.py
@@ -11,8 +11,11 @@ The module provides global instance of file cache for using
 across entire process
 """
 import h5py
-import weakref
 from collections import OrderedDict
+
+class FileRef(object):
+    __slots__ = 'nref', 'fh'
+
 
 class FileCache(object):
     """
@@ -37,8 +40,8 @@ class FileCache(object):
         """
         n = len(self._cache)
         while n > maxfiles:
-            p, rf = self._cache.popitem(last=False)
-            rf().close()
+            p, f = self._cache.popitem(last=False)
+            f.fh.close()
             n -= 1
         self._maxfiles = maxfiles
 
@@ -47,8 +50,8 @@ class FileCache(object):
         Closes all cached files
         """
         while len(self._cache):
-            p, rf = self._cache.popitem(last=False)
-            rf().close()
+            p, f = self._cache.popitem(last=False)
+            f.fh.close()
 
     def open(self, filename):
         """
@@ -60,16 +63,18 @@ class FileCache(object):
         instead of direct opening file with h5py
         """
         try:
-            f = self._cache[filename]()
+            f = self._cache[filename]
+            f.nref += 1
             self._cache.move_to_end(filename)
         except KeyError:
             if len(self._cache) >= self._maxfiles:
                 rp, rf = self._cache.popitem(last=False)
-                rf().close()
-            f = h5py.File(filename, 'r')
-            r = weakref.ref(f, lambda o: self._cache.pop(filename, None))
-            self._cache[filename] = r
-        return f
+                rf.fh.close()
+            f = FileRef()
+            f.fh = h5py.File(filename, 'r')
+            f.nref = 1
+            self._cache[filename] = f
+        return f.fh
     
     def touch(self, filename):
         """
@@ -80,15 +85,24 @@ class FileCache(object):
         """
         self._cache.move_to_end(filename)
 
+    def close(self, filename):
+        f = self._cache.get(filename, None)
+        if f is not None:
+            if f.nref <= 1:
+                f.fh.close()
+                del self._cache[filename]
+            else:
+                f.nref -= 1
+        
     def force_close(self, filename):
         """
         Pops file from cache and closes it
         
         Useful, if it is necessary to reopen some file for writing.
         """
-        rf = self._cache.pop(filename, None)
-        if rf is not None:
-            rf().close()
+        f = self._cache.pop(filename, None)
+        if f is not None:
+            f.fh.close()
 
         
 import resource
--- a/extra_data/filecache.py
+++ b/extra_data/filecache.py
@@ -1,0 +1,161 @@
+import h5py
+
+class CachedFile(object):
+    def __init__(self, filename, prio, cache_instance=None):
+        self._filename = filename
+        self._fc_prio = prio
+        self._fc_pos = -1
+        self._file = h5py.File(filename, 'r')
+        self._cache = cache_instance if cache_instance is not None else get_global_filecache()
+        
+
+    def close(self):
+        if self._file is not None:
+            self._file.close()
+            self._file = None
+            self._fc_pos = -1
+        
+    def __lt__(self, other):
+        return self._fc_prio < other._fc_prio
+
+    @property
+    def file(self):
+        if self._file is not None:
+            self._cache._touch(self._fc_pos)
+        return self._file
+    
+class DummyCachedFile(object):
+    file = None
+        
+        
+class FileCache(object):
+    dummy = DummyCachedFile
+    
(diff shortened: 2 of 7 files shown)


takluyver:
Thanks Egor. I like the structure of keeping a global queue shared between all EXtra-data objects and closing the oldest files when it fills up. But I think there's room to simplify this.

I don't think we need our own priority queue implementation. I think our requirements for this are pretty standard, and there are standard Python tools for this kind of thing. Specifically, as we only need to add/move objects to the end of the queue, I think we can use an [OrderedDict](https://docs.python.org/3/library/collections.html#collections.OrderedDict):

```python
# Add an item
od[path] = obj

# On accessing an already open file
od.move_to_end(path)

# Pop the oldest file, ready to close
path, obj = od.popitem(last=False)
```

I think that the objects we track in that can be our existing `FileAccess` objects, which can already use, lose and reopen the underlying `h5py.File` object. So I hope we can avoid needing another layer of file wrapper class.

egorsobolev:
I agree, Thomas. OrderedDict fits perfectly. Never looked at its implementation and in vain, as it turns out.

The second will not work while the file cache tracks only opened files (as I implemented). As the cache key is a path we must guarantee, that there is an only single instance of the wrapper for this file in the entire process. It means that wrapper constructor must check if there is already created instance. If the file cache tracks only open file then there is no way find an instance with a closed file.

There are two solutions to our choice. 

I chose the way to introduce additional wrapper which expires every time when its file is closing. The cache creates a new one if the file should be open again. And any object already has an expired wrapper has to request a new one from the cache. Kind of inversion of responsibilities.

Another solution is to track not only opened files but also closed ones untile they are still storing by other objects. This solution requires another dict for wrappers with closed files and wrappers has to have a counter of external reference on itself.

What does seem better for you?

takluyver:
I may well be overlooking something, but this is what I envision:
 
- `DataCollection` objects hold references to all the relevant `FileAccess` objects (open or closed)
- The cache holds only `FileAccess` objects which are open (have an associated `h5py.File` object)
- New `FileAccess` objects are only created after checking the cache for an existing one.
- When a `FileAccess` needs to open its file, it inserts itself into the cache (or moves itself to the end if it's already there)
- When a `FileAccess` drops out of the cache, it is closed, but one or more DataCollections may still have references to it. Doing something that needs data from the file will reopen it and put it back in the cache.
- When a `FileAccess` is explicitly closed  - e.g. using `with RunDirectory(...):` will close all the files when the block finishes - it removes its entry from the cache.

egorsobolev:
What if we create the second DataCollection on the partially or fully same set of files.

1. It immediately leads to an inconsistent concept, because we know nothing about the first DataCollection instance at the moment we create the second one and will duplicate FileAccess instances for closed files. We immediately fail an invariant, as we have some FileAccess instances shared between two DataCollection instances and others are private.

2. Even we will create all FileAccess instances as private, we fail invariant when one of the private FileAccess instances opens the file. It will be stored in the cache, others referred the same file will not.

3. Even if we copy the instance of h5py.File from the FileAccess instance stored in the cache to others when they try to open the file, it doesn't restore invariant. Only one instance in the cache with an open file, others are not in the cache.

4. We can properly close the file only if we manage to properly distribute the single h5py.File instance across all.

All this demonstrates that important point to have the single-to-single relation between the cache key, cached object, and h5py.File instance. 

Now I think, that it is possible to track in the cache directly h5py.File instances, if it may be identified as closed (and it has f.id.valid) and we don't need any additional attributes associated with the cache entry.

I will try to implement that and we give another look

takluyver:
Good point. My next thought would probably be to use a [WeakValueDictionary](https://docs.python.org/3/library/weakref.html#weakref.WeakValueDictionary) to hold closed `FileAccess` objects - this only keeps them if there's another reference to them somewhere. But it sounds like you've got another idea - feel free to explore that.

egorsobolev:
It seems working. Thomas, please, take a look.

extra_data/reader.py
```diff
@@ -104,14 +105,15 @@ def __init__(self, filename, _cache_info=None):
...
+        else:
+            self._file = get_global_filecache().get_or_open(self.filename)
+            
         return self._file
 
     def close(self):
-        if self._file:
-            self._file.close()
-            self._file = None
+        pass
```
takluyver: I think it should still be possible to close files explicitly, removing them from the cache as we do so. If I do:

```
with RunDirectory('...') as run:
    run.info()
```

I expect that to close any files deterministically when the block finishes.

There's some ugliness here because the same files might be open in another DataCollection, so closing them might be suboptimal. But if we have to choose one or the other, I'd prefer to be a bit too eager closing files rather than leaving open files lying around.

It may be possible to avoid having to choose one or the other, by keeping weakrefs in the cache so they automatically expire when no other references to them exist.

extra_data/filecache.py
```diff
@@ -0,0 +1,123 @@
...
+        except:
+            pass
+    
+        
+import resource
+import atexit
+
+def set_global_filecache():
+    nofile_rlimits = resource.getrlimit(resource.RLIMIT_NOFILE)
+    maxfiles = nofile_rlimits[0] >> 1
```
takluyver: Let's use integer division `// 2` instead of bitshifting to make the intent clearer.

extra_data/filecache.py
```diff
@@ -0,0 +1,123 @@
...
+        It opens new file only if the requested file is absent.
+        If new file exceeds the *maxfiles* limit it pops file accessed most far.
+
+        For use of the file cache, FileAccess shold use `get_or_open(filename)`
+        instead of direct opening file with h5py
+        """
+        try:
+            f = self._cache[filename]
+            self._cache.move_to_end(filename)
+        except:
```
takluyver: Let's catch `KeyError` specifically rather than all exceptions.

extra_data/reader.py
```diff
@@ -95,7 +96,7 @@ def __init__(self, filename, _cache_info=None):
 
             # Close the file again - crude way to avoid hitting the limit of
             # open files, which is only 1024 by default.
-            self.close()
```
takluyver: I agree that this can be removed - let's get rid of it entirely,  along with the comment.

extra_data/filecache.py
```diff
@@ -0,0 +1,123 @@
...
+        self._cache.move_to_end(filename)
+        
+    def close(self, filename):
+        """
+        Pops file from cache and closes it
+        
+        Useful, if it is necessary to reopen some file for writing.
+        """
+        try:
+            f = self._cache[filename]
```
takluyver: The [`dict.pop()` method](https://docs.python.org/3/library/stdtypes.html#dict.pop) is what we need here.

extra_data/reader.py
```diff
@@ -104,14 +105,15 @@ def __init__(self, filename, _cache_info=None):
 
     @property
     def file(self):
-        if self._file is None:
-            self._file = h5py.File(self.filename, 'r')
+        if self._file.id.valid:
```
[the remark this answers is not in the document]
egorsobolev: O, thanks! Indeed, a useful feature.

This is a tricky point:

1. Both branches in this `if` statement call `FileCache` methods with item search. So, we can remove `self._file` and simplify as
```
@property
def file(self):
    return get_global_filecache().get_or_open(self.filename)

2. Or find a way to store in self._file internal `OrderedDict` element wrapper called `Link`. Then we can move the element of OD to the end without searching in dict.

File: extra_data/reader.py (hunk 1 above)
Comment (egorsobolev) [reply extra_data/reader.py]: Even with `weakref` we cannot guarantee that the underlying file will be actually close. So it doesn't make sense for me to be eager here.
If it is nessecary to actually close the file I implemented methods `clean` and `close` in `FileCache`

Comment (takluyver) [reply extra_data/reader.py]: Looking at the code in h5py, I think files which drop out of scope will be closed correctly (it uses [`H5Idec_ref`](https://confluence.hdfgroup.org/display/HDF5/H5I_DEC_REF) on them). 

But as I said, if we're not confident we can do that properly, I would prefer to close files and potentially make some code re-open them rather than leaving unwanted file handles open.

File: extra_data/filecache.py (hunk 2 above)
Comment (egorsobolev) [reply extra_data/filecache.py]: accepted

File: extra_data/filecache.py (hunk 3 above)
Comment (egorsobolev) [reply extra_data/filecache.py]: accepted

File: extra_data/reader.py (hunk 4 above)
Comment (egorsobolev) [reply extra_data/reader.py]: accepted

File: extra_data/filecache.py (hunk 5 above)
Comment (egorsobolev) [reply extra_data/filecache.py]: accepted

File: extra_data/reader.py (hunk 1 above)
Comment (takluyver) [reply extra_data/reader.py]: In fact, h5py never calls `H5Fclose`. `file.close()` finds any objects belonging to the file, calls `H5Idec_ref` on them, and then calls `H5Idec_ref` on the file ID itself - the same function that's called when a file goes out of scope. So I think we could rely on reference counting for the cleanup. I'll open an issue on h5py to clarify this.

Comment (egorsobolev) [reply extra_data/reader.py]: I actually close explicitly all files in cache at the exit:
```
@atexit.register
def clean_global_filecache():
    _extra_data_file_cache.clean()
```

and I will add the __del__ method in FileCache then:
```
def __del__(self):
    self.clean()
```

I hope we can reset the cache without closing all files after the process fork in multiprocessing mode.

File: extra_data/reader.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (egorsobolev) [reply extra_data/reader.py]: I took a look at source code of OD once again. I think it is reasonable to go with 1 as there is c implementation of OD. I think we hardly can make faster with pure python but will pay by complications in code.

File: extra_data/reader.py (hunk 1 above)
Comment (takluyver) [reply extra_data/reader.py]: right, but `atexit` is only when the entire process exits, at which point our read-only file handles are going to get closed anyway. I still maintain that it should be possible to open a run using something like a `with` block, and have the files actually closed when you'd expect, even with the cache.

Comment (egorsobolev) [reply extra_data/reader.py]: `weakref` will require `WeakValueOrderedDict` implementation.  As an alternative to this bad idea, I can add a counter of `FileAccess` instances referred to a particular cached file. It will require to keep an `if` in `FileAccess.file`, and an additional method for soft close in `FileCache`. Not so complex as a full  `WeakValueOrderedDict`. But it doesn't make sense. 

We cannot guarantee that files after `with` block will be actually closed using a soft close. It means we must write in the documentation:
> Never expect that underling h5 files closed after destroying `DataCollection` instance. If you need forcefully close a file use `get_global_filecache().close(filename)`.

And it means that users must use this call anyway if they need to close a file or they may not care at all. Cleaning at `FileCache.__del__()` guarantee that all files will be closed properly at the end.   

So we are going waste efforts and make code more complex.

Comment (takluyver) [reply extra_data/reader.py]: I'll keep thinking about the best way to do it, but I do think the idea is important: whatever caching we add shouldn't keep files open after they would naturally be closed. I hope we can find a way to do that without a massive amount of extra complexity, but if it means some extra complexity, I think it's worth doing.

Comment (egorsobolev) [reply extra_data/reader.py]: Do you think that you are following the concept of isolation, which in fact should be accepted as a dogma in most cases. But in our case, it has no effect from the moment we make the file cache global.

However, to close our discussion, I implemented soft closing of files as you want. Although this is unnecessary in my opinion, in principle, it does not violate the invariant.

**Comment by egorsobolev** [comment]
Thank you, Thomas, for useful comments. I did corrections and implemented immediate soft close of underlying files. As an addition, I made the FileAccess serializable by resetting `_file` attribute in the state dict.
So, please, take another look.

File: extra_data/reader.py [hunk 7]
```diff
@@ -93,24 +95,27 @@ def __init__(self, filename, _cache_info=None):
...
-            self.close()
-
         # {(file, source, group): (firsts, counts)}
         self._index_cache = {}
         # {source: set(keys)}
         self._keys_cache = {}
 
+    def __del__(self):
+        if self._file:
+            self._fc.close()
```
takluyver: I think this needs the filename?

egorsobolev: yes, you are right. fixed it.

takluyver:
I've asked @tmichela to have a look at this as well - both to have a second pair of eyes in general and to consider whether I'm being overly insistent on ensuring the cache doesn't keep files open that would otherwise be closed.

takluyver:
I also realised this would need some minor changes to the `.trains()` iterators, both on `DataCollection` and the detector components, which cache dataset objects for repeated access. If the file containing a dataset is closed, the dataset reference will be invalid. Hopefully this should be easy: check the dataset coming from the cache, and fall back to the non-cached code path.

egorsobolev:
I added the datasets validity check in iterators

tmichela:
As I mentioned to @takluyver already yesterday, I believe that 99% of potential issues with open files can be mitigated by increasing the `NOFILE` to the hard limit (and eventually ask IT services to increase that hard limit if necessary), unless there in a clear downside of having too many files open?

That being said, this PR is probably useful anyway.
I have to say that I like Thomas' idea of keeping weak references, but I cannot think if any easy implementation for this...

Have you bench-marked the penalty cost of closing files when datasets are already referenced?

egorsobolev:
Benchmarks
-----------------

Maximum overhead we may expect for iterations over trains.
```python
run = open_run(2316, 70)
print("Files in run {}".format(len(run.files)))
```
```
Files in run 103
```
```python
%%time

fc.clean()
print("Files in cache {}/{}".format(len(fc._cache), fc.maxfiles))
print("Iterating over 10 runs... ", end='')
for tid, data in run.trains(train_range=by_index[:10]):
    tid
print('done')
print("Files in cache {}/{}".format(len(fc._cache), fc.maxfiles))
```
for `fc.maxfiles = 3`
```
Files in cache 0/3
Iterating over 10 runs... done
Files in cache 3/3
CPU times: user 35.5 s, sys: 4.66 s, total: 40.1 s
Wall time: 41.2 s
```
for `fc.maxfiles = 25`
```
Files in cache 0/25
Iterating over 10 runs... done
Files in cache 19/25
CPU times: user 11.9 s, sys: 2.73 s, total: 14.6 s
Wall time: 15.7 s
```

takluyver:
Thanks. I hope that in practical terms the cache will always be big enough to keep the current sequence files open, so it shouldn't introduce much overhead, but it's good to know that it works correctly when that's not the case.

Is it practical to introduce a test that covers that - make the cache size really small, and then iterate over `.trains()` for enough sources that it has to close and reopen files. It would be good to ensure that future changes don't break it.

egorsobolev:
Benchmarks show 3 times overhead for reopening files and datasets every time.

In real life, it isn't going to happen. The number of files which should be cached for next iterations is equal to the number of data aggregators: 16 for big detectors and few for all other sources, in total about 20.

It means that it is possible to work under the limit of a couple of dozen files. But we have examples when extra_data hits the limit of 4096 files. And that is ridiculous.

In practice, we will keep the size of cache big enough to keep all files cached in 99%. And popping mechanism will guarantee that in the last percent of worst cases extra_data will never crash.

egorsobolev:
I added the test. I also tried to turn from the custom implementation of weak reference to the standard one. Unfortunately, serialization fails. I still believe that the immediate closing of files is redundant.

extra_data/filecache.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        f = self._cache.pop(filename, None)
+        if f is not None:
+            f.fh.close()
+
+        
+import resource
+
+def set_global_filecache():
+    nofile_rlimits = resource.getrlimit(resource.RLIMIT_NOFILE)
```
takluyver: @tmichela suggested that we should simultaneously raise the limit to the hard limit (4096) at this point, so that all but the largest of runs can be accessed without some of their files getting closed again. I'm inclined to agree, unless you know of a reason why increading the limit of open files might cause other problems.

extra_data/filecache.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    FileCache is a collection for opened h5 files. It keeps the number of opened files 
+    under the given limit popping files accessed longest time ago.
+    """
+    def __init__(self, maxfiles=128):
+        self._maxfiles = maxfiles
+        self._cache = OrderedDict()
+        
+    def __del__(self):
+        self.clean()
```
takluyver: Is a `__del__` method necessary? In normal usage there will only be one `FileCache` and it will last until the process terminates, right?

extra_data/filecache.py
```diff
@@ -0,0 +1,125 @@
...
+        n = len(self._cache)
+        while n > maxfiles:
+            p, f = self._cache.popitem(last=False)
+            f.fh.close()
+            n -= 1
+        self._maxfiles = maxfiles
+
+    def clean(self):
+        """
+        Closes all cached files
```
takluyver: I think calling this method `.close_all()` might be clearer.

extra_data/filecache.py
```diff
@@ -0,0 +1,125 @@
...
+            p, f = self._cache.popitem(last=False)
+            f.fh.close()
+
+    def open(self, filename):
+        """
+        Returns the opened `h5.File` instance from the cache.
+        It opens new file only if the requested file is absent.
+        If new file exceeds the *maxfiles* limit it pops file accessed most far.
+
+        For use of the file cache, FileAccess shold use `get_or_open(filename)`
```
takluyver: get_or_open() -> open()

extra_data/filecache.py
```diff
@@ -0,0 +1,125 @@
...
+            f = self._cache[filename]
+            f.nref += 1
+            self._cache.move_to_end(filename)
+        except KeyError:
+            if len(self._cache) >= self._maxfiles:
+                rp, rf = self._cache.popitem(last=False)
+                rf.fh.close()
+            f = FileRef()
+            f.fh = h5py.File(filename, 'r')
+            f.nref = 1
```
takluyver: I'm inclined to give `FileRef` a `__init__()` method so these three lines are one. It may seem pointless, but it's nice to have objects which can't be left half initialised.

extra_data/filecache.py
```diff
@@ -0,0 +1,125 @@
...
+            self._cache[filename] = f
+        return f.fh
+    
+    def touch(self, filename):
+        """
+        Move the touched file to the end of the `cache`
+        
+        For use of the file cache, FileAccess should use `touch(filename)` every time 
+        it provides the underying instance of h5File for reading.
```
takluyver: h5File -> h5py.File

extra_data/filecache.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        self._cache.move_to_end(filename)
+
+    def close(self, filename):
```
takluyver: Let's have a docstring explaining that this decreases a refcount, and may not actually close the file.

extra_data/reader.py
```diff
@@ -93,24 +95,27 @@ def __init__(self, filename, _cache_info=None):
...
-            self.close()
-
         # {(file, source, group): (firsts, counts)}
         self._index_cache = {}
         # {source: set(keys)}
         self._keys_cache = {}
 
+    def __del__(self):
+        if self._file:
+            self._fc.close(self.filename)
```
takluyver: Can this call `self.close()`?

extra_data/reader.py
```diff
@@ -72,6 +73,7 @@ class FileAccess:
     file: h5py.File
         Open h5py file object
     """
+    _fc = get_global_filecache()
```
takluyver: This runs on import, which I think is what you were trying to avoid by defining a `get_global_filecache` function in the first place. It's generally good practice to avoid running things on import, so maybe this belongs in `__init__`.

extra_data/reader.py
```diff
@@ -162,6 +167,11 @@ def __eq__(self, other):
     def __repr__(self):
         return "{}({})".format(type(self).__name__, repr(self.filename))
 
+    def __getstate__(self):
```
takluyver: Let's have a brief docstring/comment to mention that this allows pickling.

extra_data/tests/test_reader_mockdata.py
```diff
@@ -621,3 +621,28 @@ def test_permission():
...
+    fc.clean()
+    run = RunDirectory(mock_spb_raw_run)
+    trains_iter = run.trains()
+    tid, data = next(trains_iter)
+    assert tid == 10000
+    device = 'SPB_IRU_CAM/CAM/SIDEMIC:daqOutput'
+    assert device in data
+    assert data[device]['data.image.pixels'].shape == (1024, 768)
+    assert len(fc._cache) == len(run.files)
+    fc.maxfiles = 3
```
takluyver: This will modify the filecache for all tests that run after this one as well. That's not a big problem while everything's working, but it's annoying if we end up with tests that pass or fail depending on what other tests have run before them.

Probably the neatest way to avoid this is to create a new pytest fixture which can reset the global filecache before and after a test which modifies it.

extra_data/tests/test_run_files_map.py
```diff
@@ -69,6 +70,9 @@ def map_paths_for_run(self, directory):
     assert isinstance(file_info['train_ids'], np.ndarray)
     assert isinstance(file_info['control_sources'], frozenset)
     assert isinstance(file_info['instrument_sources'], frozenset)
+    
+    fc = get_global_filecache()
+    fc.force_close(extra_file)
```
takluyver: If closing files is working correctly, this addition shouldn't be necessary any more, right?

extra_data/__init__.py
```diff
@@ -39,6 +39,8 @@
 from .reader import *
 from .stacking import *
 from .utils import *
+from .filecache import get_global_filecache
```
takluyver: I don't think this needs to be exposed at the top level of extra_data - it's a detail that won't matter for most users.

extra_data/filecache.py
egorsobolev: I agree. The soft limit is already equal to the hard limit from my experience. Of course, we cannot rely on that.

extra_data/filecache.py
egorsobolev: I prefer to keep. This doesn't ask for food but will work in a strange but possible situation if someone decided to delete and recreate the `Filecache`.

extra_data/filecache.py
egorsobolev: Ok

extra_data/filecache.py
egorsobolev: thanks

extra_data/filecache.py
egorsobolev: I will fix.

extra_data/reader.py
egorsobolev: Ok

extra_data/tests/test_run_files_map.py
egorsobolev: In this particular case, of course, the files are closed. As we discussed, we cannot guarantee that files will be actually closed in all cases (e.g. many `DataCollections`). Thus, the good practice is to close them explicitly anyway. I prefer to keep as is.

This code is also documentation, kind of a reminder.

extra_data/__init__.py
egorsobolev: Referring to a previous comment, users have to use it. But they need to close files relatively rare. So, let it be exposed at `extra_data.filecache`.

extra_data/filecache.py
egorsobolev: thanks

extra_data/filecache.py
egorsobolev: ok

extra_data/reader.py
egorsobolev: Yes. That is kind of my personal style, but I don't care.

extra_data/reader.py
egorsobolev: I don't see the problem with the initialization of global variables at import time (takin into account that this code runs only once). But I agree, I did it too complex. I will create the global cache in the `filecache` module and will use it directly without additional assignment to class variable. Does it suit?

extra_data/tests/test_reader_mockdata.py
egorsobolev: I agree.

extra_data/tests/test_reader_mockdata.py
```diff
@@ -14,8 +14,8 @@
 from extra_data import (
     H5File, RunDirectory, by_index, by_id,
     SourceNameError, PropertyNameError, DataCollection, open_run,
-    get_global_filecache,
 )
+from extra_data.filecache import extra_data_filecache
```
takluyver, line 18:
```suggestion
```

PR #30, opened 2020-02-27
Parallel file access

tmichela:
Multiprocessing implementation for:
- Open `FileAccess`
- run validation

Benchmark for opening files:
| # files | # process | time (s) |
|---------|-----------|----------|
| 207     | 1          | 37.9     |
| 207     | 10        | 7.49     |
| 207     | 20        | 4.42     |
| 207     | 30        | 3.57     |
| 207     | 40        | 2.69     |
| 207     | 80        | 2.27     |

---
This is a simple Multiprocessing implementation I use for testing file opening. It scale extremely well:

Running extra-data-validate goes down from 40 minutes to <1minute, with 40 processes:

```bash
[tmichela@max-exfl129]/gpfs/exfel/exp/SCS/201901/p002384/raw% time extra-data-validate ./r0144
Checking run directory: ./r0144
311/311 files                                                                                                                                         
No problems found
extra-data-validate ./r0144  37.23s user 20.75s system 101% cpu 57.280 total


[tmichela@max-exfl129]/gpfs/exfel/exp/SCS/201901/p002384/raw% time extra-data-validate ./r0144 
Checking run directory: ./r0144
311/311 files                                                                                                                                         
No problems found
extra-data-validate ./r0144  19.06s user 5.80s system 1% cpu 39:19.12 total
```

There are still issues to fix, especially handling KeyboardInterrupt (that seems a hassle), how to setup a proper timeout (opening time ca vary a lot if files are on dcache) to not get stuck infinitely if a files hangs, ...

Can be extended to RunDirectory


takluyver:
Nice, thanks.

With the timing, did you take some steps to ensure the difference wasn't due to caching? Accessing files the second time is always much quicker.

tmichela:
> With the timing, did you take some steps to ensure the difference wasn't due to caching? Accessing files the second time is always much quicker.

Yes, I ensured it was all locally cached (I ran the fast version first btw :)
Also, when the data is not yet cached there is a very large variance if time from file to file (0.5 to 5 sec)

egorsobolev:
In jupyter notebook:
```
v = RunValidator(path, term_progress=True)
v.run_checks()

timeout: 629.125

1/233 files (0 problems): RAW-R0064-DA02-S00001.h5
2/233 files (0 problems): RAW-R0064-DA02-S00005.h5
...
232/233 files (0 problems): RAW-R0064-AGIPD07-S00010.h5
233/233 files (0 problems): RAW-R0064-AGIPD11-S00010.h5
233/233 files
Exception ignored in: <function ProcessPoolExecutor._start_queue_management_thread.<locals>.weakref_cb at 0x2b9d5f8acf28>
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1/lib/python3.7/concurrent/futures/process.py", line 567, in weakref_cb
    thread_wakeup.wakeup()
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1/lib/python3.7/concurrent/futures/process.py", line 89, in wakeup
    self._writer.send_bytes(b"")
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1/lib/python3.7/multiprocessing/connection.py", line 183, in send_bytes
    self._check_closed()
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1/lib/python3.7/multiprocessing/connection.py", line 136, in _check_closed
    raise OSError("handle is closed")
OSError: handle is closed
```

takluyver:
Thanks, just checking. :-)

Looking at the code, I think we might be able to simplify things a bit using `multiprocessing.Pool` instead of concurrent.futures:

- The `imap_unordered` method should take care of submitting the tasks and getting results as they're completed.
- There's more explicit support for terminating processes, so hopefully we don't need to massacre all subprocesses. :-)

egorsobolev:
Yes, I agree. `concurrent.futures` is a high-level wrapper around multiprocessing and limits its features. `imap_unordered` looks suitable. There is an option to replace the manager and change communication to the use of shared objects if necessary. 

I've updated my comment and add some aspects.

tmichela:
I tried using `multiprocessing.Pool` that works nicely and `KeyboardInterrupt` is nicely handled. I still don't know how to handle propertly hanging files on dcache (maybe I shouldn't care). I tried various things unsuccessfully...

@egorsobolev I remember that you tried something similar when you were scanning runs, maybe you have some ideas?

I also added a progress bar :D the single number increasing was not easy to read as the cursor stays on it...


```bash
[tmichela@max-exfl015]~% time extra-data-validate /gpfs/exfel/exp/MID/201901/p002336/raw/r0120
Checking run directory: /gpfs/exfel/exp/MID/201901/p002336/raw/r0120
Progress: |██████████████████████████████████████████████████| 615/615 files : 2 problems                  
Validation failed! 2 problems:

Train IDs are not strictly increasing, e.g. at 0 (549920131 >= 549920131)
  dataset: INDEX/trainId
  file: /gpfs/exfel/exp/MID/201901/p002336/raw/r0120/RAW-R0120-AGIPD10-S00000.h5

Train IDs are not strictly increasing, e.g. at 0 (549920131 >= 549920131)
  dataset: INDEX/trainId
  file: /gpfs/exfel/exp/MID/201901/p002336/raw/r0120/RAW-R0120-AGIPD02-S00000.h5
extra-data-validate /gpfs/exfel/exp/MID/201901/p002336/raw/r0120  27.34s user 10.08s system 731% cpu 5.113 total
```

615 files
uncached: ~20s
cached: ~5s

egorsobolev:
Yes, I've implemented the scan with timeouts. But I did that with a single worker, not with pool.

The problems here that the IO operations to fast disks (and to pnfs) are not interruptible. It shouldn't be a problem but it is on the pnfs times to times. If data are not available at the moment the IO calls may be hanging forever. And we don't have any way to cancel disk operation and properly finalize (close communication connections) the worker process except explicitly kill it (with signal 9) without any finalization.

If you kill the process and it actually was removed from the memory it still may be in the process table if some parents processes wait for input/output. For example, it may be the connection to the console.

I would suggest closing the standard outputs (or reopen them and redirect to /dev/null) in the initialization procedure in order to deattach the process from the console:
`` ``
    def initializer ():
        # prevent child processes from receiving KeyboardInterrupt
        signal.signal (signal.SIGINT, signal.SIG_IGN)
        # close files
        sys.stdout.close ()
        sys.stderr.close ()
`` ``
I looked at the implementation of `Pool` in` multiprocessing` module. It looks that `Pool` doesn't track what workers execute tasks. So we may only kill the entire pool only.
Probably we may change a little implementation: enable workers to answer if they received a task, implement timeouts per task and kill only hanging workers.

extra_data/validation.py
```diff
@@ -193,24 +219,23 @@ def progress(self, line):
...
+        with Pool(initializer=initializer) as pool:
+            for fa, problems in pool.imap_unordered(_check_file, filepaths):
+                self.problems.extend(problems)
+                if fa is not None:
+                    self.file_accesses.append(fa)
+                    fa.close()
 
-        self.progress("{0}/{0} files".format(len(self.filenames)))
+                    self.progress(
+                        len(self.file_accesses), nfiles, len(self.problems))
```
takluyver: This should be outside the `if` block,  and with a separate counter rather than using `len(self.file_accesses)`.  A file that can't be opened still gets you closer to finishing validation.

extra_data/reader.py
```diff
@@ -289,18 +292,39 @@ def __init__(self, files, selection=None, train_ids=None, ctx_closes=False):
             train_ids = sorted(set().union(*(f.train_ids for f in files)))
         self.train_ids = train_ids
 
+    @staticmethod
+    def _open_file(path, cache_info=None):
+        try:
+            fa = FileAccess(path, _cache_info=cache_info)
+        except Exception as e:
+            print("Skipping file", path, file=sys.stderr)
+            print("  (error was: {})".format(e), file=sys.stderr)
```
takluyver: It might be necessary to transmit the error message back to the main process and show it there - both for the tests, and because I'm not sure printing from a multiprocessing child process would show up in Jupyter.

extra_data/reader.py
```diff
@@ -289,18 +292,39 @@ def __init__(self, files, selection=None, train_ids=None, ctx_closes=False):
...
             else:
-                files.append(fa)
+                uncached.append(path)
+
+        if uncached:
+            def initializer():
+                # prevent child processes from receiving KeyboardInterrupt
+                signal.signal(signal.SIGINT, signal.SIG_IGN)
+
+            nproc = min(len(psutil.Process().cpu_affinity()), len(uncached))
```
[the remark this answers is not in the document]
tmichela: There a different scenarios:
- data is on gpfs and not cached on the node: more processes is better (usual case where multiprocessing is used), see few tests:

| # files | # process | time (s) |
|---------|-----------|----------|
| 207     | 10        | 7.49     |
| 207     | 20        | 4.42     |
| 207     | 30        | 3.57     |
| 207     | 40        | 2.69     |
| 207     | 80        | 2.27     |

- data is on gpfs and cached locally: not much influence. I support that in this case, the `map` files are already generated so multiprocessing won't be used.

- data on `archive`: I don't know if that's just today, but it is insanely slow! maybe I should test again later (do you have any experience?) - we absolutely need to have the map files here (I remember that they are invalidated once data is moved to archive :-( )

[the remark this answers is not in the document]
takluyver: OK, so it looks like we can continue to benefit from more processes, even if not linearly. Let's leave this as is, then, and see if anything comes up that suggests a cap would be sensible. Thanks for investigating this systematically.

Archived data: as I understand it, once the data is archived on dCache, it may need to be read from tape, which may be arbitrarily slow, especially if the tape machines area already in use. I don't know if there's a convenient way to check 'is this file cached on disk?' without attempting to read it.

At the moment, even listing a directory seems to be ridiculously slow, though. Last time that happened, I think it was something Robert did...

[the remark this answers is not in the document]
tmichela: >  Last time that happened, I think it was something Robert did...

haha :D
OK so it is not a normal situation, I'll check tomorrow and email IT if that does not improve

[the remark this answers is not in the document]
tmichela: @RobertRosca what did you do last time? :o

extra_data/validation.py
```diff
@@ -177,39 +204,38 @@ def run_checks(self):
...
         return self.problems
 
-    def progress(self, line):
+    def progress(self, done, total, fname, nproblems):
         """Show a line of progress information"""
         if not self.term_progress:
             return
 
+        line = progress_bar(done, total)
+        line += f'\n{nproblems} problems: {fname}'
```
takluyver: Thanks for responding to that comment, but this won't do quite what I meant. Displaying the *next* file it's about to check is useful information if it gets stuck. I don't think it's worth showing the last checked file - it doesn't mean much, and it can be misleading if people think that's currently being checked.

I think this is just something we lose by parallelising the code. The massive speedup is worth a minor trade-off!

extra_data/validation.py
```diff
@@ -158,6 +160,31 @@ def check_index_contiguous(firsts, counts, record):
     return probs
 
 
+def progress_bar(done, total, sufix=' '):
+    line = f'Progress: {done}/{total}{sufix}[{{}}]'
```
takluyver: Minor spelling fix

```suggestion
def progress_bar(done, total, suffix=' '):
    line = f'Progress: {done}/{total}{suffix}[{{}}]'
```

extra_data/reader.py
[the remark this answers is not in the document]
takluyver: Nitpick: let's have a comment to clarify that cpu_affinity gives us cores we can use, which may be a subset of the cores the machine has. I don't think that's obvious without reading psutil's docs.

extra_data/validation.py
tmichela: What about something like:

```
Checking run directory: /gpfs/exfel/exp/XMPL/201750/p700000/raw/r0006
Progress: 21/21 [##################################################]
128 problems in 16 files (last: RAW-R0803-AGIPD03-S00000.h5)
```

?

takluyver: I'm inclined to drop the filename altogether - I'm not sure the last file checked tells you anything useful.

tmichela: ah, it's not the last checked file, it's the last checked file that has problems

takluyver: OK, fair enough.

extra_data/validation.py
```diff
@@ -177,39 +204,43 @@ def run_checks(self):
...
         return self.problems
 
-    def progress(self, line):
+    def progress(self, done, total, nproblems, badfiles):
         """Show a line of progress information"""
         if not self.term_progress:
             return
 
+        line = progress_bar(done, total)
+        line += f'\n{nproblems} problems'
```
takluyver: `line` now holds two lines ;-)

extra_data/validation.py
tmichela: I mean, if you think that it's not useful, I'll remove it, problems are showed at the end anyway (or reword if that's not clear enough that's showing file with problems)

takluyver: I'm not sure. Last with problems seems more likely to be useful than last checked, but I've been thinking about it too long to know if the meaning is clear for other people. I'm happy for this to go in either with or without that.

tmichela: Alright then, it stays for now :)

tmichela:
Thanks all for the review!

PR #32, opened 2020-02-28
WIP: Map API for parallel computation

philsmt:
For my previous work, I've developed a facilty-agnostic toolkit somewhat similar scope to EXtra-data called [xts](https://github.com/philsmt/xts). Its API is strongly focused around the map pattern, i.e. applying a common operation to all elements (in this case trains) of a collection (in this case FEL runs). Apart from personal preference for its expressiveness, this pattern allows a somewhat quick and easy path to parallelization, as the analysis of FEL data is often independant between trains and thus be distributed nicely based on train ID.

As a small side project, I've tinkered a bit with bringing such an API to EXtra-data. Due to the similarities of xts and EXtra-data, a proof-of-concept actually worked within minutes! Since then, I've reworked the code a bit and removed a lot of historical cruft from its early days.

A very simple example (actually working with proposal access!) for an average digitizer trace using several processes via reduction:

```python
import numpy as np
from extra_data.map_machinery import ProcessContext

ion_trace_src = ('SQS_DIGITIZER_UTC1/ADC/1:network',
                 'digitizers.channel_1_A.raw.samples') 
run = open_run(2310, 490) \
      .select_trains(by_index[:500]) \
      .select(*ion_trace_src)
mpc = ProcessContext(n_worker=2)

def sum_kernel(wid, tid, data):
    averages[wid] += data[ion_trace_src[0]][ion_trace_src[1]] \
                     .astype(np.float64)

averages = mpc.array((2100000,), np.float64, per_worker=True)
run.map(sum_kernel, mpc)
averages = averages.sum(axis=0) / len(run.train_ids)
```

Some early scaling [results](http://desy.de/~schmidtp/map_results.png) for this kernel over 27k trains quickly runs in the GPFS scaling limits. During beamtimes, the kernel code can become quite lengthy to compute averages, hit finding, digitization and coincidence event filtering with much better scaling results. Please find more samples at `/gpfs/exfel/data/scratch/schmidtp/extra_data_mp/`

This pattern makes heavy use of closures being nicely shared after fork or in threads. In the case of processes, the MapContext.array() method returns memory mmap'ed from shared memory. Both of these semantics make Windows-support quite a challenge. In addition, extending this to multi-node computation would probably lose out on these conveniences.

This is still WIP, but I'd like to start discussing the implementation details and namings now before going further. Some of the features in this PR may also be better split into distinct PRs.

* DataCollection has the new methods map, map_filter and map_sort.
* For convenience, there is enum_trains and split_trains.
* map and map_filter are completely uninstrusive and only require the map_machinery module.
* To support map_sort, DataCollection may now track a custom train order. This does not change any previous semantics, but is only exposed by the new by_sorted_* slicing.
* While I'm at it, I added absolute by_fraction as a shortcut as I've used such slicing quite often in the past.
* map_machinery contains the abstract interface and several implementations on map methods, currenty local, threads and processes. The default as of now is LocalContext, but I'd like a nice way to define a default context - the direct use might be too clunky for novice users. The context is also needed to allocate arrays for map operations.
* For reduction operations, an array for each worker may implicitly be allocated by passing per_worker=True. This injects an additional axis IN FRONT of the actual array. As Thomas K already pointed out, this is somewhat nebulous. In addition, this axis needs to be reduced before actual issue (as in the example above).
* There are stubs for some additional methods inspired from xts. Their implementation appears to be a bit more intrusive, so I'd like some feedback first.
* In principle, most of the get_* methods may be sped-up with this pattern.
* While I designed the MapContext interface around what's already there, it might be possible to make it work on top of dask.

Thanks!


takluyver:
Thanks! This is going to take a while to review. A couple of initial thoughts:

I think it's fine to say the parallelisation features are Unix-only, rather than aiming for full cross-platform support. I'd like to keep the basics of accessing data working on Windows, but realistically most people are using it on Linux, especially for substantial computations.

How important is the ordering part of this? I think there are significant simplifications that rely on `DataCollection` not really caring about order, and largely presenting data as it fits in the files. E.g. if you sort the collection by some variable and then use `.get_array()` or `.get_dask_array()` to access other data, does it need to shuffle the data into the order you created? Would you be open to doing this without the sorting and ordering parts initially, and discussing that as a separate step?

egorsobolev:
Great initiative as for me! I would add a couple of point for discussion:

Often computations are performed by trains or even by frames, but not always. Might it be more useful to implement a framework with different data slicing options? For example like `groupby` method, which defines some groups in data and the group pass then to the different kernels.

As I understand, DataCollection passes to the kernel as an argument, and actual data reading happens there.  At the moment it is the entire train. Is it true? And it could be too much if the user works with a big detector. And vice versa, it could be not effective to read them by frames. What do you think about a special layer which can read by slices which are most effective for reading? The computations often take no time in comparison with data reading.

Probably better to define output arrays as kernel arguments instead of global variables?

takluyver:
I discussed this in person with Philipp, but it's worth saying publicly as well: I'd like to keep the scope of this somewhat limited, and avoid building a 'framework'. There are a lot of powerful Python tools for data processing already, and the aim with EXtra-data has been to make it easy to get data into things like Dask and Xarray, rather than reinventing those wheels.

I'm persuaded that there might be a gap at the moment for simple scripts processing enough data that they benefit from reading in parallel. But I think we should set a fairly high bar for extra features, because it's easy to turn a simple interface into a complex one.

We should also seriously consider whether to build on top of something like Dask, rather than starting from multiprocessing. That would make it possible to run on multiple nodes with `dask_jobqueue`, and it might be easier to 'graduate' from our API to the full power of Dask.

takluyver:
Thinking about it more, building on Dask looks increasingly attractive. The `LocalContext`, `ThreadContext` and `ProcessContext` are neatly mirrored as Dask's synchronous, threads and processes [schedulers](https://docs.dask.org/en/latest/scheduling.html).

philsmt:
Thanks for your feedback!

> How important is the ordering part of this? I think there are significant simplifications that rely on `DataCollection` not really caring about order, and largely presenting data as it fits in the files. E.g. if you sort the collection by some variable and then use `.get_array()` or `.get_dask_array()` to access other data, does it need to shuffle the data into the order you created? Would you be open to doing this without the sorting and ordering parts initially, and discussing that as a separate step?

Yes, that is one of the item definitely worth splitting into its own PR.
A few quick details on the rationale: Due to exactly those implications I did not touch anything except slicing. The `train_order` property is meant to act as an optional overlay on top on how to view the data. "In the wild" I often encountered the problem when selecting a subset of data (e.g. "compare signal A of the top 10% of XGM intensity to signal B in the bottom 25% of XGM intensity" or "split this run into scans over source C"), but the actual computation on each of these subsets will then reduce the data anyway.

> Often computations are performed by trains or even by frames, but not always. Might it be more useful to implement a framework with different data slicing options? For example like `groupby` method, which defines some groups in data and the group pass then to the different kernels.

Such things could probably be done by extending DataCollection.trains() indepedently of this API. However, other boundaries than trains will introduce some issues with how to synchronize sources with different rates (e.g. should a kernel taking data from both AGIPD and a digitizer replicate the same trace for each frame in a train? Should be trace be split in time?). 

> As I understand, DataCollection passes to the kernel as an argument, and actual data reading happens there. At the moment it is the entire train. Is it true? And it could be too much if the user works with a big detector. And vice versa, it could be not effective to read them by frames. What do you think about a special layer which can read by slices which are most effective for reading? The computations often take no time in comparison with data reading.

Yes, I put split_trains() in DataCollection for that very reason. At the moment it's pretty naive and simply divides up the train IDs, but it could be made smarter than that based on what it contains. For a start, it could split along file boundaries.

> Probably better to define output arrays as kernel arguments instead of global variables?

To me, the possibility of using closures makes this API so "pretty". While conceptually it's comparable to OpenCL or any other modern graphics API, python allows to get away with so little boilerplate around it. See my thoughts below why I'd like to keep this simple.

> \<scope of EXtra-data\>

While I agree with Thomas' philosophy of not reinventing the wheel, I would like to add it's primary goal should be to empower our users first and honoring software engineering paradigms second. We should strive to get our users the best tools directly and avoid asking them to read the reference for package A and framework B. In a lot of cases, we might get away with both by leveraging the rich ecosystem as backend, but hiding it behind suitable APIs for our use. I'm well aware it's a fine line to NIY, but so is estranging our "customers". And for those users already familiar with other tools, we help them much earlier for free anyway (e.g. get_dask_array).

I see this as an example for it - I care for this paradigm and not the actual backend to make it work. Its original inspiration was near-online analysis in a highly interactive setting, i.e. sitting at the beamline during the experiment and needing something done in minutes, both writing code and waiting for the results. That is the very reason this API tries to be crafty with closures and shared memory, because it gets stuff done fast. I develop the kernel for one train, I slap it into map() and get results in a fraction of the original measuring time. The requirements of these situations are wildly different than for a well thought-out offline analysis or where tools with well defined data flows are present. If we can preserve this behaviour by layering it completely on top of dask (for single-node parallelization at least), let's go for it! I'd like to stress I simply have no experience with dask, it's possible all this stuff could have been done much cleaner from the start.

Finally, Egor's suggestions strike me as similar to the map_filter() or map_sort() methods. They make a particular problem significantly easier by a specific method of accessing data. I would see this well within the scope of EXtra-data.

takluyver:
> While I agree with Thomas' philosophy of not reinventing the wheel, I would like to add it's primary goal should be to empower our users first and honoring software engineering paradigms second. We should strive to get our users the best tools directly and avoid asking them to read the reference for package A and framework B.

This absolutely makes sense, but I believe we're serving users better by pointing them to existing frameworks than by making lots of custom tools; it's not just about software engineering paradigms. It gives people more to learn about initially - unless they've used the tools before - but then they get all the advantages of mature, general-purpose packages: all the bugs that someone else has found and fixed, answers on Stackoverflow when you get stuck, an ecosystem of further tools built on top, etc.

We could do better at smoothing the path to these tools with documentation. We touch on them in examples, but we're not far beyond "this gives you an xarray object - see the xarray docs!" Maybe it would be good to have notebooks with a 10-minute intro to each of pandas, Xarray and Dask, and include more about them when we run workshops about EXtra-data.

This is not to dispute the usefulness of the API you're adding, but I do want to defend my general goal of keeping EXtra-data comparatively small and relying on existing projects for a lot of the heavy lifting.

philsmt:
> [...] keeping EXtra-data comparatively small and relying on existing projects for a lot of the heavy lifting.

That's what I tried to actually say, but I've probably kept too many rhetorical doors open to layer features on features. I meant to say that I'd like us to not shy away from making it easier to leverage an existing project through a few more APIs,

That fundamental questions aside, how should we go foward with this? 
* Interface-wise, DataCollection.map(Callable) makes sense, but I already grapple with the rest of the map_* family. It feels out of place to add so much calls for a specific pattern ignored by the rest of this object. I'd like to only limit this feature for the moment to only map().
* Said convenience methods were essential to my work and I'd like to find a place for such functionality. Chaining them makes can make an analysis very expressive and sufficiently performant, e.g. `dc.map_filter(...).map_sort(...).select_trains(...).get_array(...)`. Let's try to discuss this in a following PR.
* Any sorting layer on top of DataCollection goes along with this utilities. 
* That leaves the backend: In general, I feel multiprocessing.Pool is a sufficiently common interface to provide a backend for it. I see the similarity with dask. I wonder though whether actually going the dask route makes this "simple scripts processing" quickly more complex that we'd like to. You'd like to explore layering it on top of dask first?

takluyver:
I'm sorry I missed the chance to chat with you about this in person last week.

I'm still wavering over whether we should investigate building this on top of Dask first. On the one hand, I can see the attraction of building something which really uses the advantages of fork & shared memory to tackle medium sized problems (parallel but on a single node) cleanly and efficiently. On the other, that's not what the name `map` means, and if we're going to provide a more generic interface for functional programming, I think we'd be serving users better by building on Dask, e.g. because it can easily use multiple nodes if needed.

Is it practical to split the two pieces up? i.e. A generic functional interface with parallelisation built on Dask, and some utilities to make it easy to use numpy arrays in shared memory for scenarios where workers are all on one host?

philsmt:
> that's not what the name `map` means

I would disagree that this implementation does not fall under the name `map` _as well_. Before submitting this pull request, I actually pondered about the name. I settled on map mostly based on the definition found in wikipedia, "a simple operation [that] is applied to all elements of a sequence, potentially in parallel.". I agree that there is a semantic difference to the builtin map(), but it is actually quite similar to Pool.map (hardly a surprise).

> Is it practical to split the two pieces up? i.e. A generic functional interface with parallelisation built on Dask, and some utilities to make it easy to use numpy arrays in shared memory for scenarios where workers are all on one host?

Possibly. The first candidate that caught my eye is [dask.delayed](https://docs.dask.org/en/latest/delayed.html). However, that API goes well beyond anything this tries to accomplish. It involves arbitrary graphs of operations expressed by lazily evaluated code (which in that sense looks functional) with clear inputs and outputs. I'm pretty sure this is exactly the boundary we should stay away from as you initially argued and get_dask_array() is the perfect handover. I didn't actually try it yet, but my first impression is it might be somewhat quirky, as we'd ignore the exectution graph completely and just ask for calling our function a few times via the scheduler.

After this lines, the question for me is whether dask.delayed allows to implement everything I'd like to do almost as simple as writing a Python callable. If yes, then, if any, there might be need for some dask utilities to make certain applications more straightforward. If not, then I have the feeling a less ambitious name might do the trick.

A very important caveat in this question is the environment in which it is used. I fear that properly setting up execution graphs in dask may often be just that tiny bit more involved that you simply do not want to do it minutes after the run is finished; even though from an engineering perspective, it is always the superior design.

Would you think differently about it if we call it `apply*`?  Amusingly, this makes it line up less favorably with multiprocessing.Pool. Maybe selling it as functional programming is the wrong approach and we should just treat it like Pool.map() in the context of DataCollection.

takluyver:
Sorry, I didn't explain that very well. 'map' is absolutely a good name for the API you've made, because it is, as you say, applying the same function to each thing in the sequence. But if the unique selling point of this API (compared to Dask, multiprocessing, etc.) is the convenient use of shared memory with parallel workers, then that's not obvious from a `DataCollection.map()` method.

If we were to build it on top of Dask, then maybe [dask.futures](https://docs.dask.org/en/latest/futures.html) is a better starting point than dask.delayed. AIUI, this is a lower level than the task graphs and lazy evaluation, and it includes a `map()` interface.

philsmt:
So, there I was today, sizzling in the sun and pondering my journey into dask last Friday. I'm pretty convinced now that there is room for a custom implementation to take advantage of the zero-copy shared memory scheme in this PR. However, I agree that any attempt to integrate it directly feels awkward at the moment.

So, instead of bringing parallel map support to EXtra-data, I brought EXtra-data to a parallel map library. I made the whole machinery more generic, packaged it standalone and added special EXtra-data support to make it the use of `DataCollection` as easy as an `ndarray`:
https://github.com/philsmt/pashmap

Please excuse the name for now (and please suggest a better one), maybe EXtra-pmap?
In principle, we could now:
- Keep it like this, add it to the exfel_anaconda3 environment for whoever likes it
- Go a bit further, still add the `map` call to DataCollection and this package into its namespace (something like `extra_data.pmap`), while keeping the allocation machinery nicely separated.

takluyver:
Thanks Philipp - I like the idea of providing the shared-memory processing facility as a separate package with some clear bridges between that and EXtra-data. I could see this being from both sides: the way you already have to map over a `DataCollection`, and a method in EXtra-data like `.shared_array()` or `.ndarray(shared=True)` to conveniently load data into a shared-memory array (e.g. a lot of the calibration is written to apply corrections with inplace operations).

A few thoughts on naming & API - feel free not to take this too seriously:

- If we call it EXtra-something, I'd emphasise the shared-memory aspect, e.g. EXtra-shared or EXtra-shmem. Parallel `map()` functions are everywhere - there are several in the Python standard library, and a host of third-party packages. The unusual bit is the focus on using fork & shared-memory parallelism.
  - For a similar reason, I'd make the process context the default - it's the central reason to use this code, so make it easy.
- If you're setting the default context, you have to do that *before* allocating arrays - this might not be obvious.
- `LocalContext` -> `SerialContext` ? "Local" means a lot of things, and in terms of parallelism my first guess would be "local to the current node".
- 'Targets' seems like a weird name for things you're using as inputs and reading data from. 'Sources'? 'Mappables'?
- `reduce_array` doesn't obviously suggest to me 'shape repeated per worker'. I can see the logic - it's a useful pattern for doing a reduction, but it sounds like a method to reduce an array.
- Should we add any facilities for mapping over another dimension of an array, or mapping over chunks rather than individual values? They're both conveniences, because you can always slice & dice the array before calling map, but I find it easier to follow "do X along axis 2" than "move axis 2 to 0, then do X".

philsmt:
Thanks! These are great suggestions and we should probably discuss those in its own repository. I will move the initial version in its own branch and start a PR there, we can discuss any naming schemes later.

I will close this PR here, we can come back later for a potential integration on this side. I would defintely like to move ExtraDataTarget here, as it could make use of internal informations about a DataCollection to optimize the parallelization.

PR #33, opened 2020-02-28
Update reader.py

trpreston5:
Fixes bug with run number not being correct format, or an alternative int type such as numpy.int64.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1447,18 +1447,15 @@ def open_run(proposal, run, data='raw', include='*'):
 
     prop_dir = find_proposal(proposal)
 
-    if isinstance(run, int):
-        run = 'r' + str(run).zfill(4)
-    elif isinstance(run, str):
+    if isinstance(run, str):
         if run.startswith('r'):
-            run = 'r' + str(run[1:]).zfill(4)
-        elif not run.startswith('r'):
-            run = 'r' + str(run).zfill(4)
-    else:
-        try:
-            run = int(run)
-        except ValueError:
-            print("Not a valid number")
-        run = 'r' + str(run).zfill(4)
+            run = run[1:]
+    try:
+        run = operator.index(run)
+    except ValueError:
+        print("Not a valid number")
+    if run < 1 or run > 9999:
+        raise(ValueError(f"run value '{run}' out of range"))
+    run = 'r' + str(run).zfill(4)
 
     return RunDirectory(osp.join(prop_dir, data, run), include=include)
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1451,7 +1451,7 @@ def open_run(proposal, run, data='raw', include='*'):
         if run.startswith('r'):
             run = run[1:]
     try:
-        run = operator.index(run)
+        run = int(run)
     except ValueError:
         print("Not a valid number")
     if run < 1 or run > 9999:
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks! This looks good. I think we can skip the `isinstance(run, int)` check, as that can be handled by the not-a-string branch.

It would be good to have a test for passing in a numpy int as well, so we're sure not to break it again in the future. This would be another few lines around here:

https://github.com/European-XFEL/EXtra-data/blob/ca3b59894ae5ea519725e2834ec818ad3b859922/extra_data/tests/test_reader_mockdata.py#L586-L588

trpreston5:
I agree that isinstance(run, int) is now redundant.
Also note that run = int(run) converts numpy ints and floats to integers. Strings are handled with isinstance(run, str) but it might still fail for really poorly formatted strings. The best advice is to always pass numbers that can be converted to int.

takluyver:
Good point - that reminds me. We might want `operator.index(run)` rather than `int(run)` - that will accept numpy ints but not floats.

egorsobolev:
probably exception on invalid value?
```
def strprrun(run):
    try:
        r = int(run[int(run[0] == 'r'):]) if isinstance(run, str) else run
        r.__index__()
    except:
        raise(ValueError(f"run value '{run}' is invalid"))
    # comment, if redundant
    if r > 9999:
        raise(ValueError(f"run value '{run}' is to big"))
    return r
```

takluyver:
@trpreston5 - by default we'll let you make the changes to your pull request. But if you're unlikely to get to it soon, just let us know, and we can take it from here.

extra_data/reader.py
```diff
@@ -1447,9 +1447,15 @@ def open_run(proposal, run, data='raw', include='*'):
...
 
-    if isinstance(run, int):
-        run = 'r{:04d}'.format(run)
-    elif not run.startswith('r'):
-        run = 'r' + run.rjust(4, '0')
+    if isinstance(run, str):
+        if run.startswith('r'):
+            run = run[1:]
+    try:
+        run = operator.index(run)
```
takluyver: `index()` doesn't accept strings, so we should put the try inside an `else` block.

And we'll need to import `operator`, as this is the first use in this module.

takluyver:
I prefer `index()` to `int()` because I don't think this should accept floating point numbers. I guess you could make a case that it should accept floats like 4.0, but it definitely shouldn't accept 4.5, or 4.999999.

trpreston5:
If we are going to tolerate string inputs for what is an integer then I would also err on the side of forgiveness for floats, especially if a user casts it accidentally as a float (4.0). In which case you have two options, a) accept the float and form it correctly as a integer, they may open the wrong file, or b) test if it is a float and make an error or warning because of it.

takluyver:
I allowed strings because that's what we ultimately work with, and there's no ambiguity what it means (hopefully no-one expects it to be octal or something ;-). But neither applies to floats, and it seems simpler to consistently reject all floats - as `operator.index()` does - than to accept floats and check if they make sense.

trpreston5:
But operator.index() does not accept strings so you would need to find some other way to check that your string run input (e.g. 'r24' or 'r0024') is within the range 1 to 9999. And if you then try to do an isinstance(run, float) approach to catch a float input then you fall over if it is a numpy.float32 or 64. The way it is written now will work and will convert whatever input to int, or give an error if it is not possible. If you want to be stricter on floats then you should work on this instead -- I have fixed the bug I set out to fix. I would however be more forgiving of float inputs -- I regularly define numbers as floats if I'm not sure what I'm going to do with them later, to avoid division errors, and then only convert to ints if necessary, and I'm sure many users do the same.

takluyver:
OK, thanks. I've opened PR #34 to continue these changes.

PR #34, opened 2020-03-03
Allow numpy integers for open_run()

takluyver:
This is a continuation of #33.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1447,18 +1447,15 @@ def open_run(proposal, run, data='raw', include='*'):
 
     prop_dir = find_proposal(proposal)
 
-    if isinstance(run, int):
-        run = 'r' + str(run).zfill(4)
-    elif isinstance(run, str):
+    if isinstance(run, str):
         if run.startswith('r'):
-            run = 'r' + str(run[1:]).zfill(4)
-        elif not run.startswith('r'):
-            run = 'r' + str(run).zfill(4)
-    else:
-        try:
-            run = int(run)
-        except ValueError:
-            print("Not a valid number")
-        run = 'r' + str(run).zfill(4)
+            run = run[1:]
+    try:
+        run = operator.index(run)
+    except ValueError:
+        print("Not a valid number")
+    if run < 1 or run > 9999:
+        raise(ValueError(f"run value '{run}' out of range"))
+    run = 'r' + str(run).zfill(4)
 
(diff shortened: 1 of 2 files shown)


extra_data/reader.py
```diff
@@ -1440,16 +1441,22 @@ def open_run(proposal, run, data='raw', include='*'):
...
+        # Allow integers, including numpy integers
+        proposal = 'p{:06d}'.format(index(proposal))
 
     prop_dir = find_proposal(proposal)
 
-    if isinstance(run, int):
-        run = 'r{:04d}'.format(run)
-    elif not run.startswith('r'):
-        run = 'r' + run.rjust(4, '0')
+    if isinstance(run, str):
```
trpreston5: Can just use
`    if isinstance(run, str) and run.startswith('r'):
          run = run[1:]`
or add something here to convert the string to a int and check that it is in range.

takluyver: Joining the conditions in one `if` statement changes when the `else` section runs, though. There's several different ways to achieve the same thing, but I think this one is as good as any.

I'm not too concerned about the range check - I think the `FileNotFoundError` it triggers by default is clear enough. And the actual range of runs that exist is different for each proposal, in any case.

```
FileNotFoundError: [Errno 2] No such file or directory: '/gpfs/exfel/exp/SCS/201901/p002212/raw/r0888'
```

takluyver: In fact, I wonder if it's best to take the range check out, so the error for a run that doesn't exist is consistent.

trpreston5: Okay - take it out. I only added it as your colleague suggested it.
However, be careful if someone stupidly passes the string "t024" instead of "r024" - is this handled?

takluyver: Oops, sorry for the inconsistent feedback.

Typo-ing something like `t024` will give the same kind of error, including the mistaken text as part of the path. We could catch things like that separately to make a better error message, but it's not high priority, so I'll leave it for another day.

egorsobolev: I just suggested discussing the check of a range because there is a little difference in the sense of errors: the run doesn't exist and it cannot exist in principle. But I don't think that it's important.

egorsobolev:
I think it's better to move the normalisation of `proposal` into `find_proposal`, isn't it?

Except that it looks fine to me.

takluyver:
Thanks Egor. As we're doing similar normalisation for proposal & run numbers, I prefer to keep them in the same function.

PR #35, opened 2020-03-04
Checking of file locality in order to avoid hang

egorsobolev:
This addresses the issue of hung if files locate in dCache and are available only on tape or even lost


takluyver:
Thanks. Before we get into implementation details, I think there's an important design question we need to consider: what should happen if part of a run is in an acceptable locality and part isn't?

As this is implemented, I think it will give you a DataCollection referring to only the files that it accepts, while printing out warnings for every file it rejects.

If we needed to open files to know what was in them, I would be inclined to say that the run as a whole should pass or fail, so it doesn't give you partial data. But this is complicated by the existence of cache files and (in the future) virtual overview files: when these exist, we can cheaply 'open' an entire run, and delay opening the actual data files until data is requested. So we could potentially open the run, and then error if the data the user wants is in an unwanted locality.

cc @fangohr @tmichela

egorsobolev:
The user should decide what behaviour he needs. I think we need not only print out warnings but also add in DataCollection (or return to user in any other way) the list of files with related exceptions.

If there is no additional information about the run, then there is nothing to do else.

The more interesting situation if we already have a cached index. In this case, in addition to the temporarily unavailable files, we might provide information to the user what datasets or trains are temporary missing. For example, it might be a "dead" DataCollection built by cached indexes of unavailable files.

egorsobolev:
In future developments, we might create fake `h5py.Files` which will just fill missed places with filling values or raise exceptions or both

tmichela:
Is it actually important in the user point of view to know where the data is? Is there a case where he would decide to not get data if it's on tape?

Regarding unavailability, I think that the check should be available, but probably not be done in the default case. Checking file availability is very expensive (up to 0.5 s per files). And especially because having unavailable files as now is not a normal situation, I don't expect that this will append regularly...

so I wonder if we should expose to the public method only one property that allow ignoring unavailable files (with default to not check). And I also wouldn't expose or make a configurable choice of data locality, but here maybe I miss some use case that can make use of it...

egorsobolev:
Users demands depends on their data and processing: they may ignore run (if they have many others with the data of similar data), may process another run first and then go back, may ignore missing file (if there are few of them in a huge run) and process rest part of data, or they have no other way and need to contact IT.

You should not decide for users what they actually need whether to spend a half a second and get a guarantee that processing doesn't hang or save half a second with a risk to have nothing at the end.

If there is any chance to save users efforts you must provide tools for that.

If users have no options except to contact IT and wait for weeks, they hate you.

tmichela:
> You should not decide for users what they actually need whether to spend a half a second and get a guarantee that processing doesn't hang or save half a second with a risk to have nothing at the end.

I don't want to decide anything right now, I just try to understand use cases :)

The reason I ask about that is that I know that some users don't want to use our tools because they think it is too slow, so if we add overhead of .5 seconds **per file** that can push more people away.
And the angle of my question was more: "if unavailable files occurs once in a while (lets say every few years), wouldn't it be more efficient to not check availability and have faster file opening? And when it happens, to tell them: please activate this flag as currently some files are unavailable and can hang the whole process"?

egorsobolev:
I  also don't want to say that files locality must be checked every time. It is good enough just to have this tool when it is really necessary.

Do you know also what percentage of this 0.5 sec **per file** takes in minimal reading operation? Just to know what we talking about.

takluyver:
Before we add checks at the Python level for this, I'd like to better understand:

1. What kind of problems have we seen and how often? I'm aware of times where dCache as a whole has had problems, in which case users won't be able to access any of that data, no matter what checks we have. And I recall one time where a couple of specific files became inaccessible, which we were assured was a one-off bug that had been fixed and was unlikely to recur. Are there other incidents I don't know about?

2. Have we talked to XFEL/DESY IT about these issues? It seems like a bug/misconfiguration in dCache that unavailable files simply hang rather than giving some sort of error. AIUI, DESY is one of the main organisations behind dCache, so there's a good chance we can get access to people who can either fix these things, or suggest ways that we can do things better.

I do also see the check typically taking ~0.5 seconds per file when I'm not repeating the call on the same file. If we have the cache or overview file existing and cached on disk, then there's no reason why inspecting a run or reading a small amount of data should not take < 1 second. Even if the locality check parallelises very well, for large runs it seems like it will certainly be an unwanted cost.

egorsobolev:
I personally had a number of problems:
1. Half a year ago I started to reprocess 2013 and I processed that in Moscow at the end of the day; 
2. Recently I wanted to process may run and estimate the portion of dark frames. I tried about 10 different runs on pnfs, all hang. It was a month ago, the same is right now.

The files are actually unavailable only on fast media and reading requires staging (coping from tape to the disks). Currently, any I/O operations are blocking while the file staging.

In a normal case, all files should be on disks as well as on tape. If the disk with the file is accidentally off, the staging shouldn't be too long for a few files (the rate is about hundreds MB/s). But in real life for some reasons the huge amount of files accidentally disappears from the disks simultaneously, there is a queue on tape readers for weeks and so on.

I believe that is really not a normal case. But it is the second time in a half year.

There is an option to change behaviour and return an error instead of blocking I/O operation. But I am not sure. Krzysztof sad that they decided to do it in the future. Perhaps we get it sometime.

So, all our discussion is about that by a chance I implemented a check for every time. I didn't pay attention that the check of the locality may take the time comparable with file reading.

I promise I will not check by default.

tmichela:
> I also don't want to say that files locality must be checked every time. It is good enough just to have this tool when it is really necessary.

Yes, I 100% agree

>  I recall one time where a couple of specific files became inaccessible, which we were assured was a one-off bug that had been fixed and was unlikely to recur. Are there other incidents I don't know about?

A couple lf weeks ago 1 node went down and 23000 files were lost. Took some time to recover as they had issues on gpfs in parallel. I think that most of the files are recovered by now.

> From my personal experience, there is no option to process data on pnfs at all. I never had a single success story.

I guess that's something we should feedback to IT. Your experience as a user can be valuable to make the system more useful

> In a normal case, all files should be on disks as well as on tape.

I am not sure what you mean here, and while this is true during beamtime (+a few weeks later) it is not the case and not possible if you want to process data long time after it's been taken

> There is an option to change behaviour and return an error instead of blocking I/O operation.

I wasn't aware about that. That's good new if it's possible, we should push for it

takluyver:
Would it make sense to provide a separate command line tool, like [this Fermilab script](https://cdcvs.fnal.gov/redmine/projects/novaart/repository/entry/trunk/NovaGridUtils/bin/cache_state.py), which can check the status of data on dCache, as a convenient way to inspect what's going on when we're stuck accessing it? I.e. take a similar approach to [hdf5-vds-check](https://github.com/European-XFEL/hdf5-vds-check).

In particular, I can see it being useful to summarise this for a run, with numbers of files in each locality.

egorsobolev:
Yes, it does. I will start on that. But this information is almost useless if there is no way to filter files in a run with it, is it?

egorsobolev:
```python
%%time
locality.check_directory('/gpfs/exfel/exp/SPB/201901/p002304/raw/r0023/', locality.FAST)
```
```
accepted: 207, filtered: 0
CPU times: user 64.6 ms, sys: 76 ms, total: 141 ms
Wall time: 27.6 s
```
```python
%%time
locality.check_directory('/gpfs/exfel/exp/SPB/201901/p002304/raw/r0023/', locality.FAST, 1)
```
```
accepted: 207, filtered: 0
CPU times: user 82.8 ms, sys: 101 ms, total: 184 ms
Wall time: 1min 34s
```

takluyver:
This [came up again](https://in.xfel.eu/redmine/issues/59772) over the weekend, and several runs currently have most files on disk, but one or two only on tape. So yes, I agree that we should provide a way for someone to filter the files and access only the data that's already on disk.

I put together a standalone script which can check the data locations of a run on demand. Sorry to duplicate your effort on this, but actually reading the information is pretty simple, and I wanted to quickly present it in a slightly different format:

```
% /gpfs/exfel/data/scratch/kluyvert/dcache-status.py /gpfs/exfel/exp/SPB/201802/p002160/raw/r0075
Checking 648 files in /gpfs/exfel/exp/SPB/201802/p002160/raw/r0075
646 on disk, 2 only on tape, 0 unavailable    
Only on tape:
  /gpfs/exfel/exp/SPB/201802/p002160/raw/r0075/RAW-R0075-AGIPD07-S00017.h5
  /gpfs/exfel/exp/SPB/201802/p002160/raw/r0075/RAW-R0075-AGIPD13-S00019.h5
```

egorsobolev:
So, what do we do with the PR and your script? It would be good to have some python function that returns the list of good/bad files.

takluyver:
I reckon that the basics of collecting the information by reading the `.(get)(filename.h5)(locality)` 'files' are so simple that it's easier to duplicate a bit of code than to make them use a common function. That may change if there's a more efficient way to get the information using Webdav - I've asked DESY IT about this, but I imagine it's not the top of their priority list today.

So I think this PR should be focused back on filtering the files when you use `RunDirectory()` or `open_run()` - sorry, I know this is a change of direction relative to my previous reviewing. As we discussed before, let's make the filtering possible but require the user to explicitly specify it for the time being, and only query dCache when the filtering is in use.

egorsobolev:
Added file filtering on-demand and changed output similar to your script:
```python
%%time
run = open_run(proposal=2304, run=32)
```
    CPU times: user 25.2 ms, sys: 16 ms, total: 41.3 ms
    Wall time: 50 ms

```python
%%time
run = open_run(proposal=2304, run=32, filt=locality.lc_ondisk )
```
    CPU times: user 156 ms, sys: 289 ms, total: 446 ms
    Wall time: 21.6 s
```python
%%time
locality.check_dir('/gpfs/exfel/exp/SPB/201901/p002304/raw/r0023/')
```
    Checking 207 files in /gpfs/exfel/exp/SPB/201901/p002304/raw/r0023/
    207 on disk, 0 only on tape, 0 unavailable    
    CPU times: user 614 ms, sys: 625 ms, total: 1.24 s
    Wall time: 1min 47s

takluyver:
Sorry, I forgot about this PR for quite a while. I think it's mostly in a good state now, but I have a couple of open questions:

- Should we document this as a fully public feature, or leave it semi-hidden as a sort of expert feature?
- If you select a filter other than `lc_any`, this will currently try to read the dCache locality file for every file, regardless of whether the folder is in dCache. I guess in most cases there's negligible impact of trying to open hundreds of files which don't exist, but should we consider having a shortcut that recognises that the entire folder is not on dCache?

egorsobolev:
Since external people noticed problem with dCache I think it would be good to add the section with explanation. It could be subsection 'File reading problems' in section 'Reading data files'.

Besides, new arguments of open_run and RunDirectory are already in references. I've also improved standalone script and added documentation about it.

Regarding non-dCache folders.  From one side, I don't see good implementation to check only dCache files. From another side, I don't think we need too smart program here. User will not check file locality if it makes no sence. Besides, it is a last resort.

docs/cli.rst
```diff
@@ -96,3 +96,18 @@ Make a virtual CXI file to access AGIPD/LPD detector data from a specified run:
 .. option:: --min-modules <number>
 
    Include trains where at least N modules have data (default 9).
+
+``extra-data-locality``
+------------------------
+
+Check the files localitly:
```
takluyver, line 103:
```suggestion
Check how the files are stored:
```

extra_data/locality.py
```diff
@@ -0,0 +1,146 @@
...
+    if argv is None:
+        argv = sys.argv[1:]
+
+    ap = ArgumentParser(prog='extra-data-locality', description="Checks locality of files in the directory")
+    ap.add_argument('path', help="run directory of HDF5 files.")
+    args = ap.parse_args(argv)
+    
+    if not os.path.exists(args.path):
+        print(f"Path '{args.path}' is not found")
+        return -1
```
takluyver, line 141:
```suggestion
        return 255
```

At least on my system, this appears to get translated to 255. Is it better to do that explicitly? Negative exit codes in some contexts mean a process was terminated by a signal, so it's a bit odd (for me, at least) to see something trying to return a negative exit code.

docs/cli.rst
```diff
@@ -96,3 +96,18 @@ Make a virtual CXI file to access AGIPD/LPD detector data from a specified run:
...
+
+.. code-block:: shell
+
+   extra-data-locality /gpfs/exfel/exp/XMPL/201750/p700000/raw/r0002
+
+The file reading may hang for a long time if files are unavailable or require staging
+in dCache from the tape. The program helps finding problem files.
+
+If it finds problems with the data locality, the program will produce a lists of files
```
takluyver, line 112:
```suggestion
If it finds problems with the data locality, the program will produce a list of files
```

takluyver:
Thanks Egor!

PR #36, opened 2020-03-05
Save map files on the online cluster

tmichela:
resolved path on the online cluster do not match Maxwell.
This adds matching pattern for online cluster.


takluyver:
Thanks, this LGTM. I assume that the scratch paths we build are the same on the online cluster?

tmichela:
Thanks.
Yes, the scratch paths are identical

PR #37, opened 2020-03-09
Correct code coverage for Multiprocessing

tmichela:
- Makes code coverage correct with multiprocesing.
- upgrade `coverage`

Multiprocessing support in `coverage` is available from version 5. And it seems that issues with `nbval` have been fixed.


takluyver:
Thanks, LGTM

PR #39, opened 2020-03-09
Filter trains based on flag status

tmichela:
- ignore trains in files if `INDEX/flag` is `0`
- update mockdata to include metadata and new index fields (timestamp, flag)
- update tests conf to run on both file version

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -88,7 +88,11 @@ class FileAccess:
             self.control_sources = _cache_info['control_sources']
             self.instrument_sources = _cache_info['instrument_sources']
         else:
-            tid_data = self.file['INDEX/trainId'][:]
+            if self.format_version != '0.5':
+                valid_trains = self.file['INDEX/flag'][:].nonzero()[0]
+            else:
+                valid_trains = slice(None)
+            tid_data = self.file['INDEX/trainId'][valid_trains]
             self.train_ids = tid_data[tid_data != 0]
 
             self.control_sources, self.instrument_sources = self._read_data_sources()
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -89,7 +89,7 @@ class FileAccess:
             self.instrument_sources = _cache_info['instrument_sources']
         else:
             if self.format_version != '0.5':
-                valid_trains = self.file['INDEX/flag'][:].nonzero()[0]
+                valid_trains = self.file['INDEX/flag'][:].nonzero()[0].tolist()
             else:
                 valid_trains = slice(None)
             tid_data = self.file['INDEX/trainId'][valid_trains]
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -89,6 +89,12 @@ class FileAccess:
             self.instrument_sources = _cache_info['instrument_sources']
         else:
             if self.format_version != '0.5':
+                # the INDEX/flag dataset contains a value (0 or 1) for each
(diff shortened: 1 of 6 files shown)


extra_data/tests/mockdata/base.py
```diff
@@ -139,9 +144,24 @@ def write_metadata(h5file, data_sources, chunksize=16, format_version='0.5'):
...
+        now = datetime.utcnow().replace(microsecond=0).strftime('%Y%m%dT%H%M%SZ')
+        h5file['METADATA/creationDate'] = [now.encode('ascii')]
+        h5file['METADATA/daqLibrary'] = [b'1.9.0']
+        h5file['METADATA/karaboFramework'] = [b'2.7.0']
+        h5file.create_dataset('METADATA/proposalNumber', dtype=np.uint32, data=[700000])
+        h5file.create_dataset('METADATA/runNumber', dtype=np.uint32, data=[1])
+        h5file['METADATA/runType'] = [b'Test DAQ']
+        h5file['METADATA/sample'] = [b'No Sample']
+        print(osp.basename(h5file.filename))
+        sequence = int(re.match(r'^(RAW|CORR)\-R\d+\-.*\-S(\d+).h5$', osp.basename(h5file.filename)).groups()[1])
```
takluyver: This assumes that we always use properly formatted names, which is true in conftest, but not if you do `python -m extra_data.tests.make_examples` . I'm ambivalent about whether we correct that or make a fallback to sequence = 0.

extra_data/reader.py
```diff
@@ -88,7 +88,17 @@ def __init__(self, filename, _cache_info=None):
...
+                # the INDEX/flag dataset contains a value (0 or 1) for each
+                # trainid recorded in the file. In normal case it is set to 1.
+                # If the DAQ receives a certain traindid from a data source
+                # before receiving it from the timeserver, it is assumed that
+                # the value for this trainid is wrong and we exclude it from
+                # the data collectiom.
+                valid_trains = self.file['INDEX/flag'][:].nonzero()[0].tolist()
+            else:
+                valid_trains = slice(None)
+            tid_data = self.file['INDEX/trainId'][valid_trains]
```
takluyver: Unfortunately, I think it will need several more changes in this file. We assume that the `.train_ids` array is aligned with data read from the indexes, e.g.

https://github.com/European-XFEL/EXtra-data/blob/27733a42bb3de41690f8648cd07ebcefda5f3dfa/extra_data/reader.py#L202-L204

https://github.com/European-XFEL/EXtra-data/blob/27733a42bb3de41690f8648cd07ebcefda5f3dfa/extra_data/reader.py#L592

https://github.com/European-XFEL/EXtra-data/blob/27733a42bb3de41690f8648cd07ebcefda5f3dfa/extra_data/reader.py#L1128-L1132

That's not an exhaustive list.

I think we'll need an additional attribute to select which values should be taken from indexes.

(This always relied on there being no 0s in `INDEX/trainId` except at the end. `extra-data-validate` checks for this, and I'm not aware that this has ever gone wrong.)

takluyver:
We discussed another possible approach: keep all trains at the `FileAccess` level, but provide a way for `DataCollection` to exclude trains that might be dubious. That would mean that train is excluded across all files, when it might only have a problem in one. But when the spurious train ID is after the normal end of the run, it makes no difference. And so long as out-of-sequence trains remain relatively rare, dropping them entirely might be a reasonable solution.

takluyver:
This was superseded by PR #121.

Issue #14, opened 2020-01-31
Method to get data from files into a single-source ZeroMQ stream


dallanto:
A new branch "file_stack" was created, and a prototype script pushed to it.
If this is considered useful, a merge request for a method in export.py may follow.

takluyver:
Thanks! I think this might be useful as an option for `karabo-bridge-serve-files`; it's a bit technically impure that we're pretending to be a karabo device, but as the combined detector data is a key use case for Karabo bridge, I think it makes sense to be able to present data from files in the same format that it comes online.

I'd be interested in @tmichela's thoughts too.

tmichela:
Agreed with @takluyver , I would have it as an option of `serve-files`

dallanto:
See pull requet "File stack" #40

takluyver:
Done in #40 and #51.

PR #40, opened 2020-03-10
File stack

dallanto:
@takluyver , @tmichela  
This follows up issue #14.  
My tentative stand-alone CLI front-end was integrated to the serve-files code, adding an argument "data-like" (in analogy to the bridge-server-sim tool) for optional module stacking and single-source write-out.


extra_data/export.py
```diff
@@ -258,6 +287,12 @@ def main(argv=None):
         "--key", help="Stream only matching keys ('*' is a wildcard)",
         default='*',
     )
+    ap.add_argument(
+        "--data-like", help="Optional module stacking: online -> stack module \
+           sources into a single source (default: file -> keep modules \
+           separate as in files)",
+        default='file', choices=['file', 'online']
```
takluyver: I'd call this something like `--append-detector-modules`. It's possible to connect a bridge to an individual module instead of the appender, and we've considered the possibility of using the TrainMatcher (which would send a message with 16 arrays for the separate modules) instead of the combiner. So it's best not to suggest that one specific format is the 'online' format. But of course we can document that it's similar to a common setup for online data.

extra_data/export.py
```diff
@@ -242,7 +254,24 @@ def serve_files(path, port, source_glob='*', key_glob='*', **kwargs):
...
     for tid, train_data in data.trains():
         if train_data:
-            streamer.feed(train_data)
+            if data_like == 'file':
+                streamer.feed(train_data)
+            elif data_like == 'online':
+                stacked = stack_detector_data(train_data, 'image.data')
+                merged_data = {}
+                # the data key pretends this is online data from SPB
+                merged_data['SPB_DET_AGIPD1M-1/CAL/APPEND_CORRECTED'] = {
```
takluyver: This should probably select the appropriate source names for raw & corrected data. Is it also meant to handle other detectors (DSSC, LPD, JUNGFRAU)?

extra_data/export.py
```diff
@@ -242,7 +254,24 @@ def serve_files(path, port, source_glob='*', key_glob='*', **kwargs):
     streamer.start()
     for tid, train_data in data.trains():
         if train_data:
-            streamer.feed(train_data)
+            if data_like == 'file':
+                streamer.feed(train_data)
+            elif data_like == 'online':
+                stacked = stack_detector_data(train_data, 'image.data')
```
takluyver: I think this will fail if you have any other sources selected. That's not necessarily a problem, but we should document it.

Or we could make it so that the option to stack detector modules automatically selects only the detector source from the files. But that would need to be documented too.

extra_data/export.py
```diff
@@ -242,7 +254,24 @@ def serve_files(path, port, source_glob='*', key_glob='*', **kwargs):
...
+                merged_data = {}
+                # the data key pretends this is online data from SPB
+                merged_data['SPB_DET_AGIPD1M-1/CAL/APPEND_CORRECTED'] = {
+                    'image.data': stacked
+                }
+                sec, frac = gen_time()
```
takluyver: Not part of this PR, but this reminds me: we have timestamps in files now, the streaming could use them...

extra_data/export.py
dallanto: No problem. In the first draft I took over from the karabo-bridge-server-sim implementation, we can discuss tomorrow how to adapt and generalize this in the best way.

extra_data/export.py
dallanto: My choice of source name and method corresponds to the current use-case: AGIPD appender data from the SPB online calibration pipeline. It is desirable to extend and thus generalize this, but maybe we can think of a short-term solution which is, as you say, documented.  
The alternative would be that we offer a tentative version to be installed from the branch. Which is probably better than a "unsupervised" (not version controlled etc.) script like that currently used for testing.

extra_data/export.py
dallanto: Pls. see my comment below. At the moment the use case is very "OnDA-centric".

extra_data/export.py
```diff
@@ -216,7 +217,15 @@ def feed(self, data, metadata=None):
...
-def serve_files(path, port, source_glob='*', key_glob='*', **kwargs):
+def gen_time():
+    """Create a mock timestamp for meta-data"""
+    timestamp = time()
+    sec, frac = str(timestamp).split('.')
+    frac = frac.ljust(18, '0') # attosecond resolution
+    return sec, frac
+
+
+def serve_files(path, port, data_like='file', source_glob='*', key_glob='*', **kwargs):
```
tmichela: can you add the new keyword argument at the end, to keep order intact?

extra_data/export.py
```diff
@@ -225,6 +234,9 @@ def serve_files(path, port, source_glob='*', key_glob='*', **kwargs):
         Path to the HDF5 file or file folder.
     port: int
         Local TCP port to bind socket to.
+    data_like: str
```
tmichela: this should be also changed if you change the cli argument name

extra_data/export.py
tmichela: The use case for the server sim is a bit different. The data sent there is always a stack of all modules, only the shape changes.

extra_data/export.py
tmichela: Do you think it would be too much effort to generalize it?
We could try to use the Detector classes form `extra_data.components`, would that work?

extra_data/export.py
dallanto: Done

extra_data/export.py
dallanto: Done

extra_data/export.py
dallanto: I changed the name to ``--detector-modules`` and the choices to ``file`` (default) and ``stack``. While this is not a generalization yet, it stresses more clearly what this particular option is for.

tmichela: Nitpicking, but @takluyver suggestion is clearer (`append-detector-module`) as it indicate that all modules are aggregated in a single array. I'd also suggest to not have choices, they don't really make sens in that context. Just the option without argument should be enough (present: modules are aggregated, not present: each module is a separate datasource)

dallanto: O.k. I see your point

takluyver:
I'm OK with merging this as AGIPD-specific code to be generalised later, but I do think that the limitation should be documented, and preferably it should produce a clear error/warning if you try to use it with other data.

dallanto:
> I'm OK with merging this as AGIPD-specific code to be generalised later, but I do think that the limitation should be documented, and preferably it should produce a clear error/warning if you try to use it with other data.

Thanks for the in-principle-LGTM. Yes, I take your advice serious and will take care about the suggested additions 1. warning 2. documentation. I think the merging can wait until that (tomorrow, I hope).

dallanto:
Hi Thomas, do you think the minimal "documentation" via parser-help is sufficient for the moment?  
I'd also be happy to add to the extra-data documentation, where the ZMQ streaming section mentions the command-line tool (but the corresp. framework port is much more verbose).  
If could add all optional/keyword arguments, i.e. --source, --key and --append-detector-modules to the command-line tool paragraph.

extra_data/export.py
```diff
@@ -258,6 +295,13 @@ def main(argv=None):
         "--key", help="Stream only matching keys ('*' is a wildcard)",
         default='*',
     )
+    ap.add_argument(
+        "--append-detector-modules", help="combine multiple module sources \
+           into one (will only work for AGIPD data currently).",
```
takluyver:
```suggestion
        "--append-detector-modules", help="combine multiple module sources "
        "into one (will only work for AGIPD data currently).",
```

This is a neater way of splitting strings over multiple lines (it avoids leaving a big gap in the resulting string).

extra_data/export.py
```diff
@@ -242,7 +255,31 @@ def serve_files(path, port, source_glob='*', key_glob='*', **kwargs):
     streamer.start()
     for tid, train_data in data.trains():
         if train_data:
-            streamer.feed(train_data)
+            if append_detector_modules:
+                if source_glob == '*':
+                    warn(' You are trying to stack detector-module sources \
+with a global windcard (\'*\'). If there are non-detector sources in your \
+run, this will fail.\n')
```
takluyver:
```suggestion
                    warn("You are trying to stack detector-module sources "
                         "with a global windcard ('*'). If there are non-detector "
                         "sources in your run, this will fail.")
```

extra_data/karabo-bridge-stack-files
```diff
@@ -0,0 +1,61 @@
+#!/gpfs/exfel/sw/software/xfel_anaconda3/1.1/bin/python
```
takluyver: Can you remove this file and the extra README?

extra_data/export.py
```diff
@@ -242,7 +255,31 @@ def serve_files(path, port, source_glob='*', key_glob='*', **kwargs):
...
+                }
+                # add other sources (by key) here?
+                sec, frac = gen_time() # use time-stamps from file data?
+                metadata = {}
+                metadata['SPB_DET_AGIPD1M-1/CAL/APPEND_CORRECTED'] = {
+                    'source': 'SPB_DET_AGIPD1M-1/CAL/APPEND_CORRECTED',
+                    'timestamp.tid': tid,
+                    'timestamp.sec': sec,
+                    'timestamp.frac': frac
+                }
```
takluyver: I've just noticed that the generated timestamps for other sources are conditional on a `dummy_timestamps=True` parameter to `ZMQStreamer`. I think we should keep that consistent, and not generate timestamps here.

Possibly that means that the `karabo-bridge-serve-files` command needs a `--dummy-timestamps` option as well so that it can produce data suitable for OnDA.

takluyver:
For documentation, I think ideally the streaming page should link to the command-line tools reference page (https://extra-data.readthedocs.io/en/latest/cli.html ), and we'd document the options there.

dallanto:
Your comments came in tn the same minute when I pushed a commit with additions to the streaming.rst docs page. But no problem.

takluyver:
Sorry about that, I should have been a bit quicker to respond to your comment from 2 days ago.

dallanto:
> Sorry about that, I should have been a bit quicker to respond to your comment from 2 days ago.

No problem at all : )

extra_data/karabo-bridge-stack-files
dallanto: Yes I will remove this, you already suggested that some time ago ...

extra_data/export.py
dallanto: Yes, that's better indeed

extra_data/export.py
dallanto: Same here, thanks

extra_data/export.py
dallanto: Yes, that's doable

dallanto:
Apart from passing the CI unit tests, I also tested on Maxwell:  
- the bridge glimpse makes sense  
- OnDA is happy  
I had my doubts, because I create a new (empty) metadata dict for the AGIPD source, now without timestamps, but it seems the feed method just adds that to the streamer, and if dummy timestamps are already there via constructor, that's fine.

docs/streaming.rst
```diff
@@ -13,6 +13,9 @@ To stream the data from a file or run unmodified, use the command::
 The number (4545) must be an unused TCP port above 1024. It will bind to
 this and stream the data to any connected clients.
 
+Keyword-argument options are explained on the `command-line tools <cli.html>`__
```
takluyver: Nitpicking: I'd call these command-line options.

We can also use a Sphinx cross reference to the specific section:

```rst
:ref:`cmd-serve-files`
:ref:`command reference <cmd-serve-files>`
```

The top form uses the heading text as the link, the second one lets you define the link text.

extra_data/export.py
```diff
@@ -216,7 +218,17 @@ def feed(self, data, metadata=None):
         self._buffer.put(self._serialize(data, metadata))
 
 
-def serve_files(path, port, source_glob='*', key_glob='*', **kwargs):
+def gen_time():
+    """Create a mock timestamp for meta-data"""
+    timestamp = time()
+    sec, frac = str(timestamp).split('.')
+    frac = frac.ljust(18, '0') # attosecond resolution
+    return sec, frac
```
takluyver: We should either remove this function again, or make use of it in ZMQStreamer.

extra_data/karabo-bridge-stack-files
takluyver: don't forget this :wink:

docs/cli.rst
```diff
@@ -42,6 +42,41 @@ Stream data from files in the `Karabo bridge
 <https://in.xfel.eu/readthedocs/docs/data-analysis-user-documentation/en/latest/online.html#data-stream-to-user-tools>`_
 format. See :doc:`streaming` for more information.
 
+.. code-block:: shell
+
+   karabo-bridge-serve-files /gpfs/exfel/exp/XMPL/201750/p700000/proc/r0005 4321
+
+.. program:: karabo-bridge-serve-files
+
+.. option:: --source <source>
```
takluyver: I think we can also have `.. option::` directives for the positional arguments (run & port) if you want. Not a high priority, though.

http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-option

extra_data/export.py
dallanto: Removed

docs/streaming.rst
dallanto: I changed to the 2nd form

extra_data/karabo-bridge-stack-files
dallanto: Doh, my bad ... but now finally removed, together with the tentative README_file_Stack.md

takluyver:
Thanks @dallanto, LGTM.

dallanto:
Thanks @takluyver   
Another remark: I think for flexibility reasons it is better to have it this way, with two bool keywords (append, dummy_timestamps) from the command-line, even if looking at the OnDA use case only, one could have simply called the streamer constructor with dummy-timestamps True always (when appending is True) and thus saved a command-line option.

takluyver:
Yup, I agree. Hopefully we can also figure out sending the timestamps from files by default; dummy timestamps were off by default because I didn't want to send fake information unless it was explicitly requested.

PR #42, opened 2020-03-17
Check if firsts dataset has any entry

tmichela:
Validation failed on some files when a dataset has not data at all. Example: `/gpfs/exfel/exp/FXE/202030/p900121/raw/r0300/RAW-R0300-JNGFR01-S00000.h5`

At the moment it skips completely the contiguity check for this dataset, but I guess that at some point we should think of adding functionalities to feedback this kind of cases to the users. It is not an error as the file is readable, but knowing that a data source hasn't recorded anything is probably useful information.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1141,12 +1141,17 @@ class DataCollection:
         """Show information about the selected data.
         """
         # time info
-        first_train = self.train_ids[0]
-        last_train = self.train_ids[-1]
-        train_count = len(self.train_ids)
-        seconds, deciseconds = divmod((last_train - first_train + 1), 10)
-        span_txt = '{}.{}'.format(datetime.timedelta(seconds=seconds),
-                                  int(deciseconds))
+        if len(self.train_ids) == 0:
+            first_train = last_train = '-'
+            train_count = 0
+            span_txt = '0.0'
+        else:
+            first_train = self.train_ids[0]
+            last_train = self.train_ids[-1]
+            train_count = len(self.train_ids)
+            seconds, deciseconds = divmod((last_train - first_train + 1), 10)
+            span_txt = '{}.{}'.format(datetime.timedelta(seconds=seconds),
+                                      int(deciseconds))
 
         detector_modules = {}
         for source in self.detector_sources:
--- a/extra_data/tests/test_reader_mockdata.py
+++ b/extra_data/tests/test_reader_mockdata.py
@@ -629,9 +629,4 @@ def test_permission():
 
 def test_empty_file_info(mock_empty_file, capsys):
     f = H5File(mock_empty_file)
-    f.info()
-    out, err = capsys.readouterr()
-    assert '# of trains:    0' in out
-    assert 'Duration:       0.0' in out
-    assert 'First train ID: -' in out
-    assert 'Last train ID:  -' in out
+    f.info()  # smoke test
--- a/extra_data/tests/test_reader_mockdata.py
+++ b/extra_data/tests/test_reader_mockdata.py
@@ -625,3 +625,13 @@ def test_permission():
         run = RunDirectory(d)
     assert "Permission denied" in str(excinfo.value)
     assert d in str(excinfo.value)
+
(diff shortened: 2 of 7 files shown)


takluyver:
I'd expect such a file to have train IDs, and an index containing all zeros, so I'd be inclined to make validate complain if `INDEX/trainId` has zero length. But that can be a separate question.

The implementation LGTM, but now that we've got some decent tests for validation, do you want to add one for this case? :wink:

tmichela:
> I'd expect such a file to have train IDs, and an index containing all zeros, so I'd be inclined to make validate complain if INDEX/trainId has zero length. But that can be a separate question.

The DAQ probably waits the first update to start recording any trainId, to avoid trailing empty data-sets at the beginning of each runs. I don't think this behavior will change, so we should make sure that extra-data does not brake if that happens (should not be too hard I hope)

tmichela:
I've added a test, also made sure `DataCollection.info` does not break if `train_ids` is empty.

extra_data/tests/test_reader_mockdata.py
```diff
@@ -625,3 +625,13 @@ def test_permission():
...
+
+
+def test_empty_file_info(mock_empty_file_valid, capsys):
+    f = H5File(mock_empty_file_valid)
+    f.info()
+    out, err = capsys.readouterr()
+    assert '# of trains:    0' in out
+    assert 'Duration:       0.0' in out
+    assert 'First train ID: -' in out
+    assert 'Last train ID:  -' in out
```
takluyver: I'm not super keen on testing human-readable messages like this, because I don't think we want to guarantee the particular formatting. But so long as there's only a couple such tests, it's not a big deal.

extra_data/reader.py
```diff
@@ -1141,12 +1141,17 @@ def info(self):
...
-        span_txt = '{}.{}'.format(datetime.timedelta(seconds=seconds),
-                                  int(deciseconds))
+        if len(self.train_ids) == 0:
+            first_train = last_train = '-'
+            train_count = 0
+            span_txt = '0.0'
+        else:
+            first_train = self.train_ids[0]
+            last_train = self.train_ids[-1]
+            train_count = len(self.train_ids)
```
takluyver:
```suggestion
        train_count = len(self.train_ids)
        if train_count == 0:
            first_train = last_train = '-'
            span_txt = '0.0'
        else:
            first_train = self.train_ids[0]
            last_train = self.train_ids[-1]
```

extra_data/tests/test_validation.py
```diff
@@ -26,6 +26,14 @@ def data_aggregator_file():
         yield path
 
 
+@fixture(scope='function')
+def mock_empty_file():
+    with TemporaryDirectory() as td:
+        path = osp.join(td, 'RAW-R0450-DA02-S00001.h5')
+        make_examples.make_sa3_da_file(path, ntrains=0)
+        yield path
```
takluyver: Duplicate of the fixture in `conftest.py`?

extra_data/tests/test_reader_mockdata.py
tmichela: what would you suggest instead? Only check the method does not break?

extra_data/tests/test_validation.py
tmichela: Yep, I'll fix that.

extra_data/tests/test_reader_mockdata.py
takluyver: That would be my starting point. And then maybe minimal testing of specific bits of the output that are particularly important, or particularly likely to go wrong.

Aside: in case you don't already know it, there's a nice term for tests that only check that something doesn't error: smoketests. Push the button, see if smoke comes out. :slightly_smiling_face:

tmichela: :+1:

takluyver:
Thanks, LGTM. :+1:

tmichela:
Thanks!

PR #43, opened 2020-03-19
Allow selection of sources by any iterable of glob patterns

philsmt:
Currently, `DataCollection.select` only works exactly for a single string, a list of tuples or a dictionary. This excludes other forms of iterables, e.g. sets or generators. This PR checks for the abstract base class `Iterable` rather than `list` to extend it to any iterable of glob patterns. The other two interfaces of `select` remain unchainged.

Given that strings and dicts branch out before, only sensible iterables should remain here. The official docs [suggest ](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable) an alternative way of checking for an iterable via try...iter...except, but this would only suceed additionally for weird custom types having \_\_getitem__, yet not \_\_iter__.

I fixed the docstring of `select` while I was at it and found an amusing bug actually that prevented a `TypeError` from being raised if any non-dict, list or string is passed.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -901,7 +901,7 @@ class DataCollection:
                 for (src_glob, key_glob) in selection
             )
         else:
-            TypeError("Unknown selection type: {}".format(type(selection)))
+            raise TypeError("Unknown selection type: {}".format(type(selection)))
 
         return dict(res)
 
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, the implementation LGTM.

Do you want to add/modify some test to select with an iterable other than a list?

philsmt:
I thought about tests initially, but discarded it as the selecting itself didn't change at all, just the type check. You think it's necessary?

takluyver:
Not really, it's just a reflex to ask people to write tests, I think. :wink:

Issue #45, opened 2020-04-02
Filecache tests failing intermittently


takluyver:
cc @egorsobolev 

The new tests from PR #25 failed in [one recent CI run](https://travis-ci.org/github/European-XFEL/EXtra-data/jobs/670164016). It was only the Python 3.7 job that failed, but I suspect this is a random failure rather than something related to the Python version.

Here are the failure details:

```
_____________________________ test_filecache_large _____________________________

mock_spb_raw_run = '/tmp/tmp3z89k4l6'
filecache_512 = <extra_data.filecache.FileCache object at 0x7f27749dcc50>

    def test_filecache_large(mock_spb_raw_run, filecache_512):
        fc = filecache_512
    
        files = [os.path.join(mock_spb_raw_run, f) \
                 for f in os.listdir(mock_spb_raw_run) if f.endswith('.h5')]
        run = DataCollection.from_paths(files)
    
        trains_iter = run.trains()
        tid, data = next(trains_iter)
        assert tid == 10000
        device = 'SPB_IRU_CAM/CAM/SIDEMIC:daqOutput'
        assert device in data
        assert data[device]['data.image.pixels'].shape == (1024, 768)
>       assert len(fc._cache) == len(run.files)
E       assert 17 == 18
E         -17
E         +18

extra_data/tests/test_filecache.py:42: AssertionError
_____________________________ test_filecache_small _____________________________

mock_spb_raw_run = '/tmp/tmp3z89k4l6'
filecache_3 = <extra_data.filecache.FileCache object at 0x7f27749dcc50>

    def test_filecache_small(mock_spb_raw_run, filecache_3):
        fc = filecache_3
    
        files = [os.path.join(mock_spb_raw_run, f) \
                 for f in os.listdir(mock_spb_raw_run) if f.endswith('.h5')]
        run = DataCollection.from_paths(files)
        trains_iter = run.trains()
    
        for i in range(3):
>           tid, data = next(trains_iter)

extra_data/tests/test_filecache.py:56: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
extra_data/reader.py:1406: in __iter__
    yield tid, self._assemble_data(tid)
extra_data/reader.py:1397: in _assemble_data
    source_data[key] = ds[first : first + count]
[...]

PR #46, opened 2020-04-07
Fix unwanted file cache behavior caused by delayed garbage collection

egorsobolev:
Fix the cause #45  
- linking of `FileAccess` instance to the global `FileCache` instance at the creation time
- functions to get/set global `FileCache` instance
- in the test, correct estimation of the number of files in cache independent on files order in `DataCollection`

In addition, switching to weakref:
- switch `FileCache` to the usage of `weakref`
- explicit closing of files in `DataCollection` destructor

@takluyver could you please take a look?

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -42,7 +42,7 @@ from .read_machinery import (
     find_proposal,
 )
 from .run_files_map import RunFilesMap
-from .filecache import get_global_filecache
+from . import filecache
 
 __all__ = [
     'H5File',
@@ -85,7 +85,6 @@ class FileAccess:
 
     def __init__(self, filename, _cache_info=None):
         self.filename = filename
-        self.__fc = get_global_filecache()
 
         if _cache_info:
             self.train_ids = _cache_info['train_ids']
@@ -111,9 +110,9 @@ class FileAccess:
     @property
     def file(self):
         if self.__file:
-            self.__fc.touch(self.filename)
+            filecache.extra_data_filecache.touch(self.filename)
         else:
-            self.__file = self.__fc.open(self.filename)
+            self.__file = filecache.extra_data_filecache.open(self.filename)
             
         return self.__file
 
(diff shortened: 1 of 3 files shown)


extra_data/reader.py
```diff
@@ -354,6 +356,15 @@ def initializer():
...
         return cls(files, ctx_closes=True)
+    
+    def _close_all_files(self):
+        # Close the files if this collection was created by opening them.
+        if self.ctx_closes:
+            for file in self.files:
+                file.close()
+    
+    def __del__(self):
+        self._close_all_files()
```
takluyver, line 367: While running the tests for PR #47, I saw a load of errors from `__del__` methods running at Python shutdown. I'm not exactly sure how to reproduce that, but I know such problems are fairly common with `__del__` methods. I'd really like to avoid `__del__` methods wherever possible, and let normal object cleanup happen.

I'll try to have a look into the code relating to this myself and see what I can do, so hold off on this for the moment.

extra_data/tests/test_filecache.py
```diff
@@ -32,17 +26,29 @@ def test_filecache_large(mock_spb_raw_run, filecache_512):
...
+    cnt, fix = {}, {}
+    for f in run.files:
+        a, b = os.path.basename(f.filename).rsplit('-',1)
+        try:
+            cnt[a] += 1
+        except KeyError:
+            cnt[a] = 1
+        if b == 'S00000.h5':
+            fix[a] = cnt[a] 
+    nfiles = sum(a for a in fix.values())
```
takluyver, line 40: I think we should probably just drop the `len(fc._cache)` check instead of adding all this logic, or make it more vague (e.g. `>= 17`). Going to all this trouble to figure out what the number should be seems like it would tie the test to the implementation details.

extra_data/reader.py
egorsobolev: Unfortunately it is absolutely necessary here. Without it the files stays open in the file cache after destroying `Data Collection` instance until garbage collection which can be delayed indefinitely. (there is check on it in tests, and they fail without this `__del__`)

In opposite, the `__del__` just closes files (calls `FileAccess.close()` which just assigns `None` to the `__file` attribute). I don't think it can raise exception ever on shutdown time.

extra_data/tests/test_filecache.py
egorsobolev: Yes it would. But this is also a reminder of the specifics of the implementation. I spent half a day to figure this out.
As an option, we can sort the files and make the implementation deterministic. In `DataCollection.from_paths()` replace:
```python
    return cls(files, ctx_closes=True)
```
with
```python
    return cls(sorted(files, key=lambda f: f.filename), ctx_closes=True)
```
I prefer the second option. What do you think?

takluyver:
Following discussion with @philsmt, we've merged the alternative approach in #48.

PR #47, opened 2020-04-08
Write selected detector frames to a new file

takluyver:
This adds a method on the big detector data interfaces, e.g. `AGIPD1M.write_frames()`. You pass a destination filename, and two arrays of train IDs and pulse IDs (i.e. the columns from a candidate frame list). It writes out the selected frames as a new EuXFEL-format HDF5 file, along with all of the once-per-train detector information.

The 16 detector modules remain as separate sources, but in a common file. The resulting file can be opened again in EXtra-data.

cc @dallanto and @egorsobolev

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -472,6 +472,9 @@ class FramesFileWriter(FileWriter):
         We do this as a separate step so the contents of the file are defined
         together before the main data.
         """
+        if source not in self.data.instrument_sources:
+            return super().prepare_source(source)
+
         for key in sorted(self.data.keys_for_source(source)):
             path = 'INSTRUMENT/{}/{}'.format(source, key.replace('.', '/'))
             if key.startswith('image.'):
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -523,15 +523,18 @@ class FramesFileWriter(FileWriter):
             frame_tids_piecewise.append(file_tids[ixs])
 
         frame_tids = np.concatenate(frame_tids_piecewise)
-        self._make_instrument_index(source, 'image', frame_tids)
+        self._make_index(source, 'image', frame_tids)
 
-    def copy_instrument_source(self, source):
+    def copy_source(self, source):
         """Copy all the relevant data for one detector source"""
+        if source not in self.data.instrument_sources:
+            return super().copy_source(source)
+
         all_keys = self.data.keys_for_source(source)
         img_keys = {k for k in all_keys if k.startswith('image.')}
 
         for key in sorted(all_keys - img_keys):
-            self.copy_instrument_dataset(source, key)
+            self.copy_dataset(source, key)
 
         self.copy_image_data(source, sorted(img_keys))
 
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -1,5 +1,6 @@
 """Interfaces to data from specific instruments
 """
+import h5py
 import logging
 import numpy as np
 import pandas as pd
@@ -426,15 +427,29 @@ class MPxDetectorBase:
         VirtualCXIWriter(self).write(filename)
 
     def write_frames(self, filename, trains, pulses):
-        writer = FileWriter(filename, self.data)
+        if (trains.ndim != 1) or (pulses.ndim != 1):
+            raise ValueError("trains & pulses must be 1D arrays")
+        inc_tp_ids = zip_trains_pulses(trains, pulses)
 
-        for modno, source in sorted(self.modno_to_source.items()):
-            for key in sorted(self.data.keys_for_source(source)):
-                if key.startswith('image.'):
-                    # select frames
-                    pass
-                else:
-                    writer.copy_instrument_dataset(source, key)
+        writer = FileWriter(filename, self.data, inc_tp_ids)
+        writer.write()
+        writer.file.close()
+
+
+def zip_trains_pulses(trains, pulses):
+    """Combine two similar arrays of train & pulse IDs as one struct array
+    """
+    if trains.shape != pulses.shape:
+        raise ValueError(
+            f"Train & pulse arrays don't match ({trains.shape} != {pulses.shape})"
+        )
+
+    res = np.zeros(trains.shape, dtype=np.dtype([
+        ('trainId', np.uint64), ('pulseId', np.uint64)
+    ]))
+    res['trainId'] = trains
+    res['pulseId'] = pulses
+    return res
 
 
 class FileWriter:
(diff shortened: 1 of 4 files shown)


extra_data/writer.py
```diff
@@ -34,19 +64,24 @@ def _make_control_index(self, source, data_tids):
...
+
     def _make_instrument_index(self, source, key, data_tids):
+        """Create & store an index for an instrument data group"""
         index_path = source + '/' + key.partition('.')[0]
         if index_path not in self.indexes:
             self.indexes[index_path] = self._generate_index(data_tids)
-            self.data_sources.add('INSTRUMENT/' + index_path)
 
     def _generate_index(self, data_tids):
         """Convert an array of train IDs to first/count for each train"""
```
egorsobolev, line 87: You copy yourself here.

These three methods with ***instrument*** in the middle of name and `_generate_index` almost verbatimly repeat previous three methods with ***control*** in the middle. Difference in datasource path, `index_path` and following `assert`. There is one more `assert` in `copy_control_dataset` which looks weired.

Could it better to generalize this code with one or two `if` statements?

Or even implement an academical OOP solution: make `BaseDataset` and two children `InstrumentDataset`, `ControlDataset` and move all this methods to the `BaseDataset`with minimal re-implementation in children. And build `FileWriter.data_sources` set using these classes. Then the FileWriter becames beutifully extandable on image dataset also.

extra_data/components.py
```diff
@@ -425,6 +426,112 @@ def write_virtual_cxi(self, filename):
...
+            key=lambda fa: fa.train_ids[0]
+        )
+
+        for fa in src_files:
+            _, counts = fa.get_index(source, 'image')
+            file_tids = np.repeat(fa.train_ids, counts.astype(np.intp))
+            file_pids = fa.file[f'/INSTRUMENT/{source}/image/pulseId'][:]
+            if file_pids.ndim == 2 and file_pids.shape[1] == 1:
+                # Raw data has a spurious extra dimension
+                file_pids = file_pids[:, 0]
```
egorsobolev, line 502: `numpy.squeeze`?

extra_data/components.py
```diff
@@ -425,6 +426,112 @@ def write_virtual_cxi(self, filename):
...
+        self.inc_tp_ids = inc_tp_ids
+
+    def prepare_source(self, source):
+        """Prepare all the datasets for one source.
+
+        We do this as a separate step so the contents of the file are defined
+        together before the main data.
+        """
+        for key in sorted(self.data.keys_for_source(source)):
+            path = 'INSTRUMENT/{}/{}'.format(source, key.replace('.', '/'))
```
egorsobolev, line 476: It seems that you don't expect CONTROL source here. I didn't check but I believe that is true if it is instantiated from `MPxDetectorBase.write_frames`. But what if user construct it by himself?

Would it better to keep functionality for CONTROL here? In normal case it will not work but will keep code safe.

takluyver: Good point, I've added a check to pass control sources through to the parent class.

extra_data/writer.py
takluyver: Thanks, I've followed your suggestion to combine some methods and separate pieces with `if` statements where necessary. I haven't tried to reorganise it to a more OOP approach at the moment.

extra_data/components.py
takluyver: I prefer to use an explicit check where it's practical, because if a 'real' dimension happens to have length 1, plain `np.squeeze(x)` will discard that as well.

You can tell it to squeeze a specific axis, but in our case we'd need to ensure that axis exists first.

takluyver:
The `/pr` CI which merges with master passes, but the `/push` CI doesn't, because of a streaming issue that's fixed in master.

extra_data/components.py
```diff
@@ -425,6 +426,118 @@ def write_virtual_cxi(self, filename):
...
+        """
+        if source not in self.data.instrument_sources:
+            return super().prepare_source(source)
+
+        for key in sorted(self.data.keys_for_source(source)):
+            path = 'INSTRUMENT/{}/{}'.format(source, key.replace('.', '/'))
+            if key.startswith('image.'):
+                nentries = 0  # dataset grows later
+            else:
+                nentries = self.data.get_data_counts(source, key).sum()
```
egorsobolev, line 483: It seem you need override only this part in entire method. Why do not put this in class member, like `_get_number_of_data_entries(source, key)`?

takluyver: I could do that, but I think it makes an uncomfortably tight coupling between the subclass and its parent. Returning 0 from `_get_number_of_data_entries()` would be a convenient lie - it's not the real number of entries, it's just useful to pretend it is when we set up the dataset.

So using that method in any other context would go wrong, but only when you're using the subclass. If you're just looking at the parent class, it would seem entirely reasonable to use `_get_number_of_data_entries()` in different contexts.

So, IMO, duplicating a few lines in this method is preferable to that. Another option would be to find the real number of entries for image data, but that means extra complexity as you handle all of the pulseId data again.

egorsobolev: Yes, the name of method may confuse. Call it like _guess_number_of_storing_entries() instead or any similar way. It will precisely define a possible context.

I don't want to say that any one line could never be repeated. But the good portion of logic is repeated here. Any future developers have to remember that they must change it in two places.

I also don't see too tight coupling. Suggested interface will be introduced in base class FileWriter as:
```python
def _guess_number_of_entries_to_record(source, key):
    return self.data.get_data_counts(source, key).sum()
```
Usually you don't need to redefine anything in children. When it necessary, you have an option to change `_guess_number_of_entries_to_record` at any children like here in FramesFileWriter:
```python
def _guess_number_of_entries_to_record(this, source, key):
    if key.startswith('image.'):
        return 0
    else:
        return super()._guess_number_of_entries_to_record(source, key)
```
keeping prepare_source() unchanged.

Or get rid of this hook at all by redefining prepare_source().

Looks very flexible.

Actually, IMO, the ideal solution is different. It would be good having an option to create a collection which exposes part of the real data filtered by the list of frames. Then no special recorder is needed. But implementation may be too hard at the moment.

So, I think this is a little, cheap and good improvement. Since it works as is you may merge it without changes if I haven't managed to convince you.

takluyver: Fair enough, I'm persuaded. I've attributed you on the commit, as I copied it from the code in your commen.t

egorsobolev:
I found very small peaces of the code which repeat and, I think, relate to subelemetns of DataCollecton instead of DataCollection itself or these 'FileWriters'. We discussed these new classes in #55 KeyData and SourceData.

We could isolate them in order of preparation for the future changes as:
```python
def _sourcedata_path(self, source, key, index_path=False):
    if source in self.data.instrument_sources:
       sec, path = "INSTRUMENT/", f"{source}/{key.partition('.')[0]}"
    else:
       sec, path = "CONTROL/", source
    return ("" if index_path else sec) + path

def _keydata_path(self, source, key):
    return f"{self._section(source)}/{source}/{key.replace('.', '/')}"
```
It could be used right now like
```python
self.data_sources.add(self._sourcedata_path(source, key))
```
in couple of places.

What do you think?

takluyver:
I tried putting this in and using it, but I felt the code got less clear, because path construction that you could previously see exactly what it was doing was delegating to a method somewhere else. Partly this is a challenge of naming: 'source' and 'path' can each refer to a couple of slightly different things, and we don't have good terms to distinguish those meanings.

I agree that we'll be able to simplify some of this as we introduce additional abstractions to represent keys & sources. But for now, I don't think it's worth trying to anticipate that.

Thanks for the review, I'll merge this one now. The plan is to wait for @dallanto's upcoming PR about JUNGFRAU integration, as well as #51, before making a new release, but let me know if you'd like a release including this sooner.

PR #50, opened 2020-04-17
refactor ZMQStreamer

tmichela:
- use `karabo_bridge.serializer`
- support `PUSH`, `PUB` sockets
- allow usage of infiniband interface


tmichela:
Now this looks like it would make more sens if it was part of `karabo-brige-py`, event the cli tool is called `karabo-bridge-serve-files`...

extra_data/export.py
```diff
@@ -10,171 +10,118 @@
...
+        protocol_version: ('1.0' | '2.1')
+            Which version of the bridge protocol to use. Defaults to the latest
+            version implemented.
+        dummy_timestamps: bool
+            Some tools (such as OnDA) expect the timestamp information to be in
+            the messages. We can't give accurate timestamps where these are not
+            in the file, so this option generates fake timestamps from the time
+            the data is fed in.
+        use_infiniband: bool
+            Use infiniband interface if the host has one (default False)
```
takluyver, line 66: It looks like this parameter doesn't actually affect anything at present?

extra_data/export.py
```diff
@@ -10,171 +10,118 @@
...
+                continue
+
+            if events[self.server_socket] is zmq.POLLIN:
+                msg = self.server_socket.recv()
+                if msg != b'next':
+                    print(f'Unrecognised request: {msg}')
+                    self.server_socket.send(b'Error: bad request %b' % msg)
+                    continue
+
+            self.server_socket.send_multipart(payload, copy=False)
```
takluyver, line 122: It looks like much of this class is duplicated from karabo-bridge-py. Is it practical to make more reusable alongside the serialisation code? Not as part of this PR, just a more general question.

(The answer might be that there's enough difference in the details that it's easier to keep them independent - it just looks familiar enough to be worth considering.)

extra_data/tests/test_streamer.py
```diff
@@ -53,93 +53,32 @@ def compare_nested_dict(d1, d2, path=''):
...
-
-    TODO: avoid using a private method from extra_data for tests.
-    """
-
-    def __init__(self, data):
-        self.bytes = data
-        self.buffer = data
-
+@pytest.fixture(scope='function')
+def server(protocol_version):
```
takluyver, line 57: I believe it's possible to parameterise a fixture, if all of the tests which use it should try the same set of possible parameters.

extra_data/export.py
```diff
@@ -183,44 +130,43 @@ def feed(self, data, metadata=None):
...
-                  'timestamp.tid': 1234567890  # int
-              }
-
-            If the metadata dict is not provided it will be extracted from
-            'data' or an empty dict if 'metadata' key is missing from a data
-            source.
+            Contains train metadata. If the metadata dict is not provided it
+            will be extracted from 'data' or an empty dict if 'metadata' key
+            is missing from a data source.
+            see :func:`~karabo_bridge.serializer.serialize` for details
```
takluyver, line 139: If we put this function in Sphinx docs somewhere, we can configure [Intersphinx](https://docs.readthedocs.io/en/stable/guides/intersphinx.html) to cross-link to it.

In either case, I think we should drop the `~`, which shows only the last part of the name in the docs. Showing the full name makes it clear that it's in a separate package.

extra_data/export.py
```diff
@@ -183,44 +130,43 @@ def feed(self, data, metadata=None):
...
+
+        block: bool
+            If True, block if necessary until a free slot is available or
+            'timeout' has expired. If False and there is no free slot, raises
+            'queue.Full' (timeout is ignored)
+
+        timeout: float
+            In seconds, raises 'queue.Full' if no free slow was available
```
takluyver, line 147:
```suggestion
            In seconds, raises 'queue.Full' if no free slot was available
```

takluyver:
Oh, I just saw your extra comment. Yes, maybe it would make sense to put `ZMQStreamer` in karabo-bridge-py. Then it doesn't matter much which project contains the loop over `data.trains()` and the CLI.

extra_data/export.py
tmichela: Right. The documentation for the karabo bridge project isn't actually self contained (and outdated in user-documentation...). I'll look if that makes sens to change that.

extra_data/tests/test_streamer.py
```diff
@@ -6,140 +6,26 @@
...
-        server_2_2.feed(DATA)
+    for _ in range(3):
+        server.feed(data)
 
-    for i in range(3):
-        data, metadata = client.next()
-        compare_nested_dict(data, DATA)
+    for _ in range(3):
+        d, metadata = client.next()
+        assert d == data
```
takluyver, line 28: I think all the functionality this is testing is now in `karabo_bridge`, right?

extra_data/export.py
```diff
@@ -12,215 +12,32 @@
...
 from argparse import ArgumentParser
 import os.path as osp
-from queue import Queue
-from threading import Event, Thread
-from time import time
 from warnings import warn
 
-import msgpack
-import numpy as np
-import zmq
```
takluyver, line 22: I guess we can also drop the direct requirements on msgpack & pyzmq in `setup.py`, though they'll still be installed as dependencies of karabo bridge.

pytest.ini
```diff
@@ -1,2 +1,3 @@
 [pytest]
 addopts = --ignore docs/xpd_examples.ipynb --ignore docs/xpd_examples2.ipynb --ignore docs/parallel_example.ipynb --ignore docs/dask_averaging.ipynb --ignore docs/inspection.ipynb --ignore docs/iterate_trains.ipynb
+filterwarnings = ignore::DeprecationWarning
```
takluyver, line 3: I think tests should make deprecation warnings visible in general. If we need to suppress specific ones, there are ways to temporarily filter warnings, filter warnings from specific modules, and even assert that [specific warnings are triggered](https://docs.pytest.org/en/latest/warnings.html#ensuring-code-triggers-a-deprecation-warning).

extra_data/tests/test_streamer.py
tmichela: yes, that's right. There's only a slight difference in input parameters (port number vs interface string). Do you want me to remove the test?

extra_data/export.py
tmichela: Right, thanks

pytest.ini
tmichela: fair enough

tmichela:
any more comments?

takluyver:
Thanks, I think this is looking good.

takluyver:
Actually, one more adjustment: this requires a new version of karabo-bridge, right? Let's specify that in `setup.py`.

tmichela:
Thanks!

PR #51, opened 2020-04-17
attempt to generalize multi module detector combined source

tmichela:
Generalize 'appending' modules sources in a single data source for the cli tool `karabo-bridge-serve-files`

@dallanto


tmichela:
any idea why the test is hanging on travis? It works on my machine..

takluyver:
Maybe it can't use port 3333 for some reason? Try setting timeout on the client, see if it turns the hang into a failure.

extra_data/tests/test_streamer.py
```diff
@@ -59,6 +61,18 @@ def server(protocol_version):
         yield s
 
 
+@pytest.fixture(scope='function')
+def file_server(mock_fxe_raw_run):
+    port = 3333
+    args = shlex.split(
+        f'karabo-bridge-serve-files {mock_fxe_raw_run} {port} '
+        f'--append-detector-modules'
+    )
```
takluyver, line 70:
```suggestion
    args = [
        'karabo-bridge-serve-files', str(mock_fxe_raw_run), str(port),
        '--append-detector-modules'
    ]
```

Grumble grumble, don't format things into a single string just to split them up again. E.g. if `mock_fxe_raw_run` contains a space, then formatting it into a command string like this turns it into two arguments.

takluyver:
This incorporates #50 now, right? Let's get that one in first, so it's easier to review the diff here.

tmichela:
> This incorporates #50 now, right? Let's get that one in first, so it's easier to review the diff here.

yep

takluyver:
Can you rebase this on master?

Now that the actual ZMQ streaming is not part of this repository, maybe some of the tests could mock out `streamer.feed()` and just check what gets fed to it. It would still be good to have one integration test that ensures it works together with karabo-bridge to actually stream data, though.

tmichela:
> Can you rebase this on master?

It's already on top of master

> Now that the actual ZMQ streaming is not part of this repository, maybe some of the tests could mock out streamer.feed() and just check what gets fed to it. It would still be good to have one integration test that ensures it works together with karabo-bridge to actually stream data, though.

I'm not sure to get what you mean by mocking out streamer.feed()?
The test with the cli tool should check that it plays nice with karabo-bridge, except they're hanging :'D (on travis)

takluyver:
> It's already on top of master

Right, but it has merge conflicts. I'm used to rebasing to fix that, but I'm not particular about this - you can also merge from master if you prefer (this is what GitHub does if you 'resolve conflicts' here).

> I'm not sure to get what you mean by mocking out streamer.feed()?

If we want to test the transformations we do to stack detector modules together, we could replace `ServerInThread` with a mock class which would capture the objects passed to `.feed()`. Then we could test that separately from the functionality to send data over ZMQ.

Although on reflection, maybe it's better to refactor the code a bit so the appending is a separate function/class, and we can test it without having to use mock objects, which are a bit of a workaround.

> The test with the cli tool should check that it plays nice with karabo-bridge, except they're hanging :'D 

Yeah, that's something of a pain. As a starting point, let's set a timeout in all tests which use a karabo bridge `Client()`, so that at least it fails explicitly rather than hanging until Travis gives up. Maybe there should even be a default timeout in karabo-bridge?

We'd never see errors in the subprocess we start, because we never get its exit status at all. Ideally, we'd wait in the fixture until the subprocess had successfully started, but that requires defining a means for it to notify that startup has finished successfully.

Is `localhost` the problem? Maybe we should try with `127.0.0.1` instead.

tmichela:
>  but it has merge conflicts

ah, sorry I did not see that, should be fixed not

> Although on reflection, maybe it's better to refactor the code a bit so the appending is a separate function/class, and we can test it without having to use mock objects, which are a bit of a workaround.

Yes, I'll try to do that tomorrow

takluyver:
Now it's carrying on some way past these tests and then hanging? That smells like something going wrong when garbage collection happens, but I'm not sure what. Is it possible that we need to close the ZMQ socket before destroying the context? It doesn't look like it from the pyzmq docs, but maybe we're hitting a bug there.

takluyver:
`Popen` also does some stuff in `__del__` to try to clean up its PID - maybe we should have a `.wait()` in the test to do that explicitly?

tmichela:
> Is it possible that we need to close the ZMQ socket before destroying the context?

Destroy will close all sockets (after # lingering messsages, but this is set to zero).
But that shouldn't even have any impact since we kill the process? :thinking:

takluyver:
I thought for a moment I'd spotted the problem. I haven't, but I'm pretty sure I've spotted *a* problem: once `serve_files` is finished putting files in the queue, it will stop the sender thread without waiting for them to be sent. So the last 9/10 trains will be lost.

Is it really necessary important to use two threads here? Could we just read a train, prepare it and send it? Performance with the REQ/REP socket wouldn't be great, but using PUSH/PULL and letting ZMQ do the buffering seems like a better answer than reimplementing what ZMQ does.

takluyver:
(In fact, we should also wait for ZMQ to send buffered messages, otherwise we just have the same problem there when using a PUB or PUSH socket.

tmichela:
> I thought for a moment I'd spotted the problem. I haven't, but I'm pretty sure I've spotted a problem: once serve_files is finished putting files in the queue, it will stop the sender thread without waiting for them to be sent. So the last 9/10 trains will be lost.

Yep, nice one! I've changed that.
I'm still puzzled about this sub-process issue on travis though...

takluyver:
Maybe temporarily set it to run with pytest's `-x` option, to stop on the first error. That might let us see the traceback from the failing test before it hangs.

tmichela:
Seems like stdout doesn't return anything:

```console
extra_data/tests/test_streamer.py::test_serve_files[0.5] Interface: 
FAILED
=================================== FAILURES ===================================
____________________________ test_serve_files[0.5] _____________________________
mock_fxe_raw_run = '/tmp/tmpr1kvdib8'
    def test_serve_files(mock_fxe_raw_run):
        args = ['karabo-bridge-serve-files', '-z', 'PUSH', str(mock_fxe_raw_run), str(44444)]
        interface = ''
    
        with Popen(args, stdin=PIPE, stdout=PIPE, stderr=PIPE,
                   env=dict(os.environ, PYTHONUNBUFFERED='1')) as p:
            for line in p.stdout:
                line = line.decode('utf-8')
                if line.startswith('Streamer started on:'):
                    print('line:', line)
                    interface = line.partition(':')[2].strip()
                    break
    
            print('Interface:', interface)
>           with Client(interface, sock='PULL', timeout=5) as c:
extra_data/tests/test_streamer.py:60: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../virtualenv/python3.8.0/lib/python3.8/site-packages/karabo_bridge/client.py:80: in __init__
    self._socket.connect(endpoint)
zmq/backend/cython/socket.pyx:583: in zmq.backend.cython.socket.Socket.connect
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
>   ???
E   zmq.error.ZMQError: Invalid argument
zmq/backend/cython/checkrc.pxd:26: ZMQError
```

takluyver:
I can't figure that out either. If you put the `print()` outside the `if`, does it show anything?

Or does it work if we hardcode 127.0.0.1?

tmichela:
> If you put the print() outside the if, does it show anything?

it does not show anything either

> Or does it work if we hardcode 127.0.0.1?

no, the client times out

takluyver:
Digging through some Jupyter code, it looks like it uses `socket.gethostbyname_ex('localhost')[2]` to get a list of IP addresses to use. But I believe that 127.0.0.1 as localhost is written into the IP standards, so I'd  be surprised if it wasn't valid as a local IP. :confused: 

What about if we print stderr from the subprocess? Redirecting it into stdout with `stder=subprocess.STDOUT` is probably the quickest way to do that. Maybe there's an error message we're not seeing.

takluyver:
Ah, it's setuptools getting involved :angry:

takluyver:
OK, it's working now, but I'd like to think a bit about how we can ensure we don't face similar difficulties debugging problems like this in the future. When starting the streamer subprocess, I think we need a way for the parent to wait until either it's successfully started and ready to stream, or it's crashed.

How we design this might depend on whether it's just for tests, or whether there are other applications that might launch the file streamer.

- If there are more use cases, we should make something neat without getting in the way of normal usage. E.g. the parent process might specify an extra ZMQ socket, which the streamer would connect to and send a message to say it was ready. This would also mean the streamer could pick a random port and tell the parent which it is.
- If it's just for the tests, we can do something rougher, like scanning stdout or waiting for a file to appear, so long as it doesn't get stuck if the streamer crashes.

tmichela:
Ah merci !

tmichela:
> How we design this might depend on whether it's just for tests, or whether there are other applications that might launch the file streamer.

I don't really know if an other application might launch the streamer... I've added a check in the tests for now that will raise if the interface string is empty.

takluyver:
From our discussion in chat: when it finishes normally (i.e. not cancelled with Ctrl-C or SIGTERM), we should set the ZMQ socket linger to either -1 (representing infinity), or a fairly large positive value (units are milliseconds), to ensure it waits for the last message to really be transferred before it exits.

takluyver:
@tmichela for now I've made it manually close the socket with `linger=-1`. I don't know if we want to add a higher level API for this in karabo-brige, e.g. a `sender.finished()` method or a context manager. But that would have to wait for a new release, in any case.

tmichela:
@takluyver thanks. Yes, I was thinking of something similar as well for karabo-bridge.

takluyver:
Great. I never know exactly how to manage review once I push to a PR, but this LGTM, so if you're happy with it, feel free to merge it.

tmichela:
Thanks. Not sure I'm happy with the process of this PR :'D
But the original feature is there. I'll probably update this when #61 is merged

PR #52, opened 2020-04-17
add missing dependency for test: mock

tmichela:


tmichela:
tests are breaking because of the new release of `karabo-bridge`, it is fixed in #50

takluyver:
The `mock` package actually got folded in to the standard library as [unittest.mock](https://docs.python.org/3/library/unittest.mock.html), so maybe we just need to change an import?

PR #53, opened 2020-04-29
starting commit: add Jungfrau class inheriting from MPxDetectorBase; …

dallanto:
@takluyver  
Hi Thomas, I had a start by trying to follow your recommendations in the e-mail. To pass n_modules, I use the ``modules=`` parameter as the base class constructor (``__init__``) seems to expect it this way.  
From the front-end POV, that's another optional CLI argument, passed to ``get_detector()``.  
I have a conceptual problem with the detector name as in the ``re()`` expressions, though:  
The data aggregator source name in ``/gpfs/exfel/exp/SPB/201922/p002566/proc/r0061`` HDF5 files are like: ``SPB_IRDA_JNGFR/DET/MODULE_1:daqOutput`` and I cannot easily relate this to the scheme used for AGIPD-1M.

### DIFF
--- a/extra_data/cli/make_virtual_cxi.py
+++ b/extra_data/cli/make_virtual_cxi.py
@@ -13,6 +13,11 @@ log = logging.getLogger(__name__)
 
 
 def _get_detector(data, min_modules, modules=None):
+    if modules != None and ',' in modules:
+        try:  
+            modules = [int(x) for x in modules.split(',')]
+        except ValueError:
+            modules=None
     for cls in (AGIPD1M, LPD1M, JUNGFRAU):
         try:
             return cls(data, min_modules=min_modules, modules=modules)
@@ -36,8 +41,8 @@ def main(argv=None):
         help="Include trains where at least N modules have data (default 9)"
     )
     ap.add_argument(
-        '--n_modules', type=int, default=8,
-        help="Expect this flexibly assembled detector to have N modules"
+        '--modules', type=str, default='',
+        help="String encoding a sequence of modules, e.g. 1,2,3,4"
     )
     args = ap.parse_args(argv)
     out_file = args.output
@@ -73,7 +78,7 @@ def main(argv=None):
 
     log.info("Reading run directory %s", run_dir)
     run = RunDirectory(run_dir)
-    det = _get_detector(run, args.min_modules, modules=args.n_modules)
+    det = _get_detector(run, args.min_modules, modules=args.modules)
     if det is None:
         sys.exit("No AGIPD or LPD sources found in {!r}".format(run_dir))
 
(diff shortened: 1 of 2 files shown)


dallanto:
O.k. from the CI error message I see that modules is not an integer but a sequence. I should have read the docstring more carefully: "set of ints"

dallanto:
Another thing: in the example run (as per path given), the modules are numbered 1,2,3,6,7,8 - not starting with 0 and not contiguous in this case. So maybe one should better have a kwarg for the sequence per se, not an int for the number ...

dallanto:
O.k. the current commit ist completely tentative. I know that instead of replacing source and key from AGIPD/LPD to JUNGFRAU, I should override the base class methods with distinct methods in the JUNGFRAU class, and I will do so.  
For the time being, I could quickly verify that I get further now, the sources are found and also the train check vs. min_modules returns something reasonable. Next, I am stuck with the ``get_index`` method, somehow somewhere the ``image.data`` key is set or passed, so group ``image`` is expected and not found, because the key is actually ``data.adc``. I pass ``data.adc`` at the one point I identified, but I am missing something else obviously.

takluyver:
Thanks Fabio. Yes, it looks like there are a couple of places where the data format from JUNGFRAU is a bit different from the other detectors we already support. I guess we need to move 1 or 2 hardcoded strings to class attributes.

> in the example run (as per path given), the modules are numbered 1,2,3,6,7,8 - not starting with 0 and not contiguous in this case. 

This shouldn't be a big issue for the `components` module itself, but may need some work in `stack_detector_data` and making virtual CXI files. I guess we'll need to treat those integers as a kind of name, and map them to module indices 0 to N-1.

dallanto:
Btw., because it's fitting the context somewhat, here is the hack that Henry Kirkwood applied yesterday in order to provide JUNGFRAU data to OnDA. He based it on the stand-alone "karabo-bridge-stack-files.py" (the prototype script of mine before adding the stacking option to EXtra-data) ... his adaptation was to extract the ``stack_detector_data`` method from extra_data and provide an explicit list of numbers (w/o mapping to regular indices 0 to N-1):

```
MODULES = [1,2,3,6,7,8]
def stack_detector_data(_data, key):
    """ this is a temporary replacement for th stack_detector_data function in extra-data as it currently doesn't handle jungfrau"""
    _stack = []
    for m in MODULES:
        data_path = 'SPB_IRDA_JNGFR/DET/MODULE_{}:daqOutput'.format(m)
        train_imgs = _data[data_path][key]
        _stack.append(train_imgs)
    train_stack = np.stack(_stack, axis=1)
    return train_stack
```
(code snippet)

dallanto:
Closed in order to delete branch and start all over again.

PR #56, opened 2020-05-11
Partial update of mockdata and test to support file version 1.0

tmichela:
- `mockdata` generates files closer to the real files (for version 1.0)
- run test on both format version (0.5, 1.0)

(spinoff from #39)

### DIFF
--- a/extra_data/tests/conftest.py
+++ b/extra_data/tests/conftest.py
@@ -8,82 +8,83 @@ from tempfile import TemporaryDirectory
 from . import make_examples
 
 
+@pytest.fixture(scope='session', params=['0.5', '1.0'])
+def format_version(request):
+    return request.param
+
+
 @pytest.fixture(scope='module')
-def mock_agipd_data():
+def mock_agipd_data(format_version):
     # This one uses the older index format
     # (first/last/status instead of first/count)
     with TemporaryDirectory() as td:
         path = osp.join(td, 'CORR-R9999-AGIPD07-S00000.h5')
-        make_examples.make_agipd_example_file(path)
+        make_examples.make_agipd_example_file(path, format_version=format_version)
         yield path
 
 
 @pytest.fixture(scope='module')
-def mock_lpd_data():
+def mock_lpd_data(format_version):
     with TemporaryDirectory() as td:
         path = osp.join(td, 'RAW-R9999-LPD00-S00000.h5')
-        make_examples.make_lpd_file(path)
+        make_examples.make_lpd_file(path, format_version=format_version)
         yield path
 
 
 @pytest.fixture(scope='module')
-def mock_fxe_control_data():
+def mock_fxe_control_data(format_version):
     with TemporaryDirectory() as td:
         path = osp.join(td, 'RAW-R0450-DA01-S00001.h5')
-        make_examples.make_fxe_da_file(path)
+        make_examples.make_fxe_da_file(path, format_version=format_version)
         yield path
 
 
 @pytest.fixture(scope='module')
-def mock_sa3_control_data():
+def mock_sa3_control_data(format_version):
     with TemporaryDirectory() as td:
         path = osp.join(td, 'RAW-R0450-DA01-S00001.h5')
-        make_examples.make_sa3_da_file(path)
+        make_examples.make_sa3_da_file(path, format_version=format_version)
         yield path
 
 
 @pytest.fixture(scope='module')
-def mock_sa3_control_data_fmt_1_0():
-    with TemporaryDirectory() as td:
-        path = osp.join(td, 'RAW-R0450-DA01-S00001.h5')
-        make_examples.make_sa3_da_file(path, format_version='1.0')
-        yield path
-
-
-@pytest.fixture(scope='module')
-def mock_spb_control_data_badname():
+def mock_spb_control_data_badname(format_version):
     with TemporaryDirectory() as td:
         path = osp.join(td, 'RAW-R0309-DA01-S00000.h5')
-        make_examples.make_data_file_bad_device_name(path)
+        make_examples.make_data_file_bad_device_name(path, format_version=format_version)
         yield path
 
 
 @pytest.fixture(scope='session')
-def mock_fxe_raw_run():
+def mock_fxe_raw_run(format_version):
     with TemporaryDirectory() as td:
-        make_examples.make_fxe_run(td)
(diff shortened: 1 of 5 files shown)


extra_data/tests/mockdata/base.py
```diff
@@ -142,9 +147,27 @@ def write_metadata(h5file, data_sources, chunksize=16, format_version='0.5'):
...
+        h5file['METADATA/karaboFramework'] = [b'2.7.0']
+        h5file.create_dataset('METADATA/proposalNumber', dtype=np.uint32, data=[700000])
+        h5file.create_dataset('METADATA/runNumber', dtype=np.uint32, data=[1])
+        h5file['METADATA/runType'] = [b'Test DAQ']
+        h5file['METADATA/sample'] = [b'No Sample']
+        print(osp.basename(h5file.filename))
+        # get sequence number
+        fname_pattern = r'^(RAW|CORR)\-R\d+\-.*\-S(\d+).h5$'
+        match = re.match(fname_pattern, osp.basename(h5file.filename))
+        sequence = int(match.groups()[1]) if match is not None else 0
```
takluyver, line 162: We can use `match.group(1)`, or even `match[1]` (just discovered the second form, which is apparently new in 3.6)

extra_data/tests/mockdata/base.py
```diff
@@ -142,9 +147,27 @@ def write_metadata(h5file, data_sources, chunksize=16, format_version='0.5'):
...
-    """Make a dataset of fake train IDs at the given path
+        now = datetime.utcnow().replace(microsecond=0).strftime('%Y%m%dT%H%M%SZ')
+        h5file['METADATA/creationDate'] = [now.encode('ascii')]
+        h5file['METADATA/daqLibrary'] = [b'1.9.0']
+        h5file['METADATA/karaboFramework'] = [b'2.7.0']
+        h5file.create_dataset('METADATA/proposalNumber', dtype=np.uint32, data=[700000])
+        h5file.create_dataset('METADATA/runNumber', dtype=np.uint32, data=[1])
+        h5file['METADATA/runType'] = [b'Test DAQ']
+        h5file['METADATA/sample'] = [b'No Sample']
+        print(osp.basename(h5file.filename))
```
takluyver, line 158:
```suggestion
```

extra_data/tests/mockdata/base.py
```diff
@@ -142,9 +147,27 @@ def write_metadata(h5file, data_sources, chunksize=16, format_version='0.5'):
...
+        # get sequence number
+        fname_pattern = r'^(RAW|CORR)\-R\d+\-.*\-S(\d+).h5$'
+        match = re.match(fname_pattern, osp.basename(h5file.filename))
+        sequence = int(match.groups()[1]) if match is not None else 0
+        h5file.create_dataset('METADATA/sequenceNumber', dtype=np.uint32, data=[sequence])
+        h5file['METADATA/updateDate'] = [now.encode('ascii')]
```
takluyver, line 164: Interestingly, in some files I've just looked at, `updateDate` is about 5 minutes after `creationDate`. I guess it's the start and end of writing that file. I don't think we need to simulate that, but it might be good to know.

These fields also seem to relate to the file, not the run as a whole; at least, they differ for different files.

extra_data/tests/test_streamer.py
```diff
@@ -1,147 +0,0 @@
-"""Test streaming data with ZMQ interface."""
```
takluyver, line 1: Can you rebase on master so we can check that this deletion doesn't make it into the merge? There's still a much reduced `test_streamer.py` after  #50.

extra_data/tests/conftest.py
```diff
@@ -8,83 +8,84 @@
 from . import make_examples
 
 
+@pytest.fixture(scope='session', params=['0.5', '1.0'])
+def format_version(request):
+    return request.param
```
takluyver, line 13: I'm not totally keen on parameterising every mock data test like this. pytest makes it really easy to do this, and it's good if you do need to test how different options interact, but for most of what we're testing the changes in format 1.0 don't really matter, so running the test twice is a bit of a waste. This currently adds about 30 seconds to running the tests, and that will increase as we keep adding tests.

We don't necessarily need to address this now - 30 seconds extra isn't *that* bad - I just want to register that at some point we'll want to be more selective.

extra_data/tests/mockdata/base.py
```diff
@@ -153,5 +176,18 @@ def write_train_ids(f, path, N, first=10000, chunksize=16):
         Npad = N + chunksize - (N % chunksize)
     else:
         Npad = N
-    ds = f.create_dataset(path, (Npad,), 'u8', maxshape=(None,))
+
+    if format_version != '0.5':
+        # flag
+        ds = f.create_dataset('INDEX/flag', (Npad,), 'u4', maxshape=(None,))
```
takluyver, line 182: In the file I'm looking at, flag is a signed `int32` (`'i4'` in this syntax).

extra_data/tests/mockdata/base.py
tmichela: neat, thanks

extra_data/tests/mockdata/base.py
tmichela: Interesting, so it's probably when the file is closed? I updated the value to be a few minutes after creationDate

extra_data/tests/mockdata/base.py
tmichela: oh, I thought that I checked that. updated.

extra_data/tests/conftest.py
tmichela: Yep, I thought about this as well. Probably only a subset of the tests would require testing on different version but as you said, this has not a big impact at the moment.
We could also think different tests strategies, simple/fast on branches and extended tests when merging to master, or in a nightly build, ... (we could then also run much longer tests  download real data, processing, ...)

takluyver: We could, but different test strategies and ways of switching between them means more complexity, so I'd rather put it off if possible.

Maybe the biggest win in that direction would be if we had a quick automated way of finding a newly written data file on Maxwell, we could try to have a nightly test using a new file, to alert us if something unexpected happened to the format. Though we're getting something a bit like that by running validation in the calibration pipeline.

takluyver:
LGTM

tmichela:
Thanks!

PR #59, opened 2020-05-18
write_virtual_cxi: configurable fill values

### LINKED ISSUE #58: Make virtual file: configurable fill value for missing data
Currently when making virtual hdf5 files, the fill value for missing data is `np.nan`. Some software do not play nicely with NANs (Crystfel? where peaks are found with nan intensities), we should add the option to chose the fill value.

tmichela:
usage:
```console
extra-data-make-virtual-cxi -0 ./outpu.h5 -v data 5 -v mask 99 -v gain 2 /path/to/run
```

resolves #58

### DIFF
--- a/extra_data/cli/make_virtual_cxi.py
+++ b/extra_data/cli/make_virtual_cxi.py
@@ -35,8 +35,14 @@ def main(argv=None):
         '--min-modules', type=int, default=9, metavar='N',
         help="Include trains where at least N modules have data (default 9)"
     )
+    ap.add_argument(
+        '-v', '--fill-value', action='append', nargs=2,
+        help='define fill value for individual dataset (data, gain or mask)'
+             '(default "nan")'
+    )
     args = ap.parse_args(argv)
     out_file = args.output
+    fill_values = dict(args.default_values)
 
     logging.basicConfig(level=logging.INFO)
 
@@ -73,7 +79,7 @@ def main(argv=None):
     if det is None:
         sys.exit("No AGIPD or LPD sources found in {!r}".format(run_dir))
 
-    det.write_virtual_cxi(out_file)
+    det.write_virtual_cxi(out_file, fill_values)
 
 if __name__ == '__main__':
     main()
--- a/extra_data/cli/make_virtual_cxi.py
+++ b/extra_data/cli/make_virtual_cxi.py
@@ -27,7 +27,7 @@ def _detectors():
 
 
 def main(argv=None):
-    ap = argparse.ArgumentParser('karabo-data-make-virtual-cxi')
+    ap = argparse.ArgumentParser('extra-data-make-virtual-cxi')
     ap.add_argument('run_dir', help="Path to an EuXFEL run directory")
     # Specifying a proposal directory & a run number is the older interface.
     # If the run_number argument is passed, run_dir is used as proposal.
--- a/extra_data/cli/make_virtual_cxi.py
+++ b/extra_data/cli/make_virtual_cxi.py
@@ -55,8 +55,8 @@ def main(argv=None):
     ap.add_argument(
         '--fill-value', action='append', nargs=2, metavar=('DS', 'V'),
         help='define fill value (V) for individual dataset (DS). Datasets are'
-             ' "data", "gain" and "mask". (defaults: data: nan (float32) or 0'
-             ' (uint16); gain: 0; mask: 0xffffffff)'
+             ' "data", "gain" and "mask". (defaults: data: nan (proc, float32)'
+             ' or 0 (raw, uint16); gain: 0; mask: 0xffffffff)'
     )
     args = ap.parse_args(argv)
     out_file = args.output
--- a/extra_data/cli/make_virtual_cxi.py
+++ b/extra_data/cli/make_virtual_cxi.py
@@ -55,8 +55,8 @@ def main(argv=None):
     ap.add_argument(
         '--fill-value', action='append', nargs=2, metavar=('DS', 'V'),
         help='define fill value (V) for individual dataset (DS). Datasets are'
-             ' "data", "gain" and "mask". (defaults: data: nan; gain: 0; '
-             ' mask: 0xffffffff)'
+             ' "data", "gain" and "mask". (defaults: data: nan (float32) or 0'
+             ' (uint16); gain: 0; mask: 0xffffffff)'
     )
     args = ap.parse_args(argv)
     out_file = args.output
--- a/extra_data/cli/make_virtual_cxi.py
+++ b/extra_data/cli/make_virtual_cxi.py
@@ -6,18 +6,24 @@ import re
 import sys
 
 from extra_data import RunDirectory
-from extra_data.components import AGIPD1M, LPD1M
+from extra_data.components import MPxDetectorBase
 from extra_data.exceptions import SourceNameError
 
 log = logging.getLogger(__name__)
 
 
 def _get_detector(data, min_modules):
-    for cls in (AGIPD1M, LPD1M):
+    for cls in MPxDetectorBase.__subclasses__():
         try:
             return cls(data, min_modules=min_modules)
         except SourceNameError:
             continue
+
+
+def _detectors():
+    """returns a list of names for all detector components available
+    """
+    return [d.__name__ for d in MPxDetectorBase.__subclasses__()]
 
 
 def main(argv=None):
@@ -77,7 +83,7 @@ def main(argv=None):
     run = RunDirectory(run_dir)
     det = _get_detector(run, args.min_modules)
     if det is None:
-        sys.exit("No AGIPD or LPD sources found in {!r}".format(run_dir))
+        sys.exit(f"No {_detectors()} sources found in {run_dir}")
 
     det.write_virtual_cxi(out_file, fill_values)
 
--- a/extra_data/cli/make_virtual_cxi.py
+++ b/extra_data/cli/make_virtual_cxi.py
@@ -31,12 +31,13 @@ def main(argv=None):
     example = dedent("""
         Example:
 
-          extra-data-make-virtual-cxi -o ./out_file.h5 --min-modules 15
-          --fill-value data 0 --fill-value gain 1 /path/to/source/run
+          extra-data-make-virtual-cxi -o ./out_file.h5 --min-modules 15 \\
+            --fill-value data 0 --fill-value gain 1 /path/to/source/run
     """)
     ap = argparse.ArgumentParser(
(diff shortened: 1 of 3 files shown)


extra_data/write_cxi.py
```diff
@@ -163,7 +163,26 @@ def collect_data(self):
 
         return layouts
 
-    def write(self, filename):
+    @staticmethod
+    def _fill_value(path, layout, fillvalues):
+        array_type = layout[path].dtype.type(0).item()
+        fallback = np.nan if isinstance(array_type, float) else 0
+        value = fillvalues.get(path, fallback)
+        return layout[path].dtype.type(value)
```
tmichela, line 171: this is a bit convoluted as `create_virtual_dataset` seems to silently convert `nan` to `0` for integer arrays, but this will raise an error if someone tries to fill an int array with nan instead of silently converting it to zeros.

dallanto: I see your attempt is more professional 

> > How would this dictionary argument look like from the user perspective, at the CLI:
> > `extra-data-make-virtual cxi <RUN-PATH> -o <OUT-NAME> --fill-value=data:0` ?
> 
> There is an example in the PR description #59



> > How would this dictionary argument look like from the user perspective, at the CLI:
> > `extra-data-make-virtual cxi <RUN-PATH> -o <OUT-NAME> --fill-value=data:0` ?
> 
> There is an example in the PR description #59

Very well, thanks, right on top and I missed it ...

extra_data/cli/make_virtual_cxi.py
```diff
@@ -4,24 +4,40 @@
...
-    ap = argparse.ArgumentParser('karabo-data-make-virtual-cxi')
+    example = dedent("""
+        Example:
+
+          extra-data-make-virtual-cxi -o ./out_file.h5 --min-modules 15
+          --fill-value data 0 --fill-value gain 1 /path/to/source/run
+    """)
+    ap = argparse.ArgumentParser(
+        'extra-data-make-virtual-cxi', epilog=example,
+        formatter_class=argparse.RawDescriptionHelpFormatter
```
takluyver, line 39: Fancy argparse options! :-)

extra_data/cli/make_virtual_cxi.py
```diff
@@ -4,24 +4,40 @@
...
+    return [d.__name__ for d in MPxDetectorBase.__subclasses__()]
+
+
 def main(argv=None):
-    ap = argparse.ArgumentParser('karabo-data-make-virtual-cxi')
+    example = dedent("""
+        Example:
+
+          extra-data-make-virtual-cxi -o ./out_file.h5 --min-modules 15
+          --fill-value data 0 --fill-value gain 1 /path/to/source/run
```
takluyver, line 35: I might end the first line with a `\` (escaped as `\\` in a Python string) and indent the second line a couple of spaces to make it obvious that it's one command split over two lines.

extra_data/cli/make_virtual_cxi.py
```diff
@@ -35,8 +51,15 @@ def main(argv=None):
         '--min-modules', type=int, default=9, metavar='N',
         help="Include trains where at least N modules have data (default 9)"
     )
+    ap.add_argument(
+        '-v', '--fill-value', action='append', nargs=2, metavar=('DS', 'V'),
```
takluyver, line 55: I'd prefer to skip the `-v` short form. An option taking two arguments is a bit fiddly for a short option anyway, and `-v` commonly means 'verbose mode', which it's conceivable we'd want some day.

dallanto:
> I'm actually ambivalent between this and #60. This is certainly more flexible, but that also means more complexity, and I wonder how much flexibility we need - maybe toggling between zero and nan is enough? I know I was the one talking about making it more flexible last week, but I wasn't sure then and I'm not sure now. ;-)
> 
> I'll have a think over lunch, and maybe we can have a chat about it this afternoon.

Sure we can chat.  
My current thought: If it was just about the ``image.data`` I'd say 0 or nan is enough. But as we can address all three items data, mask, gain it gets a bit more complicated anyway. I am not sure that then all-nan vs. all-zero (or 0, 1, 0) is the best choice. By making it flexible we'd delegate that choice to the user ...

takluyver:
From a user perspective: command line options which take two values are pretty rare, so they can be surprising when they occur.

As there are only 3 datasets to consider, maybe it's simplest to do a separate option for each, e.g. `--fill`, `--fill-gain`, `--fill-mask`.

dallanto:
> From a user perspective: command line options which take two values are pretty rare, so they can be surprising when they occur.
> 
> As there are only 3 datasets to consider, maybe it's simplest to do a separate option for each, e.g. `--fill`, `--fill-gain`, `--fill-mask`.

And here ``--fill`` for image data plus  ``--fill-gain``  could be boolean (we need only two values, default and alternative) while ``--fill-mask`` would allow for multiple positive integers.

tmichela:
It's probably more confusing if similar arguments have different behavior?

tmichela:
Is there a possibility to have more arrays added in addition to data, gain and mask in the future?

tmichela:
As mentioned on slack, it still probably useful to have more possibilities than nan and 0, at least for data (and mask). The 3 different options is a good idea idea

takluyver:
Yeah, I think if you can set an arbitrary value for one, we'd probably allow the same for any of them, so it's consistent.

I don't know, maybe we could get extra datasets. The AGIPD calibration prototype I worked on saved a `blShift` (baseline shift, I guess) field in the proc folder. And the pulseId and cellId should always be the same, I think, but if that's not true it might make sense to put them in the virtual file.

extra_data/write_cxi.py
```diff
@@ -163,7 +163,26 @@ def collect_data(self):
 
         return layouts
 
-    def write(self, filename):
+    @staticmethod
+    def _fill_value(path, layout, fillvalues):
+        array_type = layout[path].dtype.type(0).item()
```
takluyver, line 168: I think the type check can be a bit simpler:

```python
np.issubdtype(dt, np.floating)
# Or
dt.kind == 'f'
````

File: extra_data/cli/make_virtual_cxi.py (hunk 2 above)
Comment (tmichela) [reply extra_data/cli/make_virtual_cxi.py]: everytime I use argparse, I find new arguments :monocle_face:

File: extra_data/write_cxi.py (hunk 5 above)
Comment (tmichela) [reply extra_data/write_cxi.py]: Thanks, I removed it completely since now all 3 arrays have different defaults

Comment (tmichela) [reply extra_data/write_cxi.py]: oh, I forgot that raw `data` is integer :'D

File: extra_data/cli/make_virtual_cxi.py [hunk 6]
```diff
@@ -55,8 +55,8 @@ def main(argv=None):
     ap.add_argument(
         '--fill-value', action='append', nargs=2, metavar=('DS', 'V'),
         help='define fill value (V) for individual dataset (DS). Datasets are'
-             ' "data", "gain" and "mask". (defaults: data: nan; gain: 0; '
-             ' mask: 0xffffffff)'
+             ' "data", "gain" and "mask". (defaults: data: nan (float32) or 0'
+             ' (uint16); gain: 0; mask: 0xffffffff)'
```
dallanto, line 59: This means: ``nan`` is the default for proc data (float32), while 0 is the default for raw data (uint16), correct? There is no nan-equivalent for integers, I guess, but still 0 has quite a different meaning as Thomas pointed out. I know that 0 is exactly what we want as work-around to make CrystFEL happy, but still ...

tmichela: yep, there is no `nan` for integers. I figured `0` would be the most appropriate equivalent there?

dallanto: Practically, yes, for many scenarios. For crystallography (CrystFEL in particular) I'd tend to prefer a (absolutely spoken: large) negative value - but a reasonable default would be very situation-dependent and difficult to set - hence, the specific user value becomes important. And zero is still the best value for a default.

takluyver: There are no negative values for `uint16` (the `u` means unsigned, i.e. no way to represent negative numbers). One convention is to use the largest possible number, 65535. But I think 0 is the more obvious default.

I think the help string should say that the distinction is between corrected & raw data, though; it's not necessarily obvious why there are two defaults for data or which way round they go.

extra_data/write_cxi.py
```diff
@@ -163,13 +163,33 @@ def collect_data(self):
...
         pulse_ids = self.collect_pulse_ids()
         experiment_ids = np.core.defchararray.add(np.core.defchararray.add(
             self.train_ids_perframe.astype(str), ':'), pulse_ids.astype(str))
 
         layouts = self.collect_data()
 
+        _fillvalues = {
+            # data can be uint16 (raw) or float32 (proc)
+            'data': np.nan if layouts['data'].dtype.kind == 'f' else 0,
+            'gain': 0, 'mask': 0xffffffff}
```
takluyver, line 189:
```suggestion
            'gain': 0,
            'mask': 0xffffffff,
        }
```

extra_data/write_cxi.py
```diff
@@ -200,19 +220,22 @@ def write(self, filename):
                 dgrp['data_gain'] = h5py.SoftLink('/entry_1/data_gain')
 
             data = dgrp.create_virtual_dataset(
-                'data', layouts['data'], fillvalue=np.nan
+                'data', layouts['data'],
+                fillvalue=layouts['data'].dtype.type(_fillvalues['data'])
```
takluyver, line 224: Add a comment somewhere to say why we're doing this rather than just passing the number in? I think I recall that it was so that illegal values would error rather than being quietly converted to something else, but it would be good to record that in the code.

Also, if that is the rationale, does it make sense to do the check before we open the output file? Then if it fails, there's no incomplete file left over.

extra_data/cli/make_virtual_cxi.py
dallanto: Yes, you are correct with uint16 - actually I was referring to proc treatment with SFX application in mind (because that's what I am testing now: CrystFEL with VDS where -10000 is used as fillvalue). So for proc with float32 it's possible, but I'd leave the default at nan. And let the default for raw be 0, as I wrote.

tmichela: Alright. I've updated the docstring with proc/raw for data.
We seem to all agree on the defaults \o/

extra_data/write_cxi.py
tmichela: Good point! I've moved it before file is created

takluyver:
:+1:

PR #60, opened 2020-05-18
introduce optional cli argument 2 alternative values to fill with

dallanto:
@takluyver  @tmichela 
A very crude and simple way to implement the requested feature of having missing modules filled with 0 instead of ``np.nan`` in the virtual CXI.

### DIFF
--- a/extra_data/cli/make_virtual_cxi.py
+++ b/extra_data/cli/make_virtual_cxi.py
@@ -35,6 +35,12 @@ def main(argv=None):
         '--min-modules', type=int, default=9, metavar='N',
         help="Include trains where at least N modules have data (default 9)"
     )
+    ap.add_argument(
+        '--fill-zero', type=bool, default=False,
+        help="for missing modules, fill arrays with zeroes (default: do not"
+             "do this; fill with NaNs)",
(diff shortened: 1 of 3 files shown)


tmichela:
@dallanto I outran you :D https://github.com/European-XFEL/EXtra-data/pull/59

We should have better coordinated that task

dallanto:
It's probably wiser to fill the ``image.mask`` dataset with ones instead of zeroes. Thus we indicate there is something unusual (0 would indicate normal pixels, if I am not mistaken). If e.g. CrystFEL would thus ignore the missing, formally filled, module that's fine. If not - then it reads 0 for the ``image.data`` value, which is still better than NaN. What do you think?

dallanto:
> @dallanto I outran you :D #59
> 
> We should have better coordinated that task

Yes, my bad. Congrats for outrunning me  : )

dallanto:
Ok both, @takluyver, @tmichela  - I formally withdraw my pull request.  
Note that I commented on the issue and asked a question on you pull request, Thomas M.

takluyver:
We went for #59 instead of this one.

PR #62, opened 2020-05-28
Jungfrau make vds

dallanto:
@takluyver  
@tmichela  
This is the new attempt of extending the virtual cxi writing for the JUNGFRAU detector.  
I got somewhat further this time, also in the process at runtime, but still end-up with an error.
Clearly, the integration to a new class (e. g. outsourcing class variables) is not yet complete.


dallanto:
More concretely this is the current run-time status with error:  
```
INFO:extra_data.cli.make_virtual_cxi:Reading run directory /gpfs/exfel/exp/SPB/201922/p002566/proc/r0061
 ### DET  SPB_IRDA_JNGFR
 ### SRC {'SPB_IRDA_JNGFR/DET/MODULE_6:daqOutput': 6, 'SPB_IRDA_JNGFR/DET/MODULE_8:daqOutput': 8, 'SPB_IRDA_JNGFR/DET/MODULE_7:daqOutput': 7, 'SPB_IRDA_JNGFR/DET/MODULE_3:daqOutput': 3, 'SPB_IRDA_JNGFR/DET/MODULE_1:daqOutput': 1, 'SPB_IRDA_JNGFR/DET/MODULE_2:daqOutput': 2}
INFO:extra_data.components:Found 472 trains with data for at least 6 modules
INFO:extra_data.write_cxi:Up to 1 frames per train, 472 frames in total
INFO:extra_data.write_cxi:Identified calibrated data
INFO:extra_data.write_cxi:Virtual data shape: (472, 16, 1024, 512)
 ### Source: SPB_IRDA_JNGFR/DET/MODULE_6:daqOutput, ModNo: 6, Key: data.adc
 ### Source: SPB_IRDA_JNGFR/DET/MODULE_8:daqOutput, ModNo: 8, Key: data.adc
 ### Source: SPB_IRDA_JNGFR/DET/MODULE_7:daqOutput, ModNo: 7, Key: data.adc
 ### Source: SPB_IRDA_JNGFR/DET/MODULE_3:daqOutput, ModNo: 3, Key: data.adc
 ### Source: SPB_IRDA_JNGFR/DET/MODULE_1:daqOutput, ModNo: 1, Key: data.adc
 ### Source: SPB_IRDA_JNGFR/DET/MODULE_2:daqOutput, ModNo: 2, Key: data.adc
INFO:extra_data.write_cxi:Assembled 30 chunks for data.adc, filling 37.50% of the hyperslab
...
INFO:extra_data.write_cxi:Writing to test.cxi
  File "Software/python_frameworks/extra-data-jf/bin/extra-data-make-virtual-cxi", line 11, in <module>
    load_entry_point('EXtra-data', 'console_scripts', 'extra-data-make-virtual-cxi')()
  File "/home/dallanto/Devel/DA-Tools/EXtra-data/extra_data/cli/make_virtual_cxi.py", line 100, in main
    det.write_virtual_cxi(out_file, fill_values)
  File "/home/dallanto/Devel/DA-Tools/EXtra-data/extra_data/components.py", line 432, in write_virtual_cxi
    VirtualCXIWriter(self).write(filename, fillvalues=fillvalues)
  File "/home/dallanto/Devel/DA-Tools/EXtra-data/extra_data/write_cxi.py", line 221, in write
    layouts[self.cell_id_label])
  File "/home/dallanto/Software/python_frameworks/extra-data-jf/lib/python3.7/site-packages/h5py/_hl/group.py", line 168, in create_virtual_dataset
    maxshape=layout.maxshape, fillvalue=fillvalue)
  File "/home/dallanto/Software/python_frameworks/extra-data-jf/lib/python3.7/site-packages/h5py/_hl/dataset.py", line 193, in make_new_virtual_dset
    dcpl.set_virtual(vspace, fpath, dset, src_dspace)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5p.pyx", line 827, in h5py.h5p.PropDCID.set_virtual
ValueError: Invalid mapping selections (virtual and source space selections have different numbers of elements)
```

dallanto:
If you follow the code in ``write_cxi.py``, you see that I tried to replace the typical group and data names in the key, e.g. ``image.data`` --> ``data.adc`` by using class-member (object) variables like ``self.group_label`` or ``self.image_label``. There is likely a more elegant or more appropriate way.

dallanto:
At least the changes of the last commit do not break the existing functionality: the AGIPD VDS is still written correctly.

tmichela:
I think the problem is that the data is stored differently on AGIPD/LPD and on Jungfrau dataset. AGIPD/LPD have one row per image and JF has one row per trainId, with a second axis for pulse dimension. I guess that would require some work to make it fit...

I think there might be an easier solution for the difference in keys for datasets (for example using class attributes)

I see also at least 2 other problems:
1. naming of JF devices differ between installation, I've see at least:

  - FXE_XAD_JF1M/DET/RECEIVER-1
  - HED_IA1_JNGFR-1/DET/MODULE
  - SPB_IRDA_JNGFR/DET/MODULE_1

  We should match for all variations

2. There are different variation of detector layout (physical) between installation. We should make sure to handle them all. Maybe with a single interface (JUNGFRAU) but in the background returning a different class according on how many sources are found in the run?

takluyver:
> I think there might be an easier solution for the difference in keys for datasets (for example using class attributes)

I agree. The generic code shouldn't be checking 'is this JUNGFRAU?' - they should get the names they need from the specific detector subclasses. This will be a bit more work now just because we've written a lot of it around the data layout shared by AGIPD, LPD & DSSC.

> Maybe with a single interface (JUNGFRAU) but in the background returning a different class according on how many sources are found in the run?

I would suggest that there's one class, which sets `self.n_modules` when you instantiate it, either from the maximum module number it finds or an explicitly specified parameter. This parallels what we've already got in EXtra-geom.

> the VDS should still present the data as if it was one row per image ... that is, fusing the trains x pulses (cells) while stacking the modules

I don't actually know whether virtual datasets can fuse two axes, without mapping each train individually, which I'd rather avoid. [The design docs](https://support.hdfgroup.org/HDF5/docNewFeatures/VDS/HDF5-VDS-requirements-use-cases-2014-12-10.pdf) only say that the *number* of selected elements must match, which would appear to mean it would work. But they don't explicitly talk about reshaping, as far as I can see. And the design docs are not exactly what was implemented, in any case.

takluyver:
BTW, in terms of scope: I think we could quite easily make the data access class work - so you could load the JUNGFRAU data together in memory. Making a virtual CXI file might be a bit more complicated. If the former is any use without the latter, we could split them up.

But I'm also fine with including both parts if you want. I don't think it should be a massive task to get virtual datasets working.

tmichela:
> I would suggest that there's one class, which sets self.n_modules when you instantiate it, either from the maximum module number it finds or an explicitly specified parameter. This parallels what we've already got in EXtra-geom.

OK, I wasn't aware of the geom implementation

> ... And the design docs are not exactly what was implemented, in any case.

:sweat_smile:

takluyver:
To summarise our discussion today, I think it should be possible to do what we want without big changes to `components.py`, and without moving more complexity into the MultimodDetectorBase subclass. The key is that a lot of things that are complicated for xtdf (AGIPD-like) detector data are much simpler for JUNGFRAU data where there's a separate axis for pulses.

I think we should start by making a list of what extra features we want to add for JUNGFRAU, and then working out the simplest way to do these. E.g. for `get_array`, there are currently 3 extra parameters for xtdf detectors which aren't there for JUNGFRAU:

- `pulses=` selection. Simple slicing of pulses can already be done with the `roi=` parameter. Adding a pulses parameter to do the same thing should be fairly simple, just some extra checks and combining it into `roi` for the base class. Selecting pulse IDs might be a bit more complex, but I'm not sure JUNGFRAU stores these anyway?
- `unstack_pulses=`. The way we get the data from the files gives us the `True` option. I'm not convinced we need this, but if we want it, the `False` option should be as simple as calling [DataArray.stack](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.stack.html).
- `subtrain_index=` doesn't seem to make sense for JUNGFRAU, because there's only one option you could pass it. So I think it's better to leave it out.

We should also think about the bigger picture, though. The aim of this PR, and the bit I know people are requesting, is support for making virtual CXI files from JUNGFRAU data. Are the API changes in `components.py` necessary for that?

If needs be, we could start with some largely separate machinery for making virtual CXI files from JUNGFRAU data, even if that means some duplication. Then when it's working and useful, we can look at how much code can be shared with the xtdf-detectors-to-cxi mechanism. I think we're often too keen to avoid code duplication - it can be easier to understand two similar pieces of code doing concrete things than one mechanism with a lot of abstraction to handle different cases.

extra_data/cli/make_virtual_cxi.py
```diff
@@ -7,26 +7,20 @@
...
-from extra_data.components import XtdfDetectorBase
+from extra_data.components import detectors
 from extra_data.exceptions import SourceNameError
 
 log = logging.getLogger(__name__)
 
 
 def _get_detector(data, min_modules):
-    for cls in XtdfDetectorBase.__subclasses__():
+    for cls in detectors.list:
```
takluyver, line 17: I think we could use the `identify_multimod_detectors` function here and get rid of this little function.

extra_data/components.py
```diff
@@ -304,6 +329,25 @@ def _select_pulse_indices(pulses, firsts, counts):
         return np.concatenate(positions)
 
     def _make_image_index(self, tids, inner_ids, inner_name='pulse'):
+        """
+        Prepare indices for data per inner coordinate.
```
takluyver, line 333: Thanks for adding these! We don't normally bother writing the full lists of parameters etc. for private methods (names beginning with `_`), but there's also no need to remove them now you've added them.

extra_data/components.py
```diff
@@ -231,12 +253,15 @@ def get_dask_array(self, key, fill_value=None, astype=None):
 
         return self._concat(arrays, modnos, fill_value, astype)
 
-    def trains(self, require_all=True):
+    def trains(self, pulses=np.s_[:], require_all=True):
```
takluyver, line 256: This is in `MultimodDetectorBase` - you've added the pulses parameter here, making it equivalent to the method in `XtdfDetectorBase`, which you've then removed. However, if I'm reading the code in `MPxDetectorTrainIterator` correctly, passing `pulses=` will have no effect for JUNGFRAU - you'll get all pulses anyway.

I think we should put this and `XtdfDetectorBase.trains()` back as they are in master, at least for now. It's confusing to allow passing this parameter in for JUNGFRAU if it has no effect. If there are cases where it would be useful, we can come back and make it work at a later point.

extra_data/components.py
```diff
@@ -983,29 +1105,36 @@ def _make_image_index(self, tids, inner_ids, inner_name='pulse'):
...
+      Detector module numbers to use. By default, all available modules
+      are used.
     min_modules: int
       Include trains where at least n modules have data. Default is 1.
     """
     # We appear to have a few different formats for source names:
     # SPB_IRDA_JNGFR/DET/MODULE_1:daqOutput  (e.g. in p 2566, r 61)
     # SPB_IRDA_JF4M/DET/JNGFR03:daqOutput    (e.g. in p 2732, r 12)
     # FXE_XAD_JF1M/DET/RECEIVER-1
+    n_modules = 8
```
takluyver, line 1134: I'd like to avoid hardcoding this, if possible. JUNGFRAU detectors are built from varying numbers of modules - we happen to be looking first at a 4MPx configuration with 8 modules, but the `JF1M` name in the comments suggests we have a 2 module setup, and the [JUNGFRAU description](https://www.psi.ch/en/detectors/jungfrau) mentions 1, 3, 9 and 32 module installations at other facilities.

I think we should either detect the number of modules from the matching sources when the class is created, or let the user specify how many modules they expect in the data. For JUNGFRAU geometry in EXtra-geom, we rely on user input to know how many modules are present:

https://github.com/European-XFEL/EXtra-geom/blob/ac76cc5d71f1f7bb7510f08fdf9f3092dbda3829/extra_geom/detectors.py#L2025-L2028

extra_data/components.py
```diff
@@ -1020,10 +1149,10 @@ def _label_dims(arr):
 
         if ndim_pertrain == 4:
             arr = arr.rename({
-                'dim_0': 'pulse', 'dim_1': 'slow_scan', 'dim_2': 'fast_scan'
+                'dim_0': 'cell', 'dim_1': 'slow_scan', 'dim_2': 'fast_scan'
             })
         elif ndim_pertrain == 2:
-            arr = arr.rename({'dim_0': 'pulse'})
+            arr = arr.rename({'dim_0': 'cell'})
```
takluyver, line 1155: This is technically a breaking API change (the fact you had to change the tests backs this up). It makes sense, and I think we can probably get away with this because `JUNGFRAU` support is fairly new and hopefully not that widely used yet, but normally we'd want to avoid changing something like this once it has been released.

extra_data/components.py
```diff
@@ -1109,7 +1256,7 @@ def identify_multimod_detectors(
     *clses* may be a list of acceptable detector classes to check.
     """
     if clses is None:
-        clses = [AGIPD1M, DSSC1M, LPD1M]
+        clses = detectors.list
```
takluyver, line 1259: This too is technically a breaking change - previously if you used this function on data containing AGIPD & JUNGFRAU detectors, it would find the AGIPD detector, now it can error (if `single=True`). But I'm not too concerned, as this function is probably mostly used by us, building generic tools, rather than scientists who will normally write code for a specific detector.

We might want to adjust the docstring talking about "Various detectors record data in a similar format..." though, because JUNGFRAU is not as similar as AGIPD/LPD/DSSC.

takluyver (commented):
Thanks - I've looked through everything except `write_cxi.py`, which I'll get to tomorrow.

extra_data/write_cxi.py
```diff
@@ -92,46 +134,78 @@ def _map_chunk(self, chunk, chunk_data, target, tgt_ax1, have_data=None):
...
 
         if 'gain' in image_grp:
-            log.info("Identified calibrated data")
+            log.info("Identified calibrated or Jungfrau raw data")
             shape = (self.nframes, self.nmodules) + self.detdata.module_shape
             log.info("Virtual data shape: %r", shape)
 
             layouts = {
-                'data': VLayout(shape, dtype=image_grp['data'].dtype),
+                self.image_label: VLayout(
```
takluyver, line 206: The name `data` in the output is actually part of the [CXI file format](https://raw.githubusercontent.com/cxidb/CXI/master/cxi_file_format.pdf), so I'd suggest keeping it the same regardless of the names in the input.

(I'm not sure whether we're actually producing valid CXI files - CrystFEL just uses whatever HDF5 dataset you tell it to - but as we're calling it CXI, we should try to follow the spec)

extra_data/write_cxi.py
```diff
@@ -147,34 +221,40 @@ def collect_data(self):
...
-                     len(layout.sources), key, filled_pct)
+            log.info(f"Assembled {len(layout.sources):d} chunks for {key:s}, "
+                     f"filling {filled_pct:.2f}% of the hyperslab")
 
         return layouts
 
     def write(self, filename, fillvalues=None):
-        """Write the file on disc to filename
+        """
+        Write the file on disc to filename.
```
takluyver, line 248: You don't need to revert these changes now, but for future reference: please try to do formatting changes in a separate PR from changes that affect what the code is doing. It's less work to review the two kinds of change separately rather than in one big diff.

extra_data/write_cxi.py
```diff
@@ -21,15 +21,21 @@ class VirtualCXIWriter:
...
     """
     def __init__(self, detdata):
         self.detdata = detdata
 
         self.modulenos = sorted(detdata.modno_to_source)
 
-        frame_counts = detdata.frame_counts
+        # The following line could be tricky, but there is no pulseId
+        # in Jungfrau data.
+        self.pulse_id_label = 'image.pulseId'
```
takluyver, line 34: It's slightly surprising that all of the `_label` variables are single words except this one. :confused:

extra_data/write_cxi.py
```diff
@@ -21,15 +21,21 @@ class VirtualCXIWriter:
...
+        self.group_label, self.image_label = detdata._main_data_key.split('.')
+        self.cell_id_label = 'cellId'
```
takluyver, line 36: It would probably be clearer to have a base class which doesn't set these labels, and then two subclasses (one for JUNGFRAU, one for xtdf detectors), which know their respective names, rather than the base class being used directly for xtdf detectors.

There are probably also some complexities which are only needed for xtdf detectors and can be moved to that subclass.

extra_data/write_cxi.py
```diff
@@ -38,31 +44,50 @@ def __init__(self, detdata):
...
-                if chunk.dataset.ndim > 1:
-                    chunk_data = chunk.dataset[:, 0]
-                else:
-                    chunk_data = chunk.dataset
-                self._map_chunk(chunk, chunk_data, pulse_ids, modno)
+            module_id = self.modulenos.index(modno)
+            for chunk in self.data._find_data_chunks(source,
+                                                     self.pulse_id_label):
+                chunk_data = chunk.dataset
+                self._map_chunk(chunk, chunk_data, pulse_ids, module_id)
```
takluyver, line 90: I don't think changing `modno` to `module_id` here is correct. If module 4 is missing from our data and we exclude it, then we should leave a gap at index 4, not shift module 5 to that location.

Looking through the code, I think we've always done the wrong thing for the `module_identifier` dataset in that case - we presumably haven't noticed because CrystFEL doesn't use that.

extra_data/write_cxi.py
```diff
@@ -185,44 +265,49 @@ def write(self, filename, fillvalues=None):
...
             d = f.create_dataset('entry_1/experiment_identifier',
                                  shape=experiment_ids.shape,
                                  dtype=h5py.special_dtype(vlen=str))
             d[:] = experiment_ids
 
-            # pulseId, trainId, cellId are not part of the CXI standard, but
-            # it allows extra data.
+            # pulseId, trainId, cellId are not part of the CXI standard,
+            # but it allows extra data.
             f.create_dataset('entry_1/pulseId', data=pulse_ids)
```
takluyver, line 303: I think we should avoid creating a `pulseId` dataset for JUNGFRAU, given that it's not even claiming to store pulse IDs. This bit isn't part of the CXI format (the format is fairly flexible about adding in extra datasets).

extra_data/write_cxi.py
```diff
@@ -250,10 +335,48 @@ def write(self, filename, fillvalues=None):
...
+    detdata: extra_data.components.JUNGFRAU
+      The detector data interface for the data to gather in this file.
+    """
+    def __init__(self, detdata) -> None:
+        # Check number of cells
+        src = next(iter(detdata.source_to_modno))
+        h5file = detdata.data._source_index[src][0].file
+        image = h5file['INSTRUMENT'][src][
+            detdata._main_data_key.replace('.', '/')]
+        self._n_cells = image.shape[1]
```
takluyver, line 366: There's now a neater way of getting the shape of one entry:

```python
keydata = detdata.data[src, 'data.adc']
n_cells = keydata.entry_shape[0]
```

[entry_shape](https://extra-data.readthedocs.io/en/latest/reading_files.html#extra_data.KeyData.entry_shape) in the docs

extra_data/components.py
```diff
@@ -15,6 +15,28 @@
 MAX_PULSES = 2700
 
 
+def detectors(detector_cls):
```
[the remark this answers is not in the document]
turkot: Sure, thank you a lot for the naming suggestion - I will rename `detectors` decorator to `multimod_detectors`.

extra_data/components.py
turkot: Thank you!) Yes, I understand that full docstrings are not required for protected methods, but I added them mostly for myself while trying to understand how the code works.

extra_data/components.py
turkot: Sure, I'll put `XtdfDetectorBase.trains()` back.
I just thought to reduce repeated code, since `JUNGFRAU` class has its own `trains()` method without `pulses` parameter.

takluyver: Thanks!

My logic here is that `MultimodDetectorBase` doesn't 'know about' selecting pulses, so it shouldn't have parameters that do. If all the subclasses override it anyway, we could just have it `raise NotImplementedError` - but it's only a couple of lines of extra code, so it doesn't matter much.

turkot: :+1:

extra_data/cli/make_virtual_cxi.py
```diff
@@ -56,8 +50,8 @@ def main(argv=None):
              "By default, it is written in the proposal's scratch directory."
     )
     ap.add_argument(
-        '--min-modules', type=int, default=9, metavar='N',
-        help="Include trains where at least N modules have data (default 9)"
+        '--min-modules', type=int, default=4, metavar='N',
+        help="Include trains where at least N modules have data (default 4)"
```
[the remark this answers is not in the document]
turkot: Implemented different '--min-modules' default values for different multi-module detectors in https://github.com/European-XFEL/EXtra-data/commit/156b8bf353b59b107950a2baf566352c061acc4c.

extra_data/cli/make_virtual_cxi.py
turkot: Yes, thank you, implemented in https://github.com/European-XFEL/EXtra-data/commit/156b8bf353b59b107950a2baf566352c061acc4c.

extra_data/components.py
turkot: Implemented in https://github.com/European-XFEL/EXtra-data/commit/156b8bf353b59b107950a2baf566352c061acc4c (not used in 'write_cxi.py' just yet).

extra_data/components.py
turkot: If you would like - we can change it back. I just thought it might produce less confusion in the future if we refrain from labeling memory cell id as pulse id just for Jungfrau data.

extra_data/components.py
turkot: Since we are now using this function to identify multimod detector class in 'cli/make_virtual_cxi.py' this change is required. I've tried to adjust the docstring in https://github.com/European-XFEL/EXtra-data/commit/156b8bf353b59b107950a2baf566352c061acc4c - please let me know your opinion or suggestions.

extra_data/write_cxi.py
turkot: Thank you, this looks much better!

extra_data/cli/make_virtual_cxi.py
```diff
@@ -107,9 +99,26 @@ def main(argv=None):
...
+    det_name = det_class.__name__
+
+    min_modules = args.min_modules
+    if min_modules is None:
+        if det_name in ['AGIPD1M', 'DSSC1M', 'LPD1M']:
+            min_modules = 9
+        elif det_name in ['AGIPD500K']:
+            min_modules = 5
+        else:
+            min_modules = 1
```
takluyver, line 113: Rather than hardcoding class names here, maybe we can use the number of modules from the class if available - something like:

```python
n_modules = getattr(det_class, 'n_modules', 0)
min_modules = (n_modules // 2) + 1
```

extra_data/cli/make_virtual_cxi.py
```diff
@@ -56,8 +41,14 @@ def main(argv=None):
...
-        '--min-modules', type=int, default=9, metavar='N',
-        help="Include trains where at least N modules have data (default 9)"
+        '--min-modules', type=int, default=None, metavar='N',
+        help='Include trains where at least N modules have data (default'
+             ' 9 for AGIPD1M/DSSC1M/LPD1M, 5 for AGIPD500K and 1 otherwise)'
+    )
+    ap.add_argument(
+        '--n-modules', type=int, default=None, metavar='N',
+        help='Number of detector modules in the experiment setup.'
+             ' Should be used only for Jungfrau data.'
```
takluyver, line 51:
```suggestion
             ' Should be used only for JUNGFRAU data.'
```

PSI always capitalise [JUNGFRAU](https://www.psi.ch/en/detectors/jungfrau). It's a bit silly, but after spending many years working on IPython (*not* iPython, as many people write), I try to follow how people choose to write the name of their projects. :slightly_smiling_face:

extra_data/components.py
takluyver: I think you're right - the benefits of getting the label right *hopefully* outweigh the costs of changing it, since this class is fairly new. But it is a risk, and we should be ready to apologise if it turns out we break someone's code. :slightly_smiling_face:

extra_data/write_cxi.py
turkot: I thought to preserve 'self.image_label' as a key for the layout since it is used later on as a key to access data from the HD5 file, currently it is in line 196:
```python3
for name, layout in layouts.items():
    key = '{}.{}'.format(self.group_label, name)
```
but still a virtual dataset is created with the name 'data', currently lines 287-289:
```python3
data = dgrp.create_virtual_dataset(
    'data', layouts[data_label], fillvalue=_fillvalues[data_label]
)
```
but please let me know if you would like also to name a key in layouts as 'data' - then I'll add an if statement to the line 196.

extra_data/write_cxi.py
turkot: Ok, or maybe an additional commit in the same PR after the main PR review finished?

extra_data/write_cxi.py
turkot: Yes, I was following how it was implemented before, but now I've changed it to the single word in: https://github.com/European-XFEL/EXtra-data/commit/cc73dadb190a2c3e831a1fb08259f2752871ee5d.

extra_data/write_cxi.py
turkot: I've tried to introduce such separation into Base, Xtdf and JUNGFRAU classes in https://github.com/European-XFEL/EXtra-data/commit/cc73dadb190a2c3e831a1fb08259f2752871ee5d - please let me know your opinion.

extra_data/write_cxi.py
turkot: Implemented in https://github.com/European-XFEL/EXtra-data/commit/cc73dadb190a2c3e831a1fb08259f2752871ee5d. Please let me know what should we do with `module_identifier`? Right now it looks fine only if all modules are present in the data.

extra_data/write_cxi.py
turkot: Renamed a `pulseId` dataset for JUNGFRAU to `memoryCell`.

extra_data/cli/make_virtual_cxi.py
turkot: Thank you, implemented in https://github.com/European-XFEL/EXtra-data/commit/cc73dadb190a2c3e831a1fb08259f2752871ee5d.

extra_data/write_cxi.py
takluyver: Aha, OK. I was assuming that the keys in `layouts` matched the names the virtual datasets would have in the output file, rather than the names of the source datasets. There's no particular reason that has to be the case, though.

extra_data/write_cxi.py
takluyver: Yup, that should be workable as well.

extra_data/write_cxi.py
takluyver: Thanks. I think `module_identifier` should be like `list(range(n_modules))` for the xtdf detectors, and 1 ... N+1 for JUNGFRAU, i.e. number all of the modules which there's 'space' for, even the ones that aren't included.

extra_data/write_cxi.py
takluyver: OK. This will mean, if I'm following correctly, that it has a regular `memoryCell` dataset plus a virtual `cellId` one which have the same data when you read them? That seems a little bit confusing, but not a big problem.

extra_data/write_cxi.py
```diff
@@ -82,10 +79,10 @@ def collect_pulse_ids(self):
         pulse_ids = np.full((self.nframes, self.nmodules), NO_PULSE_ID,
                             dtype=np.uint64)
 
+        pulse_key = self.group_label + '.' + self.pulse_id_label
         for source, modno in self.detdata.source_to_modno.items():
-            module_id = self.modulenos.index(modno)
-            for chunk in self.data._find_data_chunks(source,
-                                                     self.pulse_id_label):
+            module_id = self._get_module_id(modno)
```
takluyver, line 84: It would be good to have the method in the base class, even if it raises NotImplementedError, because otherwise my IDE complains that it's calling something that doesn't exist:

Screenshot from 2021-05-31 11-42-26

Also, I don't think 'module ID' is a great name, when it's going from the numbers people use to identify the modules to something else. Maybe call it 'module index' instead? index can be abbreviated 'ix' if you need to.

extra_data/write_cxi.py
turkot: Yes, right now it produces `memoryCell` with the shape [n_frames] and `cellId` with the same data with the shape [n_frames x n_modules]. Maybe we should just drop `cellId` for JUNGFRAU? As far as I remember, we either way require `memoryCell` to be the same for each module...

extra_data/write_cxi.py
```diff
@@ -179,60 +176,29 @@ def _map_chunk(self, chunk, chunk_data, target, tgt_ax1, have_data=None):
             chunk_match_start = chunk_match_end
             chunk_tids = chunk_tids[n_match:]
 
-    def collect_data(self):
+    def _map_layouts(self, layouts):
         """
-        Prepare virtual layouts and map them to the virtual sources in
-        the data chunks.
+        Map virtual layouts to the virtual sources in the data chunks.
```
takluyver, line 181: In my mind, this goes the other way - you map sources into the layouts. :thinking:

extra_data/write_cxi.py
```diff
@@ -92,89 +131,96 @@ def _map_chunk(self, chunk, chunk_data, target, tgt_ax1, have_data=None):
...
+                f"{target_tids.shape} != {chunk_tids_target.shape}"
             assert target_tids[0] == chunk_tids[0], \
                 f"{target_tids[0]} != {chunk_tids[0]}"
 
             # How much of this chunk can be mapped in one go?
-            mismatches = (chunk_tids != target_tids).nonzero()[0]
+            mismatches = (chunk_tids_target != target_tids).nonzero()[0]
             if mismatches.size > 0:
-                n_match = mismatches[0]
+                n_match = int(mismatches[0] / self.ncells)
```
takluyver, line 148: BTW, you can do integer division (aka floor division) with the `//` operator.

extra_data/write_cxi.py
```diff
@@ -92,89 +131,96 @@ def _map_chunk(self, chunk, chunk_data, target, tgt_ax1, have_data=None):
...
+            tgt_end = tgt_start + (n_match*self.ncells)
             matched = chunk_data[chunk_match_start:chunk_match_end]
-            target[tgt_start : tgt_start+n_match, tgt_ax1] = matched
+            if self.ncells == 1:
+                # In some cases, there's an extra dimension of length 1.
+                # E.g. JUNGFRAU data with 1 memory cell per train or
+                # AGIPD/LPD raw data.
+                # Also, 'VirtualSource' object has no attribute 'ndim'.
+                if (len(matched.shape) > 1 and matched.shape[1] == 1):
+                    target[tgt_start:tgt_end, tgt_ax1] = matched[:, 0]
```
takluyver, line 162: Unfortunately, I don't think this will quite work correctly with VirtualSource objects. The selection mechanism just applies one selection, so doing a second selection like this will replace the previous selection rather than combining with it.

I think this would only come up if someone tried to assemble a virtual CXI file from LPD or DSSC raw data, so maybe it's not worth worrying about. But if not, maybe we should fail up front in those cases rather than trying something that's broken.

extra_data/write_cxi.py
```diff
@@ -250,10 +301,174 @@ def write(self, filename, fillvalues=None):
...
+    This code writes version 1.5 CXI files.
+
+    Parameters
+    ----------
+
+    detdata: extra_data.components.XtdfDetectorBase
+      The detector data interface for the data to gather in this file.
+    """
+    def __init__(self, detdata) -> None:
+        self._n_cells = 1
```
takluyver, line 337: This is a bit confusing, because these detectors all do have multiple memory cells. Maybe call the number something like 'cells per entry'? 'Entry' is the term I've used elsewhere for each part of the data along the first axis in the files.

extra_data/write_cxi.py
takluyver: Aha, I had missed the difference in their shape.

I *think* it's reasonable to check that the memory cell numbers match up across modules, but this kind of depends on the details of how the detector works. For *pulse* IDs, this is an important check because we want to be sure that a single frame from the virtual CXI file is all coming from the same pulse. In theory, different modules could store the same pulse in different memory cells, but I think the detectors are always configured so all modules act in the same way, in which case this is a useful sanity check.

extra_data/write_cxi.py
turkot: Thank you for the suggestion, implemented in https://github.com/European-XFEL/EXtra-data/commit/2758b23ae00008c8ca120cdc6ee6ae01cb6a04dd.

extra_data/write_cxi.py
turkot: Implemented in https://github.com/European-XFEL/EXtra-data/commit/6582e9954511eb0bc752851d8f64a31bb7ed1348 and https://github.com/European-XFEL/EXtra-data/commit/2758b23ae00008c8ca120cdc6ee6ae01cb6a04dd.

extra_data/write_cxi.py
turkot: Thank you, my English is far from perfect, so I appreciate corrections) Modified in https://github.com/European-XFEL/EXtra-data/commit/6582e9954511eb0bc752851d8f64a31bb7ed1348.

extra_data/write_cxi.py
turkot: Thank you, changed in https://github.com/European-XFEL/EXtra-data/commit/6582e9954511eb0bc752851d8f64a31bb7ed1348.

extra_data/write_cxi.py
turkot: Yes, thank you for a very valuable comment and suggestions on improving VDS readout performance, implemented in https://github.com/European-XFEL/EXtra-data/commit/6582e9954511eb0bc752851d8f64a31bb7ed1348.

extra_data/write_cxi.py
turkot: Thank you, renamed in https://github.com/European-XFEL/EXtra-data/commit/6582e9954511eb0bc752851d8f64a31bb7ed1348.

extra_data/write_cxi.py
```diff
@@ -38,31 +40,65 @@ def __init__(self, detdata):
...
 
     @property
     def nmodules(self):
+        """
+        Number of detector modules.
+        """
         return self.detdata.n_modules
 
+    @property
+    def ncells(self):
```
takluyver, line 54: I've got the same naming suggestion as you already implemented for the private attribute (cells_per_entry). But also, I wouldn't bother with making a private attribute and wrapping it in a property - you can just make `self.cells_per_entry` as a normal attribute.

extra_data/write_cxi.py
```diff
@@ -38,31 +40,65 @@ def __init__(self, detdata):
...
 
-        # cumulative sum gives the end of each train, subtract to get start
+        # Cumulative sum gives the end of each train, subtract to get start
         self.train_id_to_ix = frame_counts.cumsum() - frame_counts
 
     @property
     def nmodules(self):
+        """
+        Number of detector modules.
+        """
```
takluyver, line 50:
```suggestion
        """Number of detector modules."""
```

This is only a suggestion - I'm not going to get into battles over code style. But I think there's some value in using fewer lines where it doesn't hurt readability, because it means you can see more code on screen at once.

extra_data/write_cxi.py
```diff
@@ -21,15 +24,14 @@ class VirtualCXIWriter:
     Parameters
     ----------
 
-    detdata: extra_data.components.MPxDetectorBase
+    detdata: extra_data.components.XtdfDetectorBase
```
takluyver, line 27: We are back to this base class taking a `MPxDetectorBase` again now, I think.

extra_data/cli/make_virtual_cxi.py
```diff
@@ -67,8 +80,13 @@ def main(argv=None):
...
-        help="Exclude suspect trains. This tries to avoid some issues with incorrect train IDs in the data, "
-             "but may mean less data is available."
+        help='Exclude suspect trains. This tries to avoid some issues with'
+             ' incorrect train IDs in the data, but may mean less data is'
+             ' available.'
+    )
+    ap.add_argument(
+        '-seq', '--per-sequence', action='store_true',
+        help='Create separate CXI files for each data sequence.'
     )
```
takluyver, line 90: Are we confident this is actually useful? My memory of the users doing something similar before was that they were making each CXI file as the corresponding sequence was ready, so they'd want the ability to pass in a sequence number, rather than make CXI files for all sequence numbers in the directory.

If we're not pretty sure it's what users want, let's leave it out for now. Once the machinery is ready, it's easy to write a custom script to do something like this. Extra options have a cost both for users (more to read and figure out) and us (more to test and support).

takluyver:
Thanks @turkot - I think this is mostly ready now. Most of what I'm asking about now is naming & style. I'm not sure about the new `--per-sequence` option, though.

takluyver:
Oh, one more thing, don't forget to update the docs (`cli.rst`) with the changes to the command.

takluyver:
And another! We should have at least a basic test of making a virtual CXI file from JUNGFRAU data. The pieces we need should all be there in `test_components` already.

extra_data/write_cxi.py
```diff
@@ -250,10 +317,166 @@ def write(self, filename, fillvalues=None):
...
+        src = next(iter(detdata.source_to_modno))
+        keydata = detdata.data[src, 'data.adc']
+        self._cells_per_entry = keydata.entry_shape[0]
+        self.pulse_id_label = 'memoryCell'
+        self.cell_id_label = 'memoryCell'
+
+        super().__init__(detdata)
+
+        # For JUNGFRAU detectors modules are numbered from 1
+        self.modulenos = [i+1 for i in range(self.nmodules)]
```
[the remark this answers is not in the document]
turkot: Sure, modified in https://github.com/European-XFEL/EXtra-data/commit/a8c2248ec48f2cbf6a4e7d79afc0b6981c49cb98.

extra_data/write_cxi.py
turkot: Thank you, changed in https://github.com/European-XFEL/EXtra-data/commit/a8c2248ec48f2cbf6a4e7d79afc0b6981c49cb98.

extra_data/write_cxi.py
turkot: Yes, changed in https://github.com/European-XFEL/EXtra-data/commit/a8c2248ec48f2cbf6a4e7d79afc0b6981c49cb98 :)

extra_data/cli/make_virtual_cxi.py
turkot: Sure, thank you, removed in https://github.com/European-XFEL/EXtra-data/commit/5b0b07650f36da21278bb8f04a32cbc701e2cc24.

turkot:
Brief update to the documentation added with https://github.com/European-XFEL/EXtra-data/commit/7323748bead6bee4b7afa4392ce46b459557b6b1 and tests of making a virtual CXI file from JUNGFRAU data included in https://github.com/European-XFEL/EXtra-data/commit/046db61164e57e6a32897c490772dd3ca3359313.

turkot:
Commit https://github.com/European-XFEL/EXtra-data/pull/62/commits/05390051ee6efb682049a54df87f34908d22bf69 contains only style suggestions to comply with PEP8 and make pycodestyle 'almost' happy - pycodestyle is wrong on the white spaces around colon in a slice. While code lines have to fit in 79 characters, docstrings and comments should fit in 72. Please feel free to accept to reject any suggestions from this commit.

takluyver:
Thanks Oleksii! I agree with some of the style changes, but I'm going to knock that commit off the PR for now, because it looks like quite a lot of it is line-length changes. I usually aim for 80 characters, rather than 79, and I often treat it as a soft limit - going a couple of characters over often seems better than making an extra line.

I'm not totally against automated code formatting (just partially ;-), but if we're going to do it, I think we should agree on what tool and/or rules we're using. E.g. [black](https://github.com/psf/black), which is very popular now, uses a default line length of 88 characters.

extra_data/tests/test_components.py
```diff
@@ -363,6 +363,53 @@ def test_write_virtual_cxi_some_modules(mock_spb_proc_run, tmpdir):
...
+
+    test_file = osp.join(str(tmpdir), 'test.cxi')
+    det.write_virtual_cxi(test_file)
+    assert_isfile(test_file)
+
+    with h5py.File(test_file, 'r') as f:
+        det_grp = f['entry_1/instrument_1/detector_1']
+        ds = det_grp['data']
+        assert ds.shape[1:] == (8, 512, 1024)
+        assert (det_grp['module_identifier'][:] == np.arange(1,9)).all()
```
takluyver, line 411:
```suggestion
        np.testing.assert_array_equal(det_grp['module_identifier'][:], np.arange(1,9))
```

extra_data/cli/make_virtual_cxi.py
```diff
@@ -106,11 +97,17 @@ def main(argv=None):
...
-        sys.exit(f"No {_detectors()} sources found in {run_dir}")
+    inc_suspect = not args.exc_suspect_trains
+    run = RunDirectory(run_dir, inc_suspect_trains=inc_suspect)
+
+    _, det_class = identify_multimod_detectors(run, single=True)
+
+    min_modules = args.min_modules
+    if min_modules is None:
+        det_n_modules = getattr(det_class, 'n_modules', 0)
+        min_modules = (det_n_modules // 2) + 1
```
takluyver, line 108: As we're documenting that this is half + 1, we should probably respect `--n-modules` for JUNGFRAU.

extra_data/components.py
```diff
@@ -131,6 +153,8 @@ def _identify_sources(self, data, detector_name, modules=None):
             if m and m.group('detname') == detector_name:
                 source_to_modno[source] = int(m.group('modno'))
 
+        self.max_module = max(source_to_modno.values())
```
takluyver, line 156: This is only used for JUNGFRAU - we may as well just find it where it's used.

extra_data/components.py
```diff
@@ -262,6 +284,13 @@ class XtdfDetectorBase(MultimodDetectorBase):
     n_modules = 16
     _main_data_key = 'image.data'
 
+    def __init__(self, data: DataCollection, detector_name=None, modules=None,
+                 *, min_modules=1, n_modules=None):
+        if n_modules is not None:
+            raise NotImplementedError(
+                '--n-modules option is not available for AGIPD/DSSC/LPD data')
```
takluyver, line 291: I'm not super keen on having a parameter in the signature which can't actually be used. It works, but it feels a bit clumsy.

extra_data/write_cxi.py
```diff
@@ -20,16 +23,14 @@ class VirtualCXIWriter:
...
       The detector data interface for the data to gather in this file.
     """
     def __init__(self, detdata):
         self.detdata = detdata
+        self.group_label, self.image_label = detdata._main_data_key.split('.')
 
-        self.modulenos = sorted(detdata.modno_to_source)
-
-        frame_counts = detdata.frame_counts
+        frame_counts = detdata.frame_counts * self.cells_per_entry
```
takluyver, line 33: Pycharm gives me quite a few warnings because attributes like `.cells_per_entry` are used in the base class but only defined in the child classes. I might give them some defaults in the base class to help tools like this follow the code (it can also help human readers).

turkot:
> I'm not totally against automated code formatting (just partially ;-), but ...

It was manual... but I agree that often a bit longer line is better readable than split.

extra_data/tests/test_components.py
turkot: Cool, thank you!)

takluyver:
I've got this @turkot - enjoy your holiday!

extra_data/components.py
```diff
@@ -1159,7 +1157,7 @@ def __init__(self, data: DataCollection, detector_name=None, modules=None,
             self.n_modules = int(n_modules)
         else:
             # For JUNGFRAU modules are indexed from 1
-            self.n_modules = self.max_module
+            self.n_modules = max(self.source_to_modno.values())
```
turkot: But

turkot: Sorry, now I see you have fixed it in the later commit https://github.com/European-XFEL/EXtra-data/pull/62/commits/e2565ff8cfc87d9aed758a5f1e5e0e28027e88ad =)

Issue #57, opened 2020-05-15
Enable Dask arrays from very large numbers of files


takluyver:
From discussion on #48, our solutions to manage the number of open files won't work if we need to create a Dask array covering thousands of files (e.g. combining data from all detector modules in a long run). We let Dask store references to h5py `Dataset` objects, which rely on an open file underneath them. Either we leave the files open and eventually hit the limit (4096 on Maxwell), or close them and break those Dataset objects.

It seems likely that the solution will be to make an array-like type which stores a filename and opens the file on demand. According to the [dask.array API docs](https://docs.dask.org/en/latest/array-api.html#dask.array.from_array), the object we give it

> must have a `.shape`, `.ndim`, `.dtype` and support numpy-style slicing.

This could have an additional benefit if we can solve #54 as well: it should be possible for the control code to create a Dask array without opening the data files at all, and have each worker only open the specific files it needs.

PR #63, opened 2020-05-29
Create Dask arrays with DatasetProxy objects

takluyver:
Avoid storing or trying to pickle an h5py Dataset object. cc @philsmt - this does appear to fix the pickling errors for me.

Closes gh-57

### DIFF
--- a/docs/dask_averaging.ipynb
+++ b/docs/dask_averaging.ipynb
@@ -40,7 +40,7 @@
     {
      "data": {
       "application/vnd.jupyter.widget-view+json": {
-       "model_id": "e4e115d0e0db4cd3bbf59a8e1e5556b5",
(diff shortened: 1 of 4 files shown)


takluyver:
I've run through the Dask example with this, and it still works as before.

I added the `local_directory` option while I had it open, which I've been meaning to do for a while. This doesn't seem to make a big difference in this example - I don't think the workers are actually using it in this case. But it might make a difference if they do try to use it. See: https://jobqueue.dask.org/en/latest/configuration-setup.html#local-storage

philsmt:
Yup, I can confirm it works with this branch. Thanks!

Now, the performance with using processes is abysmal of course without explicit shared memory, as it probably involves copying around lots of data. But I will play around with the API a bit more, now I wondered whether we can at least re-use the clients.map machinery together with our own shared memory management.

LGTM. Bonus points for typing!

PR #65, opened 2020-06-03
Documentation refactoring

tmichela:
make the `installation` and `quickstart` sections visible in the tree view.

### DIFF
--- a/docs/get_started.rst
+++ b/docs/get_started.rst
@@ -1,0 +1,51 @@
+Installation
+------------
+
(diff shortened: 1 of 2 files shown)


takluyver:
Can you make a case for this? I had deliberately put that information on the landing page so you have the essential summary right there, without needing an extra click.

I'm fine with having the licensing information in the docs, but I don't think it needs to be so prominent. Most users won't be distributing copies of EXtra-data, so they don't need to think about the license, and I imagine that anyone used to working with open source software can find it easily enough anyway.

PR #66, opened 2020-06-05
Add file locality documentation

egorsobolev:


docs/reading_files.rst
```diff
@@ -268,3 +268,34 @@ You only need these details if you think caching may be causing problems.
 
 JSON was chosen as it can be easily inspected manually, and it's reasonably
 efficient to load the entire file.
+
+File Reading Issues
+-------------------
+
+Files at European XFEL storage migrate over time from GPFS to PNFS. The data 
+on PNFS usually always available for reading. But sometimes, reading may require
+staging from the tape to dCache. If there is staging queue, the operation can take
```
takluyver, line 277:
```suggestion
staging from the tape to disk. If there is a staging queue, the operation can take
```

We tend to talk about the whole of `/pnfs` as dCache, both the disks and tape storage. I'm not sure if that's technically accurate, but it works pretty well for our purposes, so I'd rather not use it differently here.

docs/reading_files.rst
```diff
@@ -268,3 +268,34 @@ You only need these details if you think caching may be causing problems.
...
+an indefinitely long time (days or even weeks) and any IO operations will be 
+blocked for this time.
+
+To determine the files which require staging or are lost, use the script::
+
+    extra-date-locality <run directory>
```
takluyver, line 283:
```suggestion
    extra-data-locality <run directory>
```

docs/reading_files.rst
```diff
@@ -268,3 +268,34 @@ You only need these details if you think caching may be causing problems.
 
 JSON was chosen as it can be easily inspected manually, and it's reasonably
 efficient to load the entire file.
+
+File Reading Issues
+-------------------
+
+Files at European XFEL storage migrate over time from GPFS to PNFS. The data
```
takluyver, line 275:
```suggestion
Files at European XFEL storage migrate over time from GPFS (designed for fast acess) to PNFS (designed for archiving). The data 
```

docs/reading_files.rst
```diff
@@ -268,3 +268,34 @@ You only need these details if you think caching may be causing problems.
 
 JSON was chosen as it can be easily inspected manually, and it's reasonably
 efficient to load the entire file.
+
+File Reading Issues
+-------------------
+
+Files at European XFEL storage migrate over time from GPFS to PNFS. The data 
+on PNFS usually always available for reading. But sometimes, reading may require
```
takluyver, line 276:
```suggestion
on PNFS is usually always available for reading. But sometimes, this may require
```

"Data is" vs "Data are" is a matter of debate in English. If we want to avoid that altogether, we can say "Files... are"

docs/reading_files.rst
```diff
@@ -268,3 +268,34 @@ You only need these details if you think caching may be causing problems.
...
+staging from the tape to dCache. If there is staging queue, the operation can take
+an indefinitely long time (days or even weeks) and any IO operations will be 
+blocked for this time.
+
+To determine the files which require staging or are lost, use the script::
+
+    extra-date-locality <run directory>
+
+It returns a list of files which are currently located only on slow media for some
+reasons and, separetelly, which are lost at all.
```
takluyver, line 286:
```suggestion
reasons and, separately, any which have been lost.
```

docs/reading_files.rst
```diff
@@ -268,3 +268,34 @@ You only need these details if you think caching may be causing problems.
...
+
+If the files are not essential for analysis, then they can be filtered out using 
+filter :func:`ls_ondisk` from :mod:`extra_data.locality`::
```
takluyver, line 289:
```suggestion
filter :func:`lc_ondisk` from :mod:`extra_data.locality`::
```

docs/reading_files.rst
```diff
@@ -268,3 +268,34 @@ You only need these details if you think caching may be causing problems.
...
+
+    from extra_data.locality import lc_ondisk
+    run = open_run(proposal=700000, run=1, file_filter=lc_ondisk)
+
+``file_filter`` must be a callable which takes a list as a single argument and
+returns filtered list.
+
+**Note: the reading of file locality on PNFS very expensive operation.
```
takluyver, line 297:
```suggestion
**Note: Reading the file locality on PNFS is an expensive operation.
```

'Very' seems misplaced - it's still cheap compared to loading data from tape, for instance. :-)

docs/reading_files.rst
```diff
@@ -268,3 +268,34 @@ You only need these details if you think caching may be causing problems.
...
+Use it only as a last resort.**
+
+If problem files are essential for analysis, contact `IT service <mailto:it-support@xfel.eu>`_
```
takluyver, line 300:
```suggestion
If unavailable files are essential for analysis, contact `IT service <mailto:it-support@xfel.eu>`_
```

I think we should make it clear that data on tape is not a problem from the facility's point of view - it's meant to happen, and you can still access it, you just need to be patient.

docs/reading_files.rst
```diff
@@ -268,3 +268,34 @@ You only need these details if you think caching may be causing problems.
 
 JSON was chosen as it can be easily inspected manually, and it's reasonably
 efficient to load the entire file.
+
+File Reading Issues
+-------------------
```
takluyver, line 273:
```suggestion
Issues reading archived data
----------------------------
```

docs/reading_files.rst
takluyver: I learned yesterday that for the time being, ITDM is keeping all raw data in the disk cache, and we can let them know if that's not the case. I don't know how long that holds for, though, and it doesn't cover all the calibrated data.

egorsobolev: That the reason why I would like to encourage users to give feedback to ITDM every time. It may give them a clue how dCache issues affect on users. What do you think about following:

```suggestion
If you find any files which are located only on tape or unavailable, please let know to
`ITDM <mailto:it-support@xfel.eu>`_. If you need these files for analysis mentioned
that explicitly.
```
I agree with other suggestions. Thanks.

takluyver: That sounds OK. We can always adjust it later if IT complain that too many people are contacting them.

takluyver:
Thanks :-)

PR #67, opened 2020-06-08
Root test fix

RobertRosca:
Minor change to the `test_permission` test. Checks if user is root, and if it is then euid is set to 1000 for the test and reverted back at the end.

Could we make a patch release to 1.2.1 for this test so that I can include the fix in my spack package?

### DIFF
--- a/extra_data/tests/test_reader_mockdata.py
+++ b/extra_data/tests/test_reader_mockdata.py
@@ -615,6 +615,11 @@ def test_open_file(mock_sa3_control_data):
 
 
 def test_permission():
+    #  If users is root then `PermissionError` won't be raised, so set the euid
+    #  to standard user so the tests work correctly
+    if os.geteuid() == 0:
+        os.seteuid(1000)
+
     d = mkdtemp()
     os.chmod(d, not stat.S_IRUSR)
     with pytest.raises(PermissionError) as excinfo:
--- a/extra_data/tests/test_reader_mockdata.py
+++ b/extra_data/tests/test_reader_mockdata.py
@@ -617,9 +617,11 @@ def test_open_file(mock_sa3_control_data):
 def test_permission():
     #  If users is root then `PermissionError` won't be raised, set the euid to
     #  standard user so the tests work correctly
-    original_euid = os.geteuid()
-    if original_euid == 0:
-        os.seteuid(1000)
+    euid_was_root = False
+    if hasattr(os, 'seteuid'):
+        if os.geteuid() == 0:
+            os.seteuid(1000)
+            euid_was_root = True
 
     d = mkdtemp()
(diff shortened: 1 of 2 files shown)


extra_data/tests/test_reader_mockdata.py
```diff
@@ -615,13 +615,24 @@ def test_open_file(mock_sa3_control_data):
 
 
 def test_permission():
+    #  If users is root then `PermissionError` won't be raised, set the euid to
+    #  standard user so the tests work correctly
+    euid_was_root = False
+    if hasattr(os, 'seteuid'):
+        if os.geteuid() == 0:
+            os.seteuid(1000)
```
takluyver, line 623: Somewhat to my surprise, this appears to work (at least on Linux) even with a UID that doesn't exist.

extra_data/tests/test_reader_mockdata.py
```diff
@@ -615,13 +615,24 @@ def test_open_file(mock_sa3_control_data):
...
+
     d = mkdtemp()
     os.chmod(d, not stat.S_IRUSR)
     with pytest.raises(PermissionError) as excinfo:
         run = RunDirectory(d)
     assert "Permission denied" in str(excinfo.value)
     assert d in str(excinfo.value)
 
+    if euid_was_root:
+        os.seteuid(0)
```
takluyver, line 634: This should be in a `finally:` block, so it gets reset even if the test fails. Or make it a pytest fixture.

But to be honest, it may be simpler to just skip the test if run as root. It's getting tested regularly as non-root, and it's not a particularly critical test.

RobertRosca: Made it so that the test is skipped just skipped if euid is 0 now

extra_data/tests/test_reader_mockdata.py
```diff
@@ -615,6 +615,11 @@ def test_open_file(mock_sa3_control_data):
 
 
 def test_permission():
+    #  If users is root then `PermissionError` won't be raised, set the euid to
+    #  standard user so the tests work correctly
```
takluyver, line 619: And the comment ;-)

tmichela:
Minor suggestion, I prefer to have the skip condition as decorator, to separate clearly from the test itself:
```python
@pytest.mark.skipif(hasattr(os, 'geteuid') and os.geteuid() == 0,
                    reason="cannot run permission tests as root")
def test_permission():
...
```

RobertRosca:
Huh neat, didn't know that was a thing, cheers

takluyver:
Thanks, looks good. I suspect we'll also want to skip it on Windows if we ever test there, but we can cross that bridge when we get to it.

PR #68, opened 2020-06-09
Allow slicing without by_index wrapper

takluyver:
Following discussion with @philsmt, I think we can make common slicing operations more convenient by allowing slice objects (easily made with [`np.s_`](https://numpy.org/doc/stable/reference/generated/numpy.s_.html)) or arrays where we previously required `by_index` objects.

This applies to:

- `run.select_trains()`
- `run.get_array()` (*roi* parameter)
- Various methods on big detector access classes to select ranges of pulses.

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -44,7 +44,7 @@ def _guess_axes(data, train_pulse_ids):
 def _check_pulse_selection(pulses):
     """Check and normalise a pulse selection"""
     if not isinstance(pulses, (by_id, by_index)):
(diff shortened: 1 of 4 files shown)


philsmt:
There seems to be no strong opinion on this one. Given our previous discussion, looks good to me.  It might be a good one to get this in before `KeyData` and rebase the latter on top of it. `DataCollection.select_trains()` was in dire need of a refactor, but that's already happening in the other PR.

Issue #55, opened 2020-05-06
Objects for individual sources or datasets?


takluyver:
The `DataCollection` object is the centrepiece of EXtra-data, and has a tendency to keep growing. Many of its methods take `source, key` parameters to refer to a specific dataset, e.g. `get_array()` or `get_data_counts()`. There are others possible - e.g. how do you find the dtype or entry shape for a given key? A couple take just `source`, e.g. `keys_for_source()`.

Should we add new types to represent an individual source or an individual dataset in the available data? That would allow APIs maybe like this:

```python
run.keys_for_source("SA3_XTD10_PES/ADC/1:network")  # Current
run["SA3_XTD10_PES/ADC/1:network"].keys()  # Hypothetical

# Current:
run.get_array("SA3_XTD10_PES/ADC/1:network", "digitizers.channel_4_A.raw.samples")
# Hypothetical:
run["SA3_XTD10_PES/ADC/1:network", "digitizers.channel_4_A.raw.samples"].get_array()
```

And it would give a convenient place to expose things like the dtype & entry shape of a key, without awkward methods on DataCollection.

My reservations are:

- Do more classes make it harder to understand? I recall that one of the original goals was to use basic data structures like dicts as far as possible.
- The naming challenge - I've referred to 'sources' and 'datasets' here, but e.g. 'dataset' already means one thing in HDF5 and a different thing in Xarray, so that's not a great name. 'Source' also has many meanings, although not so immediately adjacent.

@tmichela @philsmt & anyone else who's interested: thoughts? This is for discussion at the moment, I'm not asking anyone to start coding.

philsmt:
Oh, that is beautiful.

As a matter of fact, I encountered this issue in my mapAPI-experiments (#32) for determining the dtype just as you describe. Furthermore, there are slight discrepancies in the naming and possible inputs already, notwithstanding the complex machinery behind wildcard matching and possible types in the select*() calls.

Indexing could make the API much more straightforward and clear in what you mean by a particular source, i.e. applying to a source as a whole, several sources by wildcard or list, keys... one might add even train slices (and at some point pulse slices) as a third tuple component. I may be too enthusiastic here whether this can be implemented all the way like I'm dreaming of it right now, but using indexing might result in a beautiful Pythonic API that "just works". But I'm a romantic when it comes to expressive APIs hiding boilerplate behind operators.

[...]

PR #70, opened 2020-06-25
Add KeyData interface

takluyver:
As outlined in #55, this would allow for things like:

```python
# Current:
run.get_array("SA3_XTD10_PES/ADC/1:network", "digitizers.channel_4_A.raw.samples")
# Hypothetical:
run["SA3_XTD10_PES/ADC/1:network", "digitizers.channel_4_A.raw.samples"].xarray()
```

Questions:

- Should this also have a method like `.select_trains()`, duplicating some functionality of `DataCollection`? Or is it OK to do `run.select_trains(...)[source, key]`?
- Should this also have methods to get data by train, including iterating over trains or getting a train by ID?
- Should getting the KeyData object (doing `run[source, key]`) look at all of the sequence files to know how much data is available & for which trains, or should it only do this on demand? For 'get array' style operations, it's simplest to examine all the chunks up front, but if you're just getting the dtype or the entry shape, or getting data train-by-train, we could just open one file.
  - If we do use virtual overview files (#69), this is less of an issue, because we only need to read metadata & indexes up front, and this is all in the overview file.

### DIFF
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -20,7 +20,6 @@ class KeyData:
         Yields DataChunk objects.
         """
         for file in self.files:
-            file_has_data = False
             firsts, counts = file.get_index(self.source, self._key_group)
 
             # Of trains in this file, which are in selection
@@ -28,19 +27,10 @@ class KeyData:
 
             # Assemble contiguous chunks of data from this file
             for _from, _to in contiguous_regions(selected):
-                file_has_data = True
                 yield DataChunk(file, self.source, self.key,
                                 first=firsts[_from],
                                 train_ids=file.train_ids[_from:_to],
                                 counts=counts[_from:_to],
-                                )
-
-            if not file_has_data:
-                # Make an empty chunk to let e.g. get_array find the shape
-                yield DataChunk(file, self.source, self.key,
-                                first=np.uint64(0),
-                                train_ids=file.train_ids[:0],
-                                counts=counts[:0],
                                 )
 
     @property
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -1,0 +1,153 @@
+import numpy as np
+
+class KeyData:
+    def __init__(self, source, key, data_chunks, section):
+        self.source = source
+        self.key = key
+        self.section = section
+
+        dim0 = 0
+        entry_shapes, dtypes = set(), set()
+        for chunk in data_chunks:
+            dim0 += chunk.total_count
+            entry_shapes.add(chunk.dataset.shape[1:])
+            dtypes.add(chunk.dataset.dtype)
+
+        if len(entry_shapes) > 1:
+            raise Exception("Mismatched data shapes: {}".format(entry_shapes))
+        if len(dtypes) > 1:
+            raise Exception("Mismatched dtypes: {}".format(dtypes))
+
+        self.dtype = dtypes.pop()
+        self.entry_shape = entry_shapes.pop()
+        self.shape = (dim0,) + self.entry_shape
+        self.ndim = len(self.shape)
+        # Sort chunks by train ID and discard any empty ones
+        self._data_chunks = sorted(
+            [c for c in data_chunks if c.total_count > 0],
+            key=lambda c: c.train_ids[0]
+        )
+        if self._data_chunks:
+            self.train_ids = np.concatenate([c.train_ids for c in self._data_chunks])
+        else:
+            self.train_ids = np.zeros(shape=(0,), dtype=np.uint64)
+
+    def __repr__(self):
+        return f"<extra_data.KeyData source={self.source!r} key={self.key!r} " \
+               f"for {len(self.train_ids)} trains"
+
+    @property
+    def _key_group(self):
+        """The part of the key needed to look up index data"""
+        if self.section == 'INSTRUMENT':
+            return self.key.partition('.')[0]
+        else:
+            return ''
+
+    @property
+    def _data_path(self):
+        return f"/{self.section}/{self.source}/{self.key.replace('.', '/')}"
+
+    def ndarray(self, roi=()):
+        if not isinstance(roi, tuple):
+            roi = (roi,)
+
+        # Find the shape of the array with the ROI applied
+        roi_dummy = np.zeros((0,) + self.entry_shape) # extra 0 dim: use less memory
+        roi_shape = roi_dummy[np.index_exp[:] + roi].shape[1:]
+
+        out = np.empty(self.shape[:1] + roi_shape, dtype=self.dtype)
+
+        # Read the data from each chunk into the result array
+        dest_cursor = 0
+        for chunk in self._data_chunks:
+            dest_chunk_end = dest_cursor + chunk.total_count
+
+            slices = (chunk.slice,) + roi
+            chunk.dataset.read_direct(
+                out[dest_cursor:dest_chunk_end], source_sel=slices
+            )
+            dest_cursor = dest_chunk_end
+
+        return out
+
+    def _trainid_index(self):
+        """A 1D array of train IDs, corresponding to self.shape[0]"""
+        chunks_trainids = [
+            np.repeat(chunk.train_ids, chunk.counts.astype(np.intp))
+            for chunk in self._data_chunks
+        ]
+        return np.concatenate(chunks_trainids)
+
+    def xarray(self, extra_dims=None, roi=()):
+        import xarray
+
+        ndarr = self.ndarray(roi=roi)
+
+        # Dimension labels
+        if extra_dims is None:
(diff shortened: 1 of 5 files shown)


extra_data/keydata.py
```diff
@@ -0,0 +1,153 @@
...
+    @property
+    def _key_group(self):
+        """The part of the key needed to look up index data"""
+        if self.section == 'INSTRUMENT':
+            return self.key.partition('.')[0]
+        else:
+            return ''
+
+    @property
+    def _data_path(self):
```
philsmt, line 48: Why advertise this property as private?
I couldn't find any internal use for it, however, at the same time it might be useful to have at times.

takluyver: I've got into the habit of marking things private unless I've got a good justification for making them public - a smaller public API means fewer distractions for users working out what they need, and makes changes easier in the future, because it's more reasonable to break private APIs. But it's not necessarily a good thing overall, and we could make this public.

The lack of internal use is related to my third question above - should `run[source, key]` find all data chunks, or hand `KeyData` the files and let it look up chunks on demand? I started with the latter, switched to the former, and may yet switch back.

takluyver:
I've roughly drafted what the code would look like if we passed `KeyData` the relevant `FileAccess` objects and let it look up data chunks on demand. [Comparison here](https://github.com/European-XFEL/EXtra-data/compare/keydata...keydata-chunks-on-demand).

That approach makes it quicker to do certain operations which only need one file - getting `.dtype` or `.entry_shape`, or reading data for only one train.

philsmt:
Wow, I managed to completely ignore your initial question. Sorry about that.

> * Should this also have a method like `.select_trains()`, duplicating some functionality of `DataCollection`? Or is it OK to do `run.select_trains(...)[source, key]`?

I can see quite a few opportunities for this. You create your KeyData object and pass it along to some processing code, who does not need to care from which source/key it comes from. But it can still slice the trains nicely. Personally, I could also see this as an **optional** 3rd item in the index tuple, i.e. `run['SA3_XTD10_XGM/XGM/DOOCS', 'beamPosition.iX', :10].ndarray()`

> * Should this also have methods to get data by train, including iterating over trains or getting a train by ID?

Yes, definitely! Similar to the case above, I regularly like to get sources/key away upfront and then work with more discrete objects. As a small tangent here: at Hans' PBS talk, instruments expressed their dissatisfaction with long device IDs and pipeline keys. Being able to abstract those away as early as possible could go a long way. And in the long run, you can just make a long mapping of small mnemonics to device-key tuples.

> * Should getting the KeyData object (doing `run[source, key]`) look at all of the sequence files to know how much data is available & for which trains, or should it only do this on demand? For 'get array' style operations, it's simplest to examine all the chunks up front, but if you're just getting the dtype or the entry shape, or getting data train-by-train, we could just open one file.

I think we should try to keep EXtra-data's spirit in the direction of lazy loading as much as possible. Create objects, and only actually read potentially large amounts of data when necessary. In this line of thinking, dtype and entry shape should be as fast as possible, i.e. the KeyData object itself, too. I can see lots of applications for this, e.g. in EXtra-pasha. Currently, I do something like `next(run.select_trains(np.s_[:1]).select(source, key).trains())[1][source][key].shape[1:]` just to get the proper shape for allocations.

>   * If we do use virtual overview files (#69), this is less of an issue, because we only need to read metadata & indexes up front, and this is all in the overview file.

Then let's make it a potential special case, more advertisement for VOFs.

takluyver:
Thanks Philipp! I'll work today on adding some more functionality to this.

I was also coming round to favouring the lazy approach, as I realised that we can get rid of the rather ugly workaround which tries to ensure that at least one 'chunk' is always found, even if it's empty.

extra_data/keydata.py
takluyver: This is now public as `hdf5_data_path`.

takluyver:
OK, individual `KeyData` objects now have `.select_trains()`, `.train_from_id()`, `.train_from_index()` and `.trains()` methods, all similar to those on DataCollection. `.select_trains()` returns a new KeyData object, the others give you `(train_id, arr)` pairs.

Getting a KeyData object is now lazy-ish: `run[source, key]` opens exactly one file, to check the data type & shape. So `run[source, key].entry_shape` should be an efficient way to get the shape of one entry. The value of `run[source, key].shape[1:]` is the same, but it needs to open all sequence files for `.shape`.

This has some tests, but not yet docs. The array style methods are exercised by pre-existing tests, because `DataCollection` methods like `get_array()` and `get_dask_array()` are now calling the relevant `KeyData` methods.

philsmt:
Great, that will be really useful. `KeyData` now also makes a nice functor to map over.

While it's somewhat of a tangent, it may be useful to consider sooner rather than later. It came up often the other day and sometimes raised weird exceptions:  Which of these paths use only the `INDEX`, and which have look at the actual datasets in `CONTROL`/`INSTRUMENT`? The rationale for asking is that we encountered a lot of scenarios last Thursday where these two parts to not match, leading to an inconsistent run from EXtra-data's point of view. This should be fixed in the pclayer down the line, but for the moment it's there and happens. While `extra-data-validate` does report these problems, EXtra-data's API crashes rather ungracefully. I'd suggest to double-check each use here now, and then add some early warnings in an upcoming PR to please run `extra-data-validate`.

In a last attempt to push train-based slicing: Given that we want to remove `by_index` anyway, would you consider moving (or copying) `train_from_id()` and `select_trains()` by index to `KeyData.__getitem__`, i.e. `run[source, key][train]`? This would nicely separate these distinct components (unlike `run[source, key, train]`).

takluyver:
> Which of these paths use only the INDEX, and which have look at the actual datasets in CONTROL/INSTRUMENT? 

We inspect the actual datasets to get `entry_shape` and `dtype` when constructing a `KeyData` object. Other than that, only the methods which return data use the underlying datasets. Everything to do with the trains dimension trusts the index.

> Given that we want to remove `by_index` anyway, would you consider moving (or copying) `train_from_id()` and `select_trains()` by index to `KeyData.__getitem__`, i.e. `run[source, key][train]`?

I'd consider it, but which should it do? Return a narrower KeyData object like `select_train`, or read the data into an array like `train_from_*`? I'm inclined to the former, because it offers more flexibility, but it's less convenient if you do want a numpy array.

takluyver:
I followed @philsmt's suggestion to merge #68 and rebase this on top of it. :slightly_smiling_face:

philsmt:
> I'd consider it, but which should it do? Return a narrower KeyData object like `select_train`, or read the data into an array like `train_from_*`? I'm inclined to the former, because it offers more flexibility, but it's less convenient if you do want a numpy array.

Doh, excellent question for my not fully thought out suggestion. I see a couple of options:

1. Train data: This enables very fast access to data, i.e. `run[source, key][2]` or `run[source, key][-2:]` gives one or more trains. However, we'd need to elevate `ndarrays` (or any other chosen type) to be special and make its explicit method superfluous (identical to `run[source, key][:]`). It also breaks symmetry with `DataCollection` and how generic collection objects in Python behave.

2. Train selection: Stronger symmetry with `DataCollection`, with `KeyData` always returning a sliced `KeyData`. Makes data access for single trains a bit inconvenient, unless we the return values of `ndarray()` and possibly others are squeezed. Note that (as of last version), `DataCollection.select_trains()` actually does not support index access, i.e. a singular element.

3. Both: Make index access return the train data, while slices return a sliced KeyData. This would treat `KeyData` as a collection of train data and behave analogous to most collections like `list` or `ndarray`. Accessing a single train is very quick, accessing data for several trains requires a subsequent explicit call. But this solves the problem of choice for the container type in 1 and is less of a hassle than needing to index singular trains in 2. I'd assume that if you want data really quick, it's only for a single train.

takluyver:
> Both: Make index access return the train data, while slices return a sliced KeyData.

I wondered if you'd go for that. :wink: I'm not particularly keen on mixing return types like this. I think if you're used to `x[0]` returning an array, you'll expect `x[:10]` to do the same - possibly with an extra dimension, but not a totally different kind of object.

So I think if we do define `__getitem__()`, I'm inclined to make it like `.select_trains()`, even with a single index.

philsmt:
It was worth the try and my best argument for it. It all boils down whether `KeyData` should look like a `Sequence` similar to spanning several `h5py.Dataset`s, mapping trains to data, or some arbitrary higher-level object. But then we should probably also go for `__iter__` rather than a custom method name.

I agree than apart from this case, train selection makes the most sense. Given `ndarray.squeeze`, it's not terribly ugly to only get a single train's worth of data.

tmichela:
looking neat! :+1:

takluyver:
I'll add a few docstrings, and then merge if there are no further comments.

PR #72, opened 2020-06-30
Add unstack flag to get_array method

daviddoji:
Hopefully this will solve the reindexing ValueErrors when the pulseIds have all the same values for a particular run

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -332,7 +332,7 @@ class MPxDetectorBase:
           ID, by_index by index within the data being read. The default includes
           all pulses. Only used for per-train data.
         unstack_pulses: bool
-          Wheter to separate train and pulse dimensions.
+          Whether to separate train and pulse dimensions.
         """
         pulses = _check_pulse_selection(pulses)
 
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -15,7 +15,7 @@ log = logging.getLogger(__name__)
 MAX_PULSES = 2700
 
 
-def _guess_axes(data, train_pulse_ids, unstack):
+def _guess_axes(data, train_pulse_ids, unstack_pulses):
     # Raw files have a spurious extra dimension
     if data.ndim >= 2 and data.shape[1] == 1:
         data = data[:, 0]
@@ -35,7 +35,7 @@ def _guess_axes(data, train_pulse_ids, unstack):
 
     arr = xarray.DataArray(data, {'train_pulse': train_pulse_ids}, dims=dims)
 
-    if unstack:
+    if unstack_pulses:
         # Separate train & pulse dimensions, and arrange dimensions
         # so that the data is contiguous in memory.
         dim_order = ['train', 'pulse'] + dims[1:]
@@ -239,7 +239,7 @@ class MPxDetectorBase:
 
         return np.concatenate(positions)
 
-    def _get_module_pulse_data(self, source, key, pulses, unstack):
+    def _get_module_pulse_data(self, source, key, pulses, unstack_pulses):
         seq_arrays = []
         data_path = "/INSTRUMENT/{}/{}".format(source, key.replace('.', '/'))
         for f in self.data._source_index[source]:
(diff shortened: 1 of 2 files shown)


extra_data/components.py
```diff
@@ -328,6 +331,8 @@ def get_array(self, key, pulses=by_index[:]):
           Select the pulses to include from each train. by_id selects by pulse
           ID, by_index by index within the data being read. The default includes
           all pulses. Only used for per-train data.
+        unstack: bool
+          Wheter or not to separate train and pulse dimensions
```
takluyver, line 335:
```suggestion
          Whether to separate train and pulse dimensions.
```

takluyver:
I'm still trying to think if there's a better name for the public parameter. `unstack` is the Xarray method we call to achieve it, but I'm not sure the meaning is obvious from the outside.

daviddoji:
> I'm still trying to think if there's a better name for the public parameter. `unstack` is the Xarray method we call to achieve it, but I'm not sure the meaning is obvious from the outside.

Either that or "separate_dims", which is basically what it basically does.

takluyver:
A bit vague, though. `separate_pulse_dim` or `pulse_dim` (it creates a separate pulse dimension)? `align_trains` or `align_pulses` (it lines up the same pulse across different trains)? :-S

daviddoji:
I would prefer "align_pulses"

takluyver:
OK, I'm happy enough with that. We should add a test of passing `False` - see the other tests in `test_components.py`.

extra_data/components.py
```diff
@@ -608,7 +614,7 @@ def _get_pulse_data(self, source, key, tid):
             # h5py fancy indexing needs a list, not an ndarray
             data_positions = list(first + positions)
 
-        return _guess_axes(ds[data_positions], train_pulse_ids)
+        return _guess_axes(ds[data_positions], train_pulse_ids, unstack=unstack)
```
takluyver, line 617: unstack isn't defined here

tmichela:
`pulse_dim` or `add_pulse_dim` seems clearer to me as it adds a new dimension to the array with the pulse resolved data?

takluyver:
Unless you think of the existing first dimension as pulses, and this is adding a trains dimension.

Naming things is hard!

tmichela:
good point! so `add_train_dim` ? :D

daviddoji:
> good point! so `add_train_dim` ? :D

I'm happy with both.

extra_data/tests/test_components.py
```diff
@@ -104,6 +104,14 @@ def test_get_array_pulse_indexes_reduced_data(mock_reduced_spb_proc_run):
     assert np.isin(arr.coords['pulse'], [1, 7, 15, 23]).all()
 
 
+def test_get_array_as_xarray(mock_fxe_raw_run):
```
takluyver, line 107: Naming again - it returns an xarray object either way, so it's not quite right to call this test `_as_xarray`.

extra_data/components.py
```diff
@@ -328,6 +331,8 @@ def get_array(self, key, pulses=by_index[:]):
           Select the pulses to include from each train. by_id selects by pulse
           ID, by_index by index within the data being read. The default includes
           all pulses. Only used for per-train data.
+        unstack_pulses: bool
+          Wheter to separate train and pulse dimensions.
```
takluyver, line 335:
```suggestion
          Whether to separate train and pulse dimensions.
```

extra_data/tests/test_components.py
```diff
@@ -18,6 +18,9 @@ def test_get_array(mock_fxe_raw_run):
     assert arr.shape == (16, 3, 128, 256, 256)
     assert arr.dims == ('module', 'train', 'pulse', 'slow_scan', 'fast_scan')
 
+    arr = det.get_array('image.data', pulses=by_index[0], unstack_pulses=False)
+    assert arr.shape == (16, 384, 256, 256)
```
takluyver, line 22: Check the expected dimension names here as well?

takluyver:
LGTM

PR #73, opened 2020-07-01
make-virtual-cxi fill-value accepts hexadecimal/octal/binary formatting

tmichela:


extra_data/cli/make_virtual_cxi.py
```diff
@@ -54,6 +66,7 @@ def main(argv=None):
     )
     ap.add_argument(
         '--fill-value', action='append', nargs=2, metavar=('DS', 'V'),
+        type=parse_number,
```
takluyver, line 69: Doesn't this mean it will also try to parse the dataset name as an integer? I know that's unlikely to come up, but it seems like bad design. Maybe do the integer parsing separately after `parse_args`.

It also seems odd that it skips parsing base10 integers.

extra_data/cli/make_virtual_cxi.py
```diff
@@ -27,6 +27,18 @@ def _detectors():
...
 
 
+def parse_number(arg):
+    arg = arg.strip()
+    if arg.startswith('0x'):
+        return int(arg, 16)
+    elif arg.startswith('0o'):
+        return int(arg, 8)
+    elif arg.startswith('0b'):
+        return int(arg, 2)
```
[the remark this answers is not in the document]
takluyver: Neat, I didn't know about that.

extra_data/cli/make_virtual_cxi.py
tmichela: yea, could be `float` or `int`, I thought it'd be ok to let it be cast later (the current behavior)... I'll change that

extra_data/cli/make_virtual_cxi.py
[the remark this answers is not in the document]
tmichela: :+1:

extra_data/cli/make_virtual_cxi.py
```diff
@@ -27,6 +27,13 @@ def _detectors():
     return [d.__name__ for d in MPxDetectorBase.__subclasses__()]
 
 
+def cast(number:str):
```
takluyver, line 30: I liked `parse_number` better as a name for this - it's explicit about what it does.

takluyver:
Implementation LGTM, I'm just quibbling about the name. :slightly_smiling_face:

dallanto:
LGTM to me as well (I notice that there are sime CI checks failing, though: ``pluggy.manager.PluginValidationError: Plugin 'pytest_cov' could not be loaded``)

tmichela:
Thanks!

@dallanto that's a pytest version issue, see #74

PR #74, opened 2020-07-01
Upgrade pytest on Travis CI

takluyver:
Tests are failing because a newer version of pytest-cov gets installed and it requires a newer pytest.


takluyver:
Merging quickly so tests work again for other PRs.

tmichela:
:+1:

PR #75, opened 2020-07-01
Fix zero-length arrays when data is missing in TrainIterator

philsmt:
When data is missing for a particular train and `require_all=False` (the default), `TrainIterator._assemble_data` slices an array of zero length out of the dataset. The resulting data shape is odd and ugly, e.g. (0, 1024, 1024) for a pnCCD image. This tiny change will cause the corresponding key to be absent in the data dictionary instead.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1257,7 +1257,7 @@ class TrainIterator:
                 first, count = firsts[pos], counts[pos]
                 if count == 1:
                     source_data[key] = ds[first]
-                else:
+                elif count > 0:
                     source_data[key] = ds[first : first + count]
 
         return res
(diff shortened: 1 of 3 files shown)


takluyver:
The mock FXE run includes a camera with no data which we could use to test this, e.g. in this test: https://github.com/European-XFEL/EXtra-data/blob/ddb2b4c9a9232d83ab21d22ff4645fd93e96fafa/extra_data/tests/test_reader_mockdata.py#L161

philsmt:
Indeed, worked like a charm. Would you like to remove the XGM from that test, given that we test the camera _with_ data as well, to maintain symmetry of with and without data.

takluyver:
Could do, but I think it's OK as is - it's not much code or time to check a second source.

PR #76, opened 2020-07-03
Fix for failing detector data test

takluyver:
I'm not quite sure why this seemed to be passing before, but it's failing both on Travis and for me locally now.


daviddoji:
LGTM

PR #77, opened 2020-07-10
Add require_all parameter for det.trains()

### LINKED ISSUE #71: Detector.trains() method silently skips trains with missing data
I think this was a choice made for data where detector modules would sometimes randomly miss a train, and it's convenient to only process complete frames. But @daviddoji ran into a case where one module was entirely missing in the selected data, and it wasn't obvious that this was why `lpd.trains()` gave no data at all. We should at least add an option to disable this.

Now that we've got the `min_modules` parameter, it would arguably make sense to change the default to use that. But that would break compatibility for any code that was already using that `detector.trains()` interface.

takluyver:
It turns out the machinery was already there for this parameter, it just wasn't being used.

@daviddoji can you check if this works for your use case?

Fixes #71 - at least as far as we can without breaking the API for existing code.

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -407,6 +407,9 @@ class MPxDetectorBase:
         pulses: slice, array, by_index or by_id
           Select which pulses to include for each train.
           The default is to include all pulses.
+        require_all: bool
+          If True (default), skip trains where any of the selected detector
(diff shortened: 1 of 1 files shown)


tmichela:
LGTM

Issue #80, opened 2020-07-28
Issue with writing subsets of AGIPD data to new file


reiserm:
I tried to use the `AGIPD1M.write_frames()` method to write a subset of processed data to a new HDF5 file

`agp = AGIPD1M(run, min_modules=16)`
`agp.write_frames('test.h5', np.array(run.train_ids[10:20]), np.ones(10))`

This causes the following h5py error:

`ValueError: Chunk shape must not be greater than data shape in any dimension. (4096,) is not compatible with (0,)`

takluyver:
Thanks Mario. I guess one of the modules has no data in that selection, so it's trying to write an empty dataset. I'll get on this.

takluyver:
Can I ask you to post the traceback or the path of a run so I can reproduce it? I tried to write a test, but so far it's passing.

reiserm:
In [4]: run = RunDirectory("/gpfs/exfel/exp/MID/202022/p002693/proc/r0087/")

In [5]: agp = AGIPD1M(run, min_modules=16)

In [6]: agp.write_frames('test.h5', np.array(run.train_ids[10:20]), np.ones(10))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-6-955159130720> in <module>
----> 1 agp.write_frames('test.h5', np.array(run.train_ids[10:20]), np.ones(10))

/gpfs/exfel/u/scratch/MID/202022/p002693/reiserm/.venv/lib64/python3.6/site-packages/extra_data/components.py in write_frames(self, filename, trains, pulses)
    444         writer = FramesFileWriter(filename, self.data, inc_tp_ids)
    445         try:
--> 446             writer.write()
    447         finally:
    448             writer.file.close()

/gpfs/exfel/u/scratch/MID/202022/p002693/reiserm/.venv/lib64/python3.6/site-packages/extra_data/writer.py in write(self)
    133
    134         for source in d.all_sources:
--> 135             self.prepare_source(source)
    136
    137         self.write_metadata()

/gpfs/exfel/u/scratch/MID/202022/p002693/reiserm/.venv/lib64/python3.6/site-packages/extra_data/writer.py in prepare_source(self, source)
     27             src_ds1 = self.data._source_index[source][0].file[path]
     28             self.file.create_dataset_like(
---> 29                 path, src_ds1, shape=(nentries,) + src_ds1.shape[1:],
     30             )
     31             if source in self.data.instrument_sources:

[...]

PR #82, opened 2020-07-29
Fix writing frames from corrected data

takluyver:
Whereas the raw data has an unlimited first dimension, the proc data has the maximum shape the same as the actual data shape. This causes a problem when we try to create the image datasets with 0 length originally, because the length is less than the chunk size. The easiest way to avoid this is to always set maxshape to offer an unlimited first dimension.

Closes gh-80.

A couple of dependency changes seem to have broken our tests. I've restricted versions of pytest and pandas for now; I'll work on fixing those issues separately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
 
 install:
   - make install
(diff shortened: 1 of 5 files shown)


tmichela:
LGTM

PR #86, opened 2020-08-05
Create dependabot.yml

RobertRosca:
Adds dependabot daily CI

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "pip" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"


takluyver:
Would this actually do anything? I've generally seen dependabot and similar tools used on projects that follow the application pattern of pinning dependency versions. EXtra-data follows the library pattern of allowing a broad range of versions, generally only declaring minimum versions. We don't currently specify a maximum version of any dependency:

https://github.com/European-XFEL/EXtra-data/blob/7507251aa477928d4afe071c155d73b89c0ce710/setup.py#L47-L72

RobertRosca:
Ah I ran this through the web interface and accidentally made the PR a bit early. Just made this to remind myself to do it later.

You're right, the packages need to be at least slightly pinned for dependabot to work.

The main benefit of using dependabot is that it automatically creates a PR whenever a dependency is updated to bump the version number, which then also means that the CI runs on the bumped versions. This way if a new version of some dependency we use comes out and breaks functionality the CI would catch it. Very useful especially as most of us already have extra data installed so we wouldn't typically notice issues caused by dependency changes.

If you use a combination of poetry files as well as a 'traditional' `setup.py` file then dependabot (and anybody who already uses poetry) can use the `pyproject.toml` and `pyproject.lock` to track dependencies/bump versions, but you can also still install it via pip as normal, so most users won't notice anything change.

RobertRosca:
Closed by #106

PR #87, opened 2020-08-06
Add names to xarray results

philsmt:
The `xarray.DataArray`s returned by `KeyData.xarray()` or `DataCollection.get_array()` don't have an explicit name set, causing functions like `xarray.merge` to fail when converting to a `xarray.Dataset`. This PR uses `{source}.{key}` as the array name by default with an optional argument added to specify a custom name.

### DIFF
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -162,6 +162,9 @@ class KeyData:
         if name is None:
             name = f'{self.source}.{self.key}'
 
+            if name.endswith('.value') and self.section == 'CONTROL':
+                name = name[:-6]
+
         return xarray.DataArray(ndarr, dims=dims, coords=coords, name=name)
 
     def series(self):
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -133,17 +133,17 @@ class KeyData:
         ]
         return np.concatenate(chunks_trainids)
 
-    def xarray(self, extra_dims=None, name=None, roi=()):
+    def xarray(self, extra_dims=None, roi=(), name=None):
         """Load this data as a labelled xarray.DataArray.
 
         The first dimension is labelled with train IDs. Other dimensions may be
         named by passing a list of names to *extra_dims*.
 
-        The array's name may be given by *name*, by default the source and key
-        is joined by a dot.
-
         *roi* may be a ``numpy.s_[]`` expression to load e.g. only part of each
         image from a camera.
+
+        The array's name may be given by *name*, by default the source and key
+        is joined by a dot.
         """
         import xarray
 
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -175,7 +175,7 @@ class KeyData:
             raise TypeError("pandas Series are only available for 1D data")
 
         name = self.source + '/' + self.key
-        if name.endswith('.value'):
+        if name.endswith('.value') and self.section == 'CONTROL':
             name = name[:-6]
 
         index = pd.Index(self._trainid_index(), name='trainId')
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -133,11 +133,14 @@ class KeyData:
         ]
         return np.concatenate(chunks_trainids)
 
-    def xarray(self, extra_dims=None, roi=()):
+    def xarray(self, extra_dims=None, name=None, roi=()):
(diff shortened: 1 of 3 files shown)


tmichela:
Can you move the new argument to the last position in each methods? To prevent potentially breaking existing code if the argument order changes.

Else than that, LGTM

philsmt:
Doh, excellent point! I wondered a moment whether the order could matter, but missed on that.

extra_data/keydata.py
```diff
@@ -156,7 +159,10 @@ def xarray(self, extra_dims=None, roi=()):
         if self.shape[0]:
             coords = {'trainId': self._trainid_index()}
 
-        return xarray.DataArray(ndarr, dims=dims, coords=coords)
+        if name is None:
+            name = f'{self.source}.{self.key}'
```
takluyver, line 163: We already make a name for pandas Series objects, with a `/` rather than a `.`. We also remove a trailing `.value` (all control keys have `.timestamp` and `.value` in the file - we should probably limit it to control sources):

https://github.com/European-XFEL/EXtra-data/blob/master/extra_data/keydata.py#L168-L170

Not essential, but it would be good to be consistent between series and xarray objects if it's practical to do so.

takluyver:
On the parameter order, we could also start making keyword-only arguments (e.g. `..., *, name=None)`), so that we don't need to worry about the order when adding another parameter later. I think it's rarely clear to pass more than ~2 args positionally. But I don't have a strong preference on this.

extra_data/keydata.py
philsmt: Coming from live Karabo, I'm all for removing the trailing `.value` given it's a file-only peculiarity. I could see two issues here:

- Somebody might want to merge two arrays with the values and timestamp into a Dataset, which would end up with the same name
- Given file's always show this `.value` and `.timestamp`, a user without knowledge of live Karabo may be confused by it missing

I'm not so keen on breaking from it by joining with a slash. But if the decision has already been made for pandas, I suppose we should stick with it?

takluyver: > Somebody might want to merge two arrays with the values and timestamp into a Dataset, which would end up with the same name

Wouldn't one be `...foo` and the other `...foo.timestamp`?

> Given file's always show this .value and .timestamp, a user without knowledge of live Karabo may be confused by it missing

True enough if they dig into the files. There's definitely a trade-off between being more like the Karabo name and more like the name in the file.

> I'm not so keen on breaking from it by joining with a slash. 

I'm not sure what you mean by breaking it - does Karabo or some other context show them joined with a dot? I was unaware of anything showing source + key in a single string, so I picked a slash basically arbitrarily. I think it should mean that splitting on the last `/` separates source & key, whereas possibly source names could contain `.`?

philsmt: > Wouldn't one be `...foo` and the other `...foo.timestamp`?

We shall not speak of it again. But since we're already here, may I ask for automatically attaching `.value` to control sources, if neither `.value` or `.timestamp` is passed as `key`? Tripped over that way too often.

> I'm not sure what you mean by breaking it - does Karabo or some other context show them joined with a dot? I was unaware of anything showing source + key in a single string, so I picked a slash basically arbitrarily. I think it should mean that splitting on the last `/` separates source & key, whereas possibly source names could contain `.`?

All throughout karabo-gui, e.g. any widget created from a property (tooltip shown):

kgui-device-property

It get's a bit tricky for fast data, as unlike the framework, the karabo-gui always shows a `.` instead of a `:`. I'd wager we might convince CTRL to fix that though.

takluyver: > But since we're already here, may I ask for automatically attaching .value to control sources, if neither .value or .timestamp is passed as key?

Yup, I'll open a new issue for it.

> All throughout karabo-gui, e.g. any widget created from a property

My bad. I've never actually used Karabo much. In that case, I'd say stick with `.` - it's probably more valuable to be consistent with displayed names in Karabo.

philsmt: I've added the removal of a trailing `.value` to `.xarray()` and sneaked a fix in for `.series()`, which technically could have applied it to fast data, too. While it's somewhat unlikely, the key in a pipeline hash _may_ end with `.value`.

takluyver:
Thanks, LGTM.

PR #90, opened 2020-08-17
Renaming karabo-data-valitate -> extra-data-validate

daviddoji:
By mistake, I found this small bug in the help message when one does not pass a path to ```extra-data-validate```

image


takluyver:
Thanks David!

PR #92, opened 2020-08-25
KeyData: get data counts

### LINKED ISSUE #91: train_ids in KeyData
the `train_ids` attribute of the `KeyData` object is a simple copy of the train ids contained in the `DataCollection`. That leads to some inconsistencies. For example calling `.shape` will return a shape where the trainId axis match the actual count of data available in the files for the source, but calling `.train_ids` will contains train ids with and without data for this key.

One way to go around this if for example using xarray: `data.xarray().trainIds` will return only the train ids containing data, but that'd be ideal if it wasn't necessary to load the data to find which trainIds have or do not have data.

Would it be sensible to have the `train_ids` filtered (either at init or when the property is called) and only contains the train ids that actually have data for the key?

tmichela:
Implements `get_data_counts` in `KeyData`: Get a count of data points in each train.

closes #91

### DIFF
--- a/extra_data/tests/make_examples.py
+++ b/extra_data/tests/make_examples.py
@@ -269,8 +269,7 @@ def make_reduced_spb_run(dir_path, raw=True, rng=None, format_version='0.5'):
             AGIPDModule('SPB_DET_AGIPD1M-1/DET/{}CH0'.format(modno), raw=raw,
                          frames_per_train=frame_counts)
             ], ntrains=64, chunksize=32, format_version=format_version)
-    if not raw:
-        return
+
     write_file(osp.join(dir_path, '{}-R0238-DA01-S00000.h5'.format(prefix)),
                [ XGM('SA1_XTD2_XGM/DOOCS/MAIN'),
                  XGM('SPB_XTD9_XGM/DOOCS/MAIN'),
--- a/extra_data/tests/test_keydata.py
+++ b/extra_data/tests/test_keydata.py
@@ -86,7 +86,7 @@ def test_get_train(mock_spb_raw_run):
         xgm_beam_x.train_from_index(9999)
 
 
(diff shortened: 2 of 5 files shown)


extra_data/tests/make_examples.py
```diff
@@ -269,8 +269,7 @@ def make_reduced_spb_run(dir_path, raw=True, rng=None, format_version='0.5'):
             AGIPDModule('SPB_DET_AGIPD1M-1/DET/{}CH0'.format(modno), raw=raw,
                          frames_per_train=frame_counts)
             ], ntrains=64, chunksize=32, format_version=format_version)
-    if not raw:
-        return
```
tmichela, line 273: That's not directly related to the PR,
The calibration service now copies all non corrected files as well

takluyver:
Thanks! I'll look at the code tomorrow.

One quick naming thought: I've been generally dropping the `get_` prefix for KeyData, e.g. `.series()` instead of `.get_series()`. We could follow that pattern and call the method `data_counts()`. I don't feel strongly about this, though - I think most people are just going to find individual methods that are useful, so the pattern don't matter too much.

tmichela:
changed it, I also like `data_counts` better

extra_data/tests/test_keydata.py
```diff
@@ -84,3 +84,23 @@ def test_get_train(mock_spb_raw_run):
 
     with pytest.raises(IndexError):
         xgm_beam_x.train_from_index(9999)
+
+
+def test_get_data_count(mock_reduced_spb_proc_run):
```
takluyver, line 89:
```suggestion
def test_data_counts(mock_reduced_spb_proc_run):
```

takluyver:
Other than that little suggestion, LGTM.

tmichela:
Thanks!

PR #93, opened 2020-09-23
FIX: validate run when a file cannot be opened

tmichela:
Overlooked that the path to the file was not defined when an error is raised while opening a file in the validation code.

@daviddoji


daviddoji:
LGTM

tmichela:
Thanks :)

PR #94, opened 2020-10-01
Check that timestamps on control data is always increasing

tmichela:
I've got a request from IT to check that all timestamps are always increasing (or not changing if the value isn't updated).

This PR checks all timestamps  for all `CONTROL` datasets and check if a timestamp is decreasing, or a timestamp has a zero value before the last train.

There is a run example with a case of decreasing timestamp: /gpfs/exfel/exp/SPB/202031/p900145/raw/r0952/RAW-R0952-DA01-S00000.h5 (data is on the online cluster).

~~I'd like to ask if someone has a good strategy for 'additional' checks like this one? Or should it be part of the overall checks?~~
Any view @takluyver ? 

I need to benchmark this to see the impact of reading all control datasets.
edit: Impact is negligeable on the few tests I ran, so just running it by default.

### DIFF
--- a/extra_data/tests/test_validation.py
+++ b/extra_data/tests/test_validation.py
@@ -2,6 +2,7 @@ import os.path as osp
 from pathlib import Path
 
 from h5py import File
+import numpy as np
 from pytest import fixture, raises
 from tempfile import TemporaryDirectory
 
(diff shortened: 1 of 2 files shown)


takluyver:
I'm not sure what's happened with the commits here; Github seems to be picking up many more commits than are really involved.

takluyver:
Thanks, this looks OK.

I assume you've run it on a few runs and found that it generally passes? Out of order timestamps shouldn't actually give us any problems reading the files, so if this was a common issue I might be inclined to hide it behind a flag. But so long as it's fairly rare, I think doing it by default makes sense.

tmichela:
Yes, it should not be a common issue. It happened for a particular experiment. I haven't hit that particuar case in any recent run I've tried it with

takluyver:
In that case, LGTM, thanks!

tmichela:
Thanks!

PR #96, opened 2020-10-08
DOC: KeyData object

tmichela:
Documents the new `KeyData` object.

### DIFF
--- a/docs/reading_files.rst
+++ b/docs/reading_files.rst
@@ -71,31 +71,33 @@ to refer to all data associated with that 0.1 second window.
 Getting data by source & key
 ----------------------------
 
-Where data will fit into memory, it's usually quickest and most convenient
-to load it like this.
+.. note::
+
+   When applicable, we recommand getting data using :class:`extra_data.keydata.KeyData`
+   objects.
+
+   One can select a single data source as a :class:`extra_data.keydata.KeyData` object.
+   It offers the same functionality you can find in the :class:`DataCollection` object.
+   To select a data source from a run::
+
+      xgm_intensity = run['SA1_XTD2_XGM/XGM/DOOCS:output', 'data.intensityTD']
+
+   :ref:`See more about KeyData<keydata>`.
+
+We can also use methods directly on a :class:`DataCollection` object to load data. This
+can be useful if you want to iterate over multiple data source at a time, or get a
+:ref:`Pandas.DataFrame<getdf>` or a :ref:`virtual dataset<getvds>` from run data.
 
 .. class:: DataCollection
    :noindex:
 
    .. automethod:: get_array
 
-     .. seealso::
-       `xarray documentation <http://xarray.pydata.org/en/stable/indexing.html>`__
-         How to use the arrays returned by :meth:`~.get_array`
-
-       :doc:`xpd_examples`
-         Examples using xarray & pandas with EuXFEL data
-
    .. automethod:: get_dask_array
 
-     .. seealso::
-       `Dask Array documentation <https://docs.dask.org/en/latest/array.html>`__
-         How to use the objects returned by :meth:`~.get_dask_array`
-
-       :doc:`dask_averaging`
-         An example using Dask with EuXFEL data
-
    .. automethod:: get_series
+
+   .. _getdf:
 
    .. automethod:: get_dataframe
 
@@ -106,6 +108,8 @@ to load it like this.
 
         :doc:`xpd_examples`
           Examples using xarray & pandas with EuXFEL data
+
+   .. _getvds:
 
    .. automethod:: get_virtual_dataset
 
@@ -118,6 +122,13 @@ to load it like this.
 
 Getting data by train
 ---------------------
+
+.. note::
+
+   When applicable, we recommand getting data using :class:`extra_data.keydata.KeyData`
+   objects.
+
+   :ref:`See more about KeyData<keydata>`.
 
 Some kinds of data, e.g. from AGIPD, are too big to load a whole run into
 memory at once. In these cases, it's convenient to load one train at a time.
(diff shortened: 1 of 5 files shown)


takluyver:
I think I'd be inclined to put this under *Getting data by source & key*, as that's primarily what it fits with. Then we could probably make the documentation of the matching `DataCollection.get_*` methods much shorter, like:

> Get a pandas Series for a given data field. See `KeyData.series` for details.

At some point, now that this has been released for a while, I'd also like to update the examples to use the new API. But that doesn't need to be part of this PR.

docs/reading_files.rst
```diff
@@ -156,6 +156,18 @@ data, so you use them like this::
...
    .. automethod:: union
 
+Selecting a single data source
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+One can select a single data source as a ``KeyData`` object. It offers the same
+functionality you can find in the ``DataCollection`` object. To select a data source
+from a run::
+
+  keydata = run[source, key]
```
takluyver, line 166:
```suggestion
  xgm_intensity = run['SA1_XTD2_XGM/XGM/DOOCS:output', 'data.intensityTD']
```

docs/reading_files.rst
```diff
@@ -156,6 +156,18 @@ data, so you use them like this::
...
+
+.. autoclass:: extra_data.keydata.KeyData
+   :members:
```
takluyver, line 169: I have a slight preference for listing the methods manually, because then we can put them in a meaningful order. E.g. at the moment, the first method is `dask_array`, which is probably not a good starting point for understanding the API. And it would be good to group `ndarray` and `xarray` together, as they're conceptually very similar.

takluyver:
The docs build has a warning because there's a reference to the `select_trains` method which is now ambiguous:

`.../EXtra-data/docs/reading_files.rst:233: WARNING: more than one target found for cross-reference 'select_trains': extra_data.DataCollection.select_trains, extra_data.keydata.KeyData.select_trains`

docs/reading_files.rst
tmichela: Good point, I'll update that

tmichela:
> I think I'd be inclined to put this under _Getting data by source & key_, as that's primarily what it fits with.

I originally wanted to put it there, but my reasoning to put it in *Selecting & combining data* is that it effectively returns an object similar to a `DataCollection`, only containing a single source .

the methods listed under _Getting data by source & keys_ either *read* the data in files and return it (ndarray, xarray, serie, dataframe) or return a view-like kind of object (dask array, virtual dataset) on the data, which isn't what getting a KeyData object do.

Does that make sens?

> Then we could probably make the documentation of the matching DataCollection.get_* methods much shorter ...

Er... so currently the methods on `DataCollection` are better described than their equivalent in `KeyData`. Do you reckon to move the more detailed definition to the KeyData objects? (currently `DataCollection` is exposed to the API, but `KeyData` isn't, that would be an argument in favor to leave the more verbose docstring there?)

takluyver:
> the methods listed under Getting data by source & keys either read the data in files and return it (ndarray, xarray, serie, dataframe) or return a view-like kind of object (dask array, virtual dataset) on the data, which isn't what getting a KeyData object do.

Getting a KeyData object doesn't, but I'm imagining that you'll often get it and immediately call one of its methods that do, like `run['SA1_XTD2_XGM/XGM/DOOCS:output', 'data.intensityTD'].xarray()` .

> currently the methods on DataCollection are better described than their equivalent in KeyData. Do you reckon to move the more detailed definition to the KeyData objects?

I'm roughly thinking that we'll gradually make the KeyData methods the primary way to do those things, and de-emphasise the DataCollection methods that achieve the same thing. I think it makes sense to advertise just one way to do these things - there's enough to understand already without also working out that this and that are the same, and whether one is better than the other.

So if we think the `run[source, key].xarray()` style is better than `run.get_array(source, key)` - which I do, though I'm not fixed on that - then yes, I would transfer the detail to KeyData and have shorter docstrings for the matching DataCollection methods. This is also part of why I'm thinking to put it under 'getting data by source and key' - if someone starts reading from that heading, we can show them the new recommended way to do those things.

tmichela:
hum... So I tried to go in that direction but it is not easy with the current structure and it felt really awkward to me. Especially because `DataCollection` has extra methods (get_dataframe, get_virtual_datasets) and some methods have extra attributes (e.g. trains / train_from...). I don't really see a good way of showing everything with taking the `KeyData` object as reference...

What about this: keep the `KeyData` in `Selecting & combining data` for now, and adding notes at the beginning of the previous chapters advising to use `KeyData` when applicable? (that's what I've done in the last commit)

We can think of refactoring more thoroughly the doc if you think that's necessary at a later time. I'd like to have at least a reference to the KeyData object to advertise it in the next newsletter (should be released this week).

(I also realized that the implementation of `get_serie` is different in the datacollection and keydata objects.)

takluyver:
Sorry for not getting back to this sooner. I've just pushed some changes which put KeyData front and centre, while focusing on how it's used rather than the name of it. I'm wondering whether to break it up across the 'Getting data by train' and 'Selecting & combining data' sections, or whether that just makes it more confusing.

> I also realized that the implementation of get_serie is different in the datacollection and keydata objects.

Yes - `run.get_series()` checks if there is >1 entry for any train, and tries to additionally load the pulse IDs for AGIPD/LPD/DSSC data and make a multi-level index. But by the time we made KeyData, the idea was that you would use the `extra_data.components` module for those big detectors, so `key.series()` just leaves it with train ID labels, even if they're not unique.

I suspect that no-one cares about `run.get_series()` labelling pulse IDs - it only works for 1D data, so there are very few cases it could apply to. But I was hesitant to take it out just in case someone was relying on it. Maybe I'll deprecate it.

takluyver:
(I rebased the branch to stop it testing on Travis and get RTD building the PR, so be careful when pulling it)

tmichela:
:+1: Thanks going through it, I agree it is more meaningful arranged like it is now.

Regarding the question of organizing the doc per class or usage, I would say that the current organisation is more helpful for new users as they can go through the index and easily find which chapter describe what one would want to do.

takluyver:
Thanks @tmichela :-)

PR #98, opened 2020-10-29
FIX: Components Dask array dtype 

tmichela:
While playing with dask, I realized that the dtype of the array returned by `get_dask_array` changed the data type to `float64`.

This ensure the the datatype is kept as in the source arrays by setting the fill value to `0` for arrays of dtype != `float`

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -324,13 +324,16 @@ class MPxDetectorBase:
     def _fill_value(value, dtype):
         if value is None:
             if dtype.kind != 'f':
-                return 0
+                value = 0
             else:
-                return np.nan
+                value = np.nan
+
+        # enforce that fill value is compatible with array dtype
+        value = dtype.type(value)
         return value
 
     def get_array(self, key, pulses=np.s_[:], unstack_pulses=True,
-                  fill_value=None, join='outer'):
+                  fill_value=None):
         """Get a labelled array of detector data
 
         Parameters
@@ -347,17 +350,6 @@ class MPxDetectorBase:
         fill_value: int or float, optional
             Value to use for missing values. If None (default) the fill value
             is 0 for integers and np.nan for floats.
-        join: str, optional
-            String indicating how to combine detector modules in the new array:
-            - “outer”: use the union of object indexes
-            - “inner”: use the intersection of object indexes
-            - “left”: use indexes from the first object with each dimension
-            - “right”: use indexes from the last object with each dimension
-            - “exact”: instead of aligning, raise ValueError when indexes to be
-                aligned are not equal
-            - “override”: if indexes are of same size, rewrite indexes to be
-                those of the first object with that dimension. Indexes for the
-                same dimension must have the same size in all objects.
         """
         pulses = _check_pulse_selection(pulses)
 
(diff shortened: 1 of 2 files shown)


tmichela:
as a side effect (could be pure luck) after changing this, I managed to run the dask example using an dask array of the whole detector from `AGIPD1M` for the first time:

https://max-jhub.desy.de/user-redirect/notebooks/GPFS/exfel/data/scratch/tmichela/notebooks/dask_averaging_component.ipynb

takluyver:
I think this makes sense by default, but should we have an option? E.g. if you take a mean, filling missing data with 0 will make the result wrong. Or maybe we should default to doing an inner join (i.e. only keeping frames where all modules have data)?

If the raw data is uint16, then converting it to float64 makes it 4 times larger in memory. So, based on my guess that the crashes are due to memory usage, it wasn't luck that it worked after this change, but it's also not problem solved.

tmichela:
Good points. What would be your preferred solution:
- default to inner join
- default to outer join (with casting to float)
- default to outer join with default fill value to 0 for ints
- something else?
and exposing the fill_value and join option

tmichela:
> ... based on my guess that the crashes are due to memory usage ...

That's also my feeling after playing a bit.
I would guess that the job of splitting the work in smaller tasks does not know about the global memory available or even the memory available to each workers...

takluyver:
I think I'd go for outer join by default, as that's what the API already does, with 0 for missing integers. But let's make the downside of that clear in the docstring, and add an option to use a different fill value.

Do we need to do the same in `.get_array()`?

extra_data/components.py
```diff
@@ -396,7 +396,9 @@ def get_dask_array(self, key, subtrain_index='pulseId'):
 
             arrays.append(mod_arr)
 
-        return xarray.concat(arrays, pd.Index(modnos, name='module'))
+        # set the fill_value to prevent xarray from changing the dtype to float
+        fill_value = 0 if arrays[0].dtype.kind != 'f' else xarray.core.dtypes.NA
```
takluyver, line 400: Looking at the code in xarray, I think the `NA` sentinel always gets turned into `np.nan` for floats, if you want a shorter way to write it.

There are also other dtypes - complex, object, datetime, timedelta - which xarray is prepared to handle, though I don't think our data would ever be in those.

extra_data/components.py
```diff
@@ -334,6 +344,20 @@ def get_array(self, key, pulses=np.s_[:], unstack_pulses=True):
...
+            String indicating how to combine detector modules in the new array:
+            - “outer”: use the union of object indexes
+            - “inner”: use the intersection of object indexes
+            - “left”: use indexes from the first object with each dimension
+            - “right”: use indexes from the last object with each dimension
+            - “exact”: instead of aligning, raise ValueError when indexes to be
+                aligned are not equal
+            - “override”: if indexes are of same size, rewrite indexes to be
+                those of the first object with that dimension. Indexes for the
+                same dimension must have the same size in all objects.
```
takluyver, line 360: I don't think this description from Xarray is very helpful for someone using EXtra-data.

We can already control inner vs outer join by setting `min_modules` when creating the detector object - `min_modules=16` gives us the same effect as an inner join, `min_modules=1` (default) gives us an outer join. I'm not sure anyone would ever use the other options in this context, so maybe we don't need this parameter after all.

Issue #100, opened 2020-11-08
selection of cellId in the LPD1M object


camilabc23:
The LPD1M object is giving us train ids and pulse ids but it would be essential that we also get the cell Ids, since all the analysis fails when using veto patterns on the LDP otherwise, since the memory cell allocation is not equivalent to the pulse ids.

takluyver:
Would it be useful to have an option to get the data back with cell IDs in the index instead of pulse IDs? We can do that quite easily.

You can already get the cell IDs with `lpd.get_array('image.cellId')`, though this will give you cell IDs for each module (i.e. 16 IDs per frame), whereas the index logic checks that they're all the same and gives you a single cell ID per frame.

takluyver:
PR #101 would add an option to get data with cell IDs in the index instead of pulse IDs. Please take a look and let me know if that would be useful to you. :)

camilabc23:
Thank you. Getting the cell IDs in addition to the pulse IDs will be very helpful

takluyver:
This would be more 'instead' rather than 'in addition'. It would let you choose whether the data is labelled with pulse IDs or cell IDs, but there still wouldn't be an easy way to get both. Is that still useful?

camilabc23:
As long as it is a choice but we can change it back and forth that is ok. The issue is that there are certain thing she need the cell ID as the identifier and some other parameters were we will need to compare pulse IDs.

takluyver:
Thanks!

PR #101, opened 2020-11-09
Allow retrieving big detector data with cell IDs in the index instead of pulse IDs

takluyver:
This would be used like this:

```python
lpd = LPD1M(run)
lpd.get_array('image.data', subtrain_index='cellId')
```

Might be relevant for #100


takluyver:
This crucially assumes that different modules use the same cell ID for the same frame, because it will align the data by cell ID. If different modules can be on different cells at the same time, this should not be used!

daviddoji (commented):
LGTM.
Sorry for the late review!

PR #104, opened 2020-11-13
FIX: set default fill value to 0 for ints in stack_detector_data

### LINKED ISSUE #103: stack_detector_data has wrong default fill value for uint64 data
We've assumed that `np.nan` is universally converted to 0 for integer dtypes, but that doesn't hold for uint64:

```
In [4]: np.full(1, np.nan, dtype=np.uint32)                            
Out[4]: array([0], dtype=uint32)

In [5]: np.full(1, np.nan, dtype=np.uint64)                            
Out[5]: array([9223372036854775808], dtype=uint64)
```

Karim noticed that `stack_detector_data` was giving different results for missing pulseId (uint64) than missing cellId (uint16). We have documented it as filling with 0 by default for integers.

tmichela:
Fixes #103

### DIFF
--- a/extra_data/stacking.py
+++ b/extra_data/stacking.py
@@ -53,7 +53,7 @@ def stack_data(train, data, axis=-3, xcept=()):
     return np.stack(ordered_arrays, axis=axis)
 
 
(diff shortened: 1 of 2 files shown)


RobertRosca:
LGTM 👍

PR #105, opened 2020-11-16
Transition to GitHub Workflow

RobertRosca:
With recent changes to Travis we should move to using GitHub Workflows for the testing instead.


RobertRosca:
Oops, did the PR from my fork so the tests won't run here for security reasons.

Made the same PR on my fork and the tests run successfully there: https://github.com/RobertRosca/EXtra-data/pull/1

.github/workflows/tests.yml
```diff
@@ -0,0 +1,42 @@
+name: Tests
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
tmichela, line 7: should we run it on all PR instead?

tmichela (commented):
You should reuse the existing makefile to set the environment and run test, unless you have a good reason to change that?

.github/workflows/tests.yml
RobertRosca: Can do, that's not a problem, it's just usually 'convention' to only run on PR's to the main branch (at least from what I've seen).

RobertRosca:
Mostly a preference from me, I think running it from the Makefile obscures what's going on in the CI logs and also when you're reading the workflow. Now you can just read what it's doing, but if you call 'make test' or something then you have to also open the Makefile to see what that does. "Explicit is better than implicit" #zen

So I'd flip the question and ask why should it run the Makefile instead 😜

Also, I can add the commands for how to run the tests locally with the act tool if you'd like? That also (currently) requires a small workaround in the workflow which I'd have to add.

tmichela:
> So I'd flip the question and ask why should it run the Makefile instead stuck_out_tongue_winking_eye

Well, my opinion is that a good CI should only contains the workflow definition. all of the logic should be in a separate file so it is easy to reproduce (execute locally) what it is doing outside the ci runner.
And yes I know you can install the tool for local runner, but that's heavy machinery, does not always work, ...

And, by experience, for complex build/test/deployment procedure, it is definitely the best option to decouple those, it makes it easier to update anything, and also makes it easier to migrate the CI to a different framework :)

RobertRosca:
Ah yes I said this in person but I'll write it here as well:

I agree that this should be done for complex procedures for the reasons you mentioned, and that's how it's done for complex tests in projects like numpy (they have a ['travis-test.sh'](https://github.com/numpy/numpy/blob/master/tools/travis-test.sh) script that sets up and runs tests, which is what's called now by an [action](https://github.com/numpy/numpy/blob/master/.github/actions/action.yml) used by their [github workflows](https://github.com/numpy/numpy/blob/master/.github/workflows/build_test.yml) as well).

But in our case the workflow is `python3 -m pip install -U -e .[test]` and `python3 -m pytest -v --nbval-lax --cov=extra_data`, so calling `make test` in this case doesn't really seem worth it to me.

If the tests execution expands in the future and become more complex then it'd be worth moving it to a separate file but until then I don't think it's work keeping it like that just in case we might need it later.

.github/workflows/tests.yml
takluyver: Let's run on all PRs. It doesn't come up often, but if I make a secondary PR against a branch other than master, it's still useful to have CI run on it.

(I also tend to have it run on push for all branches, but I appreciate that's kind of wasteful if you're just going to make a PR and run the same tests anyway)

takluyver:
@RobertRosca tells me you want a tie-breaker on this. Somewhat boringly, I can see both sides. I like the idea of using platform-agnostic tools like make or tox so it's easy to switch to other CI providers, and the Makefile will probably bitrot if we're not testing it. But when a command like `make test` just launches one reasonably short test command, it also seems reasonable to run that directly. And it's nice to make installation & running tests two separate steps in CI so it's easy to see what failed.

As I don't have a strong preference one way or the other, I'm inclined to let it stand as Robert has already implemented it.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,42 @@
...
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - uses: actions/cache@v1
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-${{ matrix.python-version }}-pip-${{ hashFiles('**/setup.py') }}
```
takluyver, line 29: Does this actually save any significant time? I don't know about GH actions specifically, but I recall one of the CI platforms said that caching typically wasn't worth it for data that's downloaded over a fast link.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,42 @@
...
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install codecov pytest
+        python3 -m pip install .[test]
```
takluyver, line 35: Can we use one consistent way to invoke `pip`, rather than 3 different ways? :wink: 

Pytest is one of our test dependencies - `.[test]` will include it, so we don't need to list it separately.

The [codecov Github action](https://github.com/codecov/codecov-action) doesn't use the Python codecov package, so I don't think we need to install that. (It, uh... [downloads a bash script to execute instead](https://github.com/codecov/codecov-action/blob/239febf655bba88b16ff5dea1d3135ea8663a1f9/index.js#L32-L45)... maybe it's nicer to use the Python package directly. :-S )

.github/workflows/tests.yml
RobertRosca: Doesn't make much of a difference in this case since I don't think there are too many dependencies (just checked and `site-packages` is ~ 320M in my venv for it), but I listened to a podcast a few weeks ago with some pypi people and one of them mentioned their hosting/data transfer costs are like 800,000$+ a month, since then I try to cache pip stuff where possible, especially when running a matrix of tests :joy:

.github/workflows/tests.yml
RobertRosca: Oops yes that shouldn't be there, I'll fix it now, thanks!

.github/workflows/tests.yml
RobertRosca: Removed the restriction, now it runs on a push to master, and any PR

.github/workflows/tests.yml
takluyver: Fair enough. Though once it has cached a set of packages, it won't cache any newer versions because the cache key already exists (until we change `setup.py`). I don't know if/when caches are automatically dropped and have to be renewed.

(For the record, Fastly provides free CDN service for PyPI, so the $800k is probably an estimate of what it would cost if they stopped that. It's unlikely the PSF could actually afford that.)

RobertRosca: Oooooh, okay I feel less bad now :D

That's a good point about the cache not updating when newer packages are released, I guess this is a good time to bring up the option of adding in more restrictive package requirements, which would also let us use dependabot to automatically run tests when our dependencies have new releases.

I'm a fan of the slightly convoluted approach of using Poetry to specify dependencies, then dephell to convert that to a 'traditional' setup.py file ([see this as an example](https://github.com/RobertRosca/oscovida/blob/e430d5cf4b0699b6487e59968cb1be752060966f/setup.py#L45)) which lets you really control what dependencies you have.

But in this case we don't have that many dependencies so it'd be pretty easy to specify them by hand initially and then have dependabot take care of the updates for us. Think we discussed this before a bit, the main (and imo very large) benefit of this is that you'll get an automatic PR with the changes which would then trigger the tests as well, so we can find out if a dependency update caused issues much more easily and consistently.

takluyver: So my take is that the requirements in `setup.py` for a library should be 'abstract requirements', i.e. broad ranges of compatibility, so it can be installed alongside other packages which also require e.g. NumPy.

But maybe it makes sense to run the automated tests with pinned versions of dependencies, for the reasons you describe. It's always annoying when your tests suddenly start failing until you fix a compatibility issue with a new version of something.

Do you know if there's a good way to have abstract dependencies in setup.py and keep a  `test-requirements.txt` file up to date based on that?

RobertRosca: Yes, it would be possible to have a separate `requirements.txt` file which is used for (1) installing dependencies for the test and (2) used by dependabot to manage provide version bump PRs.

Do you think it would be worth adding that in?

PR #106, opened 2020-11-23
Add frozen requirements

RobertRosca:
Added a constraints file under `.github/dependabot/constraints.txt`. Constraints are used when installing the package for testing.

With the constraints file there we can now use dependabot to automatically create a PR when one of our dependencies has a new release, meaning that our CI will run when a dependency is updated and we can catch issues more quickly. I've put through the request for dependabot to have access to this repo but since I'm not an organization admin I can't approve it.

The constraints file is also important for package caching done as part of the CI, since without a list of package versions the cache will continue to be reused even if a dependency has a newer version available.

edit: also added python 3.9 to the test matrix since previously we only tested on 3.6 to 3.8


fangohr:
I have approved the dependabot request.

takluyver:
Thanks Robert!

RobertRosca:
🥳 I'll do the same for EXtra-geom later as well.

PR #110, opened 2020-12-09
Stack from xarray

daviddoji:
As discussed with @takluyver , the new 'stack_from_xarray' method will make it possible to stack (big) detector data from an existing xarray.
The method is basically the same as 'stack_detector_data', so we could simplify both.

### DIFF
--- a/extra_data/stacking.py
+++ b/extra_data/stacking.py
@@ -131,6 +131,42 @@ def stack_detector_data(train, data, axis=-3, modules=16, fillvalue=None,
     return stack
 
 
+def stack_from_xarray(xrdata, axis=-3, modules=16, fillvalue=np.nan,
+                      real_array=True):
+    
+    dtypes, shapes, empty_mods = set(), set(), set()
+    modno_arrays = {}
+    for (modno, data) in zip(xrdata.module.values, xrdata):
+        array = data
+        dtypes.add(array.dtype)
+        shapes.add(array.shape)
+        modno_arrays[modno] = array
+
+    if len(dtypes) > 1:
+        raise ValueError("Arrays have mismatched dtypes: {}".format(dtypes))
+    if len(shapes) > 1:
+        s1, s2, *_ = sorted(shapes)
+        if len(shapes) > 2 or (s1[0] != 0) or (s1[1:] != s2[1:]):
+            raise ValueError("Arrays have mismatched shapes: {}".format(shapes))
+        empty_mods = {n for n, a in modno_arrays.items() if a.shape == s1}
+        for modno in empty_mods:
+            del modno_arrays[modno]
+        shapes.remove(s1)
+
+    if max(modno_arrays) >= modules:
+        raise IndexError("Module {} is out of range for a detector with {} modules"
+                         .format(max(modno_arrays), modules))
+
+    dtype = dtypes.pop()
+    shape = shapes.pop()
+    stack = StackView(
+        modno_arrays, modules, shape, dtype, fillvalue, stack_axis=axis
+    )
+    if real_array:
+        return stack.asarray()
+
+    return stack
+
 class StackView:
     """Limited array-like object holding detector data from several modules.
 
--- a/extra_data/stacking.py
+++ b/extra_data/stacking.py
@@ -134,12 +134,13 @@ def stack_detector_data(train, data, axis=-3, modules=16, fillvalue=None,
 
 def stack_from_xarray(xrdata, axis=-3, modules=16, fillvalue=None,
                       real_array=True):
-    """Stack data from a xarray.
+    """Stack detector data from an xarray DataArray.
 
     Parameters
     ----------
-    xrdata: xarray
+    xrdata: xarray.DataArray
         (Big) detector data. Expected dims are:
+        ('module', 'dim_1', 'dim_2', 'dim_3', 'dim_4'). e.g.:
         ('module', 'train', 'pulse', 'slow_scan', 'fast_scan')
     axis: int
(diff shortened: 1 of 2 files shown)


extra_data/stacking.py
```diff
@@ -131,6 +131,42 @@ def stack_detector_data(train, data, axis=-3, modules=16, fillvalue=None,
...
+    if len(dtypes) > 1:
+        raise ValueError("Arrays have mismatched dtypes: {}".format(dtypes))
+    if len(shapes) > 1:
+        s1, s2, *_ = sorted(shapes)
+        if len(shapes) > 2 or (s1[0] != 0) or (s1[1:] != s2[1:]):
+            raise ValueError("Arrays have mismatched shapes: {}".format(shapes))
+        empty_mods = {n for n, a in modno_arrays.items() if a.shape == s1}
+        for modno in empty_mods:
+            del modno_arrays[modno]
+        shapes.remove(s1)
```
takluyver, line 154: We can skip all of these checks, and the sets for dtypes, shapes, empty_mods. They're necessary when coming from a dict-of-arrays, because the arrays could have different shapes or dtypes. But when we're extracting pieces of a single array, that can't happen.

extra_data/stacking.py
```diff
@@ -131,6 +131,42 @@ def stack_detector_data(train, data, axis=-3, modules=16, fillvalue=None,
     return stack
 
 
+def stack_from_xarray(xrdata, axis=-3, modules=16, fillvalue=np.nan,
+                      real_array=True):
+    
+    dtypes, shapes, empty_mods = set(), set(), set()
+    modno_arrays = {}
+    for (modno, data) in zip(xrdata.module.values, xrdata):
```
takluyver, line 139: I think this iteration will only work if modules is the first axis. We should do something like `data = xrdata.sel(module=modno)` to get the right data.

extra_data/stacking.py
```diff
@@ -131,6 +131,42 @@ def stack_detector_data(train, data, axis=-3, modules=16, fillvalue=None,
     return stack
 
 
+def stack_from_xarray(xrdata, axis=-3, modules=16, fillvalue=np.nan,
```
takluyver, line 134: Rather than taking an `axis=` parameter, I would suggest that we check which dimension is labelled `module` in the DataArray, and use its position, so the array (or `StackView`) has its axes in the same order as the input.

takluyver:
Thanks! Still needs:

- [ ] Tests (at least a basic one)
- [ ] Docstring
- [ ] Add in to the reference docs (near `stack_detector_data`)

Feel free to ask if it's not clear how to do any of those.

extra_data/stacking.py
takluyver: We should probably check that module numbers are unique, so that we can give a nice error message if they're not. But that's much simpler to write.

extra_data/tests/test_stacking.py
```diff
@@ -155,3 +156,16 @@ def test_stackview_squeeze():
 
     with pytest.raises(np.AxisError):
         sv.squeeze(axis=4)
+
+def test_stack_from_xarray():
+    coords = {'module': np.arange(2),
```
takluyver, line 161: I'd like the test to go as far as having a gap in the module numbers (e.g. `[0, 2, 3]`), and verifying that there is data in the expected place and the fill value exposed in the gap.

extra_data/stacking.py
```diff
@@ -131,6 +132,59 @@ def stack_detector_data(train, data, axis=-3, modules=16, fillvalue=None,
...
 
+def stack_from_xarray(xrdata, axis=-3, modules=16, fillvalue=None,
+                      real_array=True):
+    """Stack data from a xarray.
+
+    Parameters
+    ----------
+    xrdata: xarray
+        (Big) detector data. Expected dims are:
+        ('module', 'train', 'pulse', 'slow_scan', 'fast_scan')
```
takluyver, line 143: I think this only expects that there is a `module` dimension, right? The other names shouldn't matter.

extra_data/stacking.py
```diff
@@ -131,6 +132,59 @@ def stack_detector_data(train, data, axis=-3, modules=16, fillvalue=None,
...
+        (not a number) for floating-point data, and 0 for integers.
+    real_array: bool
+        If True (default), copy the data together into a real numpy array.
+        If False, avoid copying the data and return a limited array-like wrapper
+        around the existing arrays. This is sufficient for assembling images
+        using detector geometry, and allows better performance.
+
+    Returns
+    -------
+    stack: numpy.array
```
takluyver, line 159:
```suggestion
    stack: numpy.ndarray
```

extra_data/stacking.py
```diff
@@ -131,6 +132,59 @@ def stack_detector_data(train, data, axis=-3, modules=16, fillvalue=None,
...
+        Stacked data for requested data path.
```
takluyver, line 160:
```suggestion
        Stacked data in an unlabelled array.
```

takluyver (commented):
I've done some detailed review, but looking at this again, I'm back to wondering if it would be neater to accept xarray objects in EXtra-geom directly. I might try to look into that this afternoon.

takluyver:
It took me a while to get to it, but from EXtra-geom 1.2 (PR https://github.com/European-XFEL/EXtra-geom/pull/62/), you can pass an Xarray DataArray directly to be assembled as an image. I think that removes the need for this. Sorry it took so long!

I'll close this one now, but if there's a use case I've overlooked, we can reopen it.

PR #111, opened 2020-12-10
Add support for flat dictionaries when iterating over trains

philsmt:
This PR adds an optional field `flat_keys` to `DataCollection.trains()`, which arranges the data result returned each train by `(source, key)` tuples rather that nested `[source][key]` dictionaries. The motivation is to enable access similar to the new `KeyData` interface (also in pasha ^_^).

In theory the code is slightly slower due to a function call + dict acess (each 40-50 ns on my office machine) for each source and key, but this is probably not a big deal for this method.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -267,7 +267,8 @@ class DataCollection:
 
         return False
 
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks!

PR #113, opened 2020-12-15
Add support to match trains for selected sources/keys

philsmt:
When working with multiple sources, quite often I find it annoying to manually match the trains for which each source and/or key has valid data in a `DataCollection`. I'd like to provide this already on the `DataCollection` level, so that any further operations can safely assume to never run into empty trains.

This is a proof-of-concept for this by adding an optional argument `require_all` similar to the `.trains()` method, which causes the returned `DataCollection` to not only be selected down to the requested sources/keys, but also those trains, for which all said sources/keys have at least one data entry.

An alternative could be to add a new method specific to that purpose, e.g. `.match_trains(...)`, with the same signature as `.select(...)`, which only touches the trains and not sources. Also, one might extend the `.select_trains()` method to support passing source/key globs for selecting, something like `match_by=...`.

I chose `.select()` as I'd typically want to perform this operation when selecting the sources I'm interested in. As a positive side effect, we might get people to select their `DataCollection`s more often.


extra_data/reader.py
```diff
@@ -708,7 +712,50 @@ def select(self, seln_or_source_glob, key_glob='*'):
...
+            # respective INDEX section.
+
+            train_ids = self.train_ids
+
+            for source, keys in selection.items():
+                if source in self.instrument_sources:
+                    # For INSTRUMENT sources, the INDEX is saved by
+                    # key group, which is the first hash component. In
+                    # practice, this is always 'data', but this is not
+                    # guaranteed and there may be more than one.
```
takluyver, line 727:
```suggestion
                    # key group, which is the first hash component. In
                    # many cases this is 'data', but not always.
```

Modules of the 'big' detectors (AGIPD, LPD, DSSC) have 4 key groups: detector, header, image, trailer. I think I've been told at some point that the 'key groups' (as we call them) are designed around those detectors.

extra_data/reader.py
```diff
@@ -708,7 +712,50 @@ def select(self, seln_or_source_glob, key_glob='*'):
...
+                for f, group in product(self._source_index[source], groups):
+                    # Add the trains with data found for each key group
+                    # in each file.
+                    source_tids = np.union1d(
+                        f.train_ids[f.get_index(source, group)[1] > 0],
+                        source_tids)
+
+                # Remove any trains previously selected, for which this
+                # selected source had no data.
+                train_ids = np.intersect1d(train_ids, source_tids)
```
takluyver, line 749: I think we should intersect the different key groups within a source, rather than unioning them. If you've selected keys in two groups from the same source, you want `require_all=True` to select only trains where all of those keys have data.

With the DAQ writing the data, I think all key groups in a source will have data for the same trains, so this doesn't matter. But it's not a property of the file structure, and data reduction might not preserve that property, so we may as well anticipate the possibility.

takluyver (commented):
Thanks. I like the idea of doing it as part of `.select()`. I've suggested a somewhat pedantic change to the implementation.

extra_data/reader.py
philsmt: Hmmm, I fully agree, not sure what I was thinking here. A union makes no sense, as it may include trains where one or more selected key group are missing data. It may be a remnant of a previous attempt using sets and I misinterpreted the binary operator.

philsmt:
Thanks for the feedback. Are you happy with the wording `require_all`? After a few days, I wonder whether it is confusing, as people may assume it is related to the fact of selecting sources, and wondering what it means to not require all selected sources. You have a good idea to include the train aspect in there?

- `match_trains`
- `matching_trains`
- `full_trains`
- `require_matching_trains`
- `require_full_trains`
- `skip_partial_trains`
- `partial_trains` (detaulting to `True`)
- ...

takluyver:
I wondered about that too - I kind of want to get 'train' in the name, but I've not found a wording I like yet, and I don't want to make it too long. Using the same parameter as the `.trains()` method has something to be said for it, as well.

philsmt:
Alright! Let me add some tests then and we can give it a go.

takluyver:
Don't do too much work programming during your holidays, though. :wink: I'm still at work for today.

philsmt:
Working on EXtra-data is the fun part though, so don't worry ^_^
Added a test to the commit, done from my side.

extra_data/tests/test_reader_mockdata.py
```diff
@@ -489,6 +489,17 @@ def test_select(mock_fxe_raw_run):
     assert 'FXE_DET_LPD1M-1/DET/0CH0:xtdf' in sel2.instrument_sources
 
 
+def test_select_require_all(mock_sa3_control_data):
+    run = H5File(mock_sa3_control_data).select('*/BEAMVIEW2:daqOutput')
+    subrun = run.select('*/BEAMVIEW2:daqOutput', require_all=True)
```
takluyver, line 494: Can we include another source, with data for all trains, so the test checks that it's correctly getting the intersection of train IDs (as opposed to e.g. the union across selected sources)? Even the CONTROL data of the same camera should do.

takluyver (commented):
I'm happy with the implementation, but I'd like to see the test extended a bit. :+1:

extra_data/tests/test_reader_mockdata.py
philsmt: You're right, that's actually the most relevant use case. I've refactored the test to try several different selections including `*`.

takluyver:
Thanks @philsmt !

PR #114, opened 2020-12-18
Accept DataCollection and KeyData for source selection

philsmt:
Somewhat related to #113, this PR adds support to select sources (in `DataCollection`) ~~and trains (in both `DataCollection` and `KeyData`)~~ via those very rame types, effectively copying a particular property of those to another object. The respective semantics are unchanged, i.e.

- `.select()` raises an exception if sources are selected which are not actually part of the `DataCollection`
- ~~`.select_trains()` returns the intersection of the current object and whatever is used to select trains~~

During the implementation I found a peculiar bug in `DataCollection._expand_selection()`, which prevented an empty iterable (which evaluated to `False` in the check `keys or None` check) to be accepted even though this is advertised as selecting all keys. As a part of this fix, it also accepts `None` now to select all keys (which is used internally in the `DataCollection.selection` property).

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -615,7 +615,10 @@ class DataCollection:
                 if source not in self.all_sources:
                     raise SourceNameError(source)
 
-                res[source].update(keys or None)
+                if keys:
+                    res[source].update(keys)
+                else:
+                    res[source] = None
 
         elif isinstance(selection, Iterable):
             # selection = [('src_glob', 'key_glob'), ...]
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -626,6 +626,10 @@ class DataCollection:
                 self._select_glob(src_glob, key_glob)
                 for (src_glob, key_glob) in selection
             )
+        elif isinstance(selection, DataCollection):
+            return self._expand_selection(selection.selection)
+        elif isinstance(selection, KeyData):
+            res[selection.source] = {selection.key}
         else:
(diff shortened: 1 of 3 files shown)


takluyver:
Thanks! I think this makes sense for `.select()`, where there's currently no documented way to select sources & keys matching an existing DataCollection. I'm not so sure about selecting trains - it only saves a bit of typing over the explicit `run.select_trains(other_data.train_ids)`, and the `train_ids` attribute is documented. So my inclination is to point people to that  rather than handling another distinct possibility.

Good catch on the `_expand_selection` bug. I see you've also made a fix in `.deselect()`, which makes sense to me.

philsmt:
Thanks! Fair point on `train_ids` being documented as opposed to `selection`, it's only an abbreviation then. Would you prefer it being removed from the PR?

takluyver:
I'm inclined to say yes to removing it - I don't think the abbreviation is worth having a more complicated `.select_trains()` API accepting more kinds of object. I don't want to be the EXtra-dictator, though - if you think the abbreviation is worth it, we can ask Thomas M to provide a casting vote.

philsmt:
I've discussed this with Thomas M a bit, and we agreed on removing the train portion from this PR. He'd actually gone further and made `DataCollection.selection`, but I'd disagree here, as there'd be little use to this except using it to `DataCollection.select`, wo we might as well make it part of the API.

takluyver:
This LGTM - you're welcome to merge it if there's nothing more you want to do on it.

PR #122, opened 2021-01-04
LPD parallel gain

takluyver:
This is an attempt to support reading LPD data taken in parallel gain mode. Specifically, the `LPD1M` object should be able to construct useful indexes when retrieving the data, so you can refer to it with gain stage and pulse/cell ID.

```python
lpd = LPD1M(run, parallel_gain=True)
lpd.get_array('image.data')
```

@daviddoji - please could you try this out. It would be useful to know if it works with some real data taken in parallel gain mode, but also whether you think the interface makes sense as someone working with the data - does the returned data look like you'd expect?

To do:

- [x] Integrate with `.get_dask_array()`
- [x] Integrate with `.trains()` iterator
- [ ] Integrate with writing virtual CXI files (does anyone use this with LPD?)
- [x] Tests
- [x] Docs


takluyver:
Tried this with @daviddoji the other day. It seemed to work OK for one run we tried, but another one still had index issues. On investigating, we found that there seemed to be a misconfiguration when the data was recorded, so it only had 25 unique pulse IDs although there were 100 frames per train (in each of the 3 gain stages).

David said he would enquire about how that happens and how often it might occur. We may need to provide some way to get the data when neither pulse ID nor cell ID are useful.

daviddoji:
After discussion with the detector experts, the run where this PR fails seems to be indeed a misconfiguration in the detector settings and it should not happen. That particular run was one of the first trials in which the "parallel gain" mode for the LPD detector was used. We were playing around a bit with the values, as these settings were never used.

I have also tried with other runs with proper configuration and it works as expected ;)

takluyver:
Thanks David. I'll go ahead with the to-do list for this, then.

daviddoji:
No problem. Let me know if I can help with some of those tasks.

daviddoji:
I've noticed that using the 'detector' object, one can get the following expecting result:

image

But if one tries to access data from a 'run' without using the 'detector' object, one gets unexpected results:

image

Notice the 'trainId' coordinate showing a value of 300, instead of 'module: 1 train: 1 gain: 3 pulse: 100 slow_scan: 256 fast_scan: 256', as it is shown in the picture above.

After obtaining this array, one has to reconstruct a new reshaped xarray with:

```
coords = {
    "train" : list(set(image_data_raw.trainId.values)),
    "gain" : [0,1,2],
    "pulse" : list(range(8,408,4))
}
dims = ["train", "gain", "pulse", "slow_scan", "fast_scan"]
data = image_data_raw.values.reshape((len(list(set(image_data_raw.trainId.values))), 3,
                                                                  len(range(8,408,4)), 256, 256))

image_data_raw_ = xr.DataArray(data, dims=dims, coords=coords)
```

takluyver:
Ah, I missed that you were using the run object rather than the detector.

That is kind of expected, though the way it shows up is a bit confusing. `run.get_array()` doesn't know about the pulse information, so it gives you 300 frames from the detector, labelled with the train IDs. Of course, in your example, they will all have the same train ID, but there are still 300 entries along that direction. The job of the detector object is to know about how the detector labels data within a train.

daviddoji:
Thanks for the clarification. The only reason, at the moment, to use the `run.get_array()` method is because slicing operations are rather straightforward. 
I know that you were also working on providing such features to the detector objects :)

takluyver:
I've now added an `roi=` parameter for `det.get_array()`, as it's fairly easy to do, and I didn't want to do it in a separate PR and get a merge conflict.

Selecting a region of pixels may only make sense with a single module. If you use it with multiple modules, it will select the same region of data from each module.

takluyver:
OK, I think I'm ready for this to be reviewed. I've left parallel gain support out of the CXI file creation for now as I suspect it's not needed, but `det.get_array()`, `det.get_dask_array()` and `det.trains()` should all work with parallel gain data (using `parallel_gain=True`).

docs/lpd_data.ipynb
```diff
@@ -287,11 +287,11 @@
...
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
+   "cell_type": "markdown",
    "metadata": {},
-   "outputs": [],
-   "source": []
+   "source": [
+    "LPD data may also be recorded in *parallel gain* mode, resulting in high-, medium- and low-gain frames for each pulse. To read this kind of data with the correct labels, use `LPD1M(run, parallel_gain=True)`."
```
daviddoji, line 293: "... resulting in high-, medium- and low-gain, presented as 0, 1, and 2, respectively, frames ..."

daviddoji:
Just a minor addition. LGTM

takluyver:
Thanks for reviewing it :slightly_smiling_face:

PR #128, opened 2021-01-15
Fix select's dict API and broadening of keys by re-selecting

### LINKED ISSUE #95: Selected keys can be overridden
`.select()` is meant to always make a narrower selection from the current one, but @daviddoji found a case where it can get broader again:

```python
source = 'FXE_DET_LPD1M-1/DET/2CH0:xtdf'

sel = run.select(source, "image.data")
print(sel.keys_for_source(source))
sel = sel.select(source, '*')
print(sel.keys_for_source(source))
```

philsmt:
Fixes #95 by restricting any subsequent key selections to the already selected keys, in particular occuring when re-selecting `*`.

While addressing this issue, I've realized missing checks in the `dict` API of `DataCollection.select`, which I fixed before in addition to adding some tests for it.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -623,6 +623,14 @@ class DataCollection:
                     raise SourceNameError(source)
 
                 if keys:
+                    # keys may be None, which is allowed as an
+                    # undocumented behaviour to support passing a
+                    # DataCollection as the selector.
+
+                    for key in keys:
+                        if not self._has_source_key(source, key):
+                            raise PropertyNameError(key, source)
+
                     res[source].update(keys)
                 else:
                     res[source] = None
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -631,7 +631,15 @@ class DataCollection:
                         if not self._has_source_key(source, key):
                             raise PropertyNameError(key, source)
 
-                    res[source].update(keys)
+                    if self.selection[source] is not None:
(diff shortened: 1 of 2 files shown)


extra_data/reader.py
```diff
@@ -623,7 +623,23 @@ def _expand_selection(self, selection):
...
+                        # If the current DataCollection also selected
+                        # specific keys, make sure the new keys are a
+                        # subset of it.
+                        res[source].update(keys & self.selection[source])
+                    else:
+                        res[source].update(keys)
+                elif self.selection[source] is not None:
+                    res[source] = self.selection[source]
                 else:
                     res[source] = None
```
takluyver, line 644: I think this is correct, but it would be easier IMO to follow the code and be sure that all 4 cases (current & new selection limit keys, only current limits, only new limits, neither limit) are handled correctly if it was arranged either as four cases in one if/elif chain, or maybe two nested pairs of branches (which is exactly equivalent to the current code, just not using `elif` as a shortcut).

Or maybe it would be OK with just some clearer comments. I just know that I found it a bit unclear reading the diff as is.

philsmt: Funny, that was my initial implementation, until I realized 2/4 cases are `pass`. I'll gladly add them back to make it explicit.

takluyver:
Thanks Philipp!

PR #130, opened 2021-01-19
Try a <= constraint on scipy version in tests

takluyver:
The automatic PR to test with the latest scipy (#119) fails because it pins that specific version, and that's not compatible with Python 3.6. I still want EXtra-data to support 3.6 for a while, because that's what the main Anaconda installation on Maxwell has, and I suspect many people use EXtra-data in that.

So this is an attempt to see if dependabot will still work with a `<=` dependency, as a cheap way of allowing the tests to fall back to older versions as needed by the Python version. If it works, I'll do the same for all dependencies.

cc @RobertRosca

### DIFF
--- a/.github/dependabot/constraints.txt
+++ b/.github/dependabot/constraints.txt
@@ -14,6 +14,6 @@ pyparsing==2.4.7
 python-dateutil==2.8.1
 pytz==2020.5
 pyzmq==21.0.1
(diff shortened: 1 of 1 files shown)


takluyver:
Merging this because the aim is to see what dependabot does with #119, and it will only do that once this is in master.

PR #131, opened 2021-01-19
Warn about deprecated get_series() behaviour for some big detector data

takluyver:
Once we can remove this special case, then `run.get_series()` can become a simple wrapper around `key.series()`, like some of the other `run.get_*` methods. I suspect no-one is relying on this, but if anyone is, hopefully this warning will give us a chance to either rethink it, or suggest alternatives.


tmichela:
LGTM :rocket:

PR #139, opened 2021-02-03
JUNGFRAU multi-module data access

takluyver:
This enables things like this:

```python
jf = JUNGFRAU(run.select_trains(np.s_[:10]))
jf.get_array('data.adc')
```

Virtual CXI files and writing selected frames aren't included. I'm not sure if writing selected frames can ever really be done with the current data format, as it doesn't really allow storing a different number of frames from train to train.

I'd like to refactor this at some point, because currently the base class provides the extra functionality for selecting pulses on AGIPD/DSSC/LPD, which the JUNGFRAU data doesn't need. However, that complexity only kicks in for keys starting with `image.`, which JUNGFRAU doesn't have, so it's not a big practical problem.

cc @dallanto

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -86,6 +86,7 @@ class MPxDetectorBase:
     """
 
     _source_re = re.compile(r'(?P<detname>.+)/DET/(\d+)CH')
+    _main_data_key = 'image.data'  # This is correct for AGIPD/DSSC/LPD
     n_modules = 16
     # Override in subclass
     module_shape = (0, 0)
@@ -107,7 +108,7 @@ class MPxDetectorBase:
         # pandas' missing-data handling converts the data to floats if there
         # are any gaps - so fill them with 0s and convert back to uint64.
         mod_data_counts = pd.DataFrame({
-            src: data.get_data_counts(src, 'image.data')
+            src: data.get_data_counts(src, self._main_data_key)
             for src in source_to_modno
         }).fillna(0).astype(np.uint64)
 
@@ -886,6 +887,7 @@ class JUNGFRAU(MPxDetectorBase):
     _source_re = re.compile(
         r'(?P<detname>.+_(JNGFR|JF4M))/DET/(MODULE_|JNGFR)(?P<modno>\d+)'
     )
+    _main_data_key = 'data.adc'
     module_shape = (512, 1024)
 
 
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -59,7 +59,6 @@ class MultimodDetectorBase:
     """
 
     _source_re = re.compile(r'(?P<detname>.+)/DET/(\d+)CH')
-    n_modules = 16
     # Override in subclass
     _main_data_key = ''  # Key to use for checking data counts match
     module_shape = (0, 0)
@@ -264,6 +263,7 @@ class MultimodFastDetectorBase(MultimodDetectorBase):
     different number of frames to be stored for each train, which makes
     access more complicated.
     """
+    n_modules = 16
     _main_data_key = 'image.data'
 
     @staticmethod
@@ -850,7 +850,7 @@ class AGIPD1M(MultimodFastDetectorBase):
     ----------
 
     data: DataCollection
-      A data collection, e.g. from RunDirectory.
+      A data collection, e.g. from :func:`.RunDirectory`.
     modules: set of ints, optional
       Detector module numbers to use. By default, all available modules
       are used.
@@ -871,7 +871,7 @@ class DSSC1M(MultimodFastDetectorBase):
     ----------
 
     data: DataCollection
-      A data collection, e.g. from RunDirectory.
+      A data collection, e.g. from :func:`.RunDirectory`.
     modules: set of ints, optional
       Detector module numbers to use. By default, all available modules
       are used.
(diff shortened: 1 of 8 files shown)


extra_data/components.py
```diff
@@ -864,6 +863,118 @@ def _make_image_index(self, tids, inner_ids, inner_name='pulse'):
...
+      if the dataset includes more than one JUNGFRAU detector.
+    min_modules: int
+      Include trains where at least n modules have data. Default is 1.
+    """
+    # We appear to have a couple of different formats for source names:
+    # SPB_IRDA_JNGFR/DET/MODULE_1:daqOutput  (e.g. in p 2566, r 61)
+    # SPB_IRDA_JF4M/DET/JNGFR03:daqOutput    (e.g. in p 2732, r 12)
+    # This should catch both. We may need to generalise it more in the future.
+    _source_re = re.compile(
+        r'(?P<detname>.+_(JNGFR|JF4M))/DET/(MODULE_|JNGFR)(?P<modno>\d+)'
```
tmichela, line 888: Last time I checked, these 2 other naming form existed (hopefully there are no additional in the meantime :fearful: )
`FXE_XAD_JF1M/DET/RECEIVER-1`
`HED_IA1_JNGFR-1/DET/MODULE`

takluyver: Thanks, I'll add those to the regex. Do you have proposal & run numbers handy containing those patterns?

takluyver: Hmm, if it's really `HED_IA1_JNGFR-1/DET/MODULE` that might not work - this expects to find a module number. But if it's just one module, you don't really need this interface anyway, because you can do e.g. `run['HED_IA1_JNGFR-1/DET/MODULE', 'data.adc'].xarray()`.

tmichela: `HED_IA1_JNGFR-1/DET/MODULE`
`HED_IA1_JNGFR-2/DET/MODULE`
...
?

:upside_down_face:

tmichela: But I am also not certain that HED is using JF in a aggregated way :thinking:
I think they will be used together (if you remember the publication with the projected plan), but they won't be assembled in a case

dallanto: This would explain the naming - they are considering these as completely independent detectors, single-module...

takluyver: OK, let's leave the HED naming out for now, and find a way to make it work if we need to. I assume that the FXE naming has modules `RECEIVER-1`, `RECEIVER-2`, etc.?

tmichela: Looks like all JF devices have been consistently renamed :relieved: 
I interpret that the `500K` are independent detectors, while others are multimodules

# FXE

FXE_XAD_JF1M/DET/JNGFR01
FXE_XAD_JF1M/DET/JNGFR02

FXE_XAD_JF500K/DET/JNGFR03

# HED

HED_IA1_JF500K1/DET/JNGFR01
HED_IA1_JF500K2/DET/JNGFR02
HED_IA1_JF500K3/DET/JNGFR03
HED_IA1_JF500K4/DET/JNGFR04

# SPB

SPB_IRDA_JF4M/DET/JNGFR01
SPB_IRDA_JF4M/DET/JNGFR02
...
SPB_IRDA_JF4M/DET/JNGFR08

dallanto: As far as I know, SPB will stick to their assembly and naming (no further changes to be expected). And I think it's easier to deal with the HED naming now, as they number not only the JF500KN as independent detectors, but have a number in the terminal module segment as well.

takluyver: Yes, we can work with the `HED_IA1_JF500K2/DET/JNGFR02` style names, in case that's useful.

takluyver: I just realised that there's a question to for names like `HED_IA1_JF500K1/DET/JNGFR01`. The code tries to identify and check a 'detector name'. The obvious way to do that is to have `HED_IA1_JF500K1` as the name, but then it won't let you combine that data with `HED_IA1_JF500K2`, because that's a separate instance. We could make `HED_IA1_JF500K` the name, so it can pull together data from all of them in a single array - but that might be confusing if people do see them as separate detectors. Or we can leave the 500K names out of this for now (you can still use them as individual sources, of course).

Thoughts?

takluyver:
I've done a bit of refactoring. The diff has got quite big, but a lot of it is just shuffling around code that was already there.

There are now two base classes:

- `MultimodDetectorBase` has common machinery for all these detectors made of multiple sources.
- `MultimodFastDetectorBase` (I know, awkward name) extends that for the `image.data` detectors: AGIPD, LPD & DSSC.

extra_data/components.py
tmichela: > Or we can leave the 500K names out of this for now

hum, I lean in this direction. Seems clear that `500K` devices are meant to be used as standalone detectors. And this class represent mutimodular detectors (If we do this, then maybe emphasize in the docstring of the JUNGFRAU object that it is mean to abstract multimodular jungfrau detectors only). Unless there are use cases where it makes sens to handle and combine data from multiple detector (non modular) but we can review this once the case comes up.

tmichela:
> * `MultimodFastDetectorBase` (I know, awkward name)

An alternative would be something like `MultimodXtdfDetectorBase`, but that's maybe too much internal jargon

takluyver:
I'm not too concerned about internal jargon there, because that name isn't exposed to the users at all. If Xtdf is more precise, let's go with that. (Do we know what it stands for, btw?)

tmichela:
`XFEL Train Data Format`

tmichela:
LGTM :+1:

PR #140, opened 2021-02-05
Data access class for AGIPD 500K

takluyver:
All the machinery is already there for AGIPD1M, this just makes it easier to use for the mini-half.


tmichela:
LGTM \o/

tmichela (approved):

PR #141, opened 2021-02-05
Add parameters to let stack_detector_data work for JUNGFRAU

takluyver:
From the test, these are the parameters you need:

```python
_, data = run.select('*JF4M/DET/*', 'data.adc').train_from_index(0)

comb = stack_detector_data(
    data, 'data.adc', modules=8, pattern=r'/DET/JNGFR(\d+)', startsat1=True
)
```

The JUNGFRAU data access class from PR #139 is maybe a bit more convenient, because it doesn't need so many non-default parameters.

cc @dallanto

### DIFF
--- a/extra_data/stacking.py
+++ b/extra_data/stacking.py
@@ -55,7 +55,7 @@ def stack_data(train, data, axis=-3, xcept=()):
 
 def stack_detector_data(
         train, data, axis=-3, modules=16, fillvalue=None, real_array=True, *,
-        pattern=r'/DET/(\d+)CH', startsat1=False,
+        pattern=r'/DET/(\d+)CH', starts_at=0,
 ):
     """Stack data from detector modules in a train.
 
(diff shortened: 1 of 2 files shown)


tmichela:
hum, what do you think about searching for multiple patterns instead of the additional `pattern` argument? Are there broader use for it beyond JF?

dallanto:
If I get it right, both the ´´pattern´´and the ´´startsat´´arguments are required (only) for JUNGFRAU currently, and were introduced for it? I don't mind, but then Thomas' question would also be applicable for ´´startsat´´ ...

takluyver:
I reckoned that trying to put all the naming conventions into one pattern would get messy, and there might be other multi-module detectors in the future with other naming schemes. It might also get confusing if you have e.g. AGIPD & JUNGFRAU data together (from something like `.select('*/DET/*')`, whereas a more specific regex will probably fail in a more obvious way.

But it's definitely a trade-off, and there's a case for trying to be smarter about what the caller wants. The classes in components do a bit more guesswork, but they're specific to a particular type of detector, which makes that easier.

tmichela:
I don't have strong feeling either way. I'm just a bit concerned that regex are esoteric for many people :upside_down_face: 
Maybe we could give an example of a pattern that would work for jungfrau in the docstring?

tmichela:
`startsat1`: I can guess which from a computer scientists and physicist, made either of the AGIPD and JUNGFRAU module number indexing :D

nitpicking: I read startsat -> "start sat", can we make it like `starts_at=0`?

takluyver:
Good points both about the documentation and the parameter name. I've applied both suggestions.

tmichela:
:+1: LGTM

PR #144, opened 2021-02-08
Fix: allow `np.nan` as fill value for integer arrays in `Component.get_array`

tmichela:
and `get_dask_array`

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -177,10 +177,10 @@ class MultimodDetectorBase:
         fill_value = dtype.type(fill_value)
 
         return xarray.concat(
-            arrays,
+            [a.astype(dtype, copy=False) for a in arrays],
             pd.Index(index, name='module'),
             fill_value=fill_value
-        ).astype(dtype, copy=False)
+        )
 
     def get_array(self, key, *, fill_value=None, roi=(), astype=None):
         """Get a labelled array of detector data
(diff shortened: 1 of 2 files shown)


extra_data/components.py
```diff
@@ -170,18 +170,19 @@ def __repr__(self):
...
+        fill_value = dtype.type(fill_value)
 
-        # enforce that fill value is compatible with array dtype
-        value = dtype.type(value)
-        return value
+        return xarray.concat(
+            arrays,
+            pd.Index(index, name='module'),
+            fill_value=fill_value
+        ).astype(dtype, copy=False)
```
takluyver, line 183:
```suggestion
        return xarray.concat(
            [a.astype(dtype, copy=False) for a in arrays],
            pd.Index(index, name='module'),
            fill_value=fill_value
        )
```

It's not actually documented (or not obviously) that `xarray.concat` will convert the dtype if given `fill_value=nan`. I guess it's unlikely to change, but doing astype before that ensures it would still do the right thing if xarray ever changed that.

takluyver (commented):
I think there's one more `get_array` and `get_dask_array` to be edited (those for JUNGFRAU)

tmichela:
Thanks @takluyver indeed I forgot JF :sweat_smile:

takluyver:
Thanks :-)

PR #147, opened 2021-02-17
Scipy is not a dependency

takluyver:
I think this was left over from karabo_data, when geometry was part of the same package.


takluyver:
I'm going to merge this one without review, because it's simple. I've checked the CI logs, and scipy is not getting installed as a dependency of one of our other dependencies.

PR #154, opened 2021-03-08
Add architecture doc

takluyver:
Based on the reasoning of [this blog post](https://matklad.github.io//2021/02/06/ARCHITECTURE.md.html). I've chosen
to do this as part of the docs rather than a standalone markdown file, though that's up for discussion if someone thinks a separate file is better.

### DIFF
--- a/docs/architecture.rst
+++ b/docs/architecture.rst
@@ -1,0 +1,57 @@
+Architecture
+============
+
+.. note::
+
+   This page describes technical details about EXtra-data. You shouldn't need
+   this information to use it.
+
+Objects
+-------
+
+The :class:`DataCollection` class is the central piece of EXtra-data. It
+represents a collection of XFEL data sources and their keys, for a set of train
+IDs. It refers to data in one or more files (a run directory is often the
+starting point). A subset of its sources/keys or train IDs may be selected to
+make a new, more restricted :class:`DataCollection`.
+
+:class:`KeyData` represents data for a single source & key, selected from a
+``DataCollection`` like ``run[source, key]``. This data may still be spread
+across several files. The data can be loaded into a NumPy array, among other
+types.
+
+:class:`FileAccess` manages access to a single EuXFEL format HDF5 file,
+including caching index information. There should only be one ``FileAccess``
+object per file on disk, even if multiple ``DataCollection`` and ``KeyData``
+objects refer to it.
+
+Modules
+-------
+
+- ``exceptions`` defines some custom error classes.
+- ``file_access`` contains :class:`FileAccess` (described above), along with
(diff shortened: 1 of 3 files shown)


docs/architecture.rst
```diff
@@ -0,0 +1,57 @@
...
+:class:`FileAccess` manages access to a single EuXFEL format HDF5 file,
+including caching index information. There should only be one ``FileAccess``
+object per file on disk, even if multiple ``DataCollection`` and ``KeyData``
+objects refer to it.
+
+Modules
+-------
+
+- ``exceptions`` defines some custom error classes.
+- ``file_access`` contains :class:`FileAccess` (described above), along with
```
takluyver, line 32: I tried to order this list from low-level to high-level, but I'm not convinced that's useful - lots of modules don't fit neatly into that structure. Should it just be alphabetic? Or is there a better way to organise them?

tmichela:
I'd probably order it alphabetically, but the list of file is rather short so that is not an issue to order then differently if you think that makes more sens

LGTM

docs/architecture.rst
```diff
@@ -0,0 +1,57 @@
...
+objects refer to it.
+
+Modules
+-------
+
+- ``exceptions`` defines some custom error classes.
+- ``file_access`` contains :class:`.FileAccess` (described above), along with
+  machinery to keep the number of open files under a limit.
+- ``locality`` can check whether files are available on disk or on tape,
+  and is specific to EuXFEL & DESY infrastructure.
```
egorsobolev, line 35: It should work with any dCache, not only with DESY or EuXFEL

takluyver: Thanks! I thought we were checking for specific paths as part of it, but I was wrong. I've corrected that.

docs/architecture.rst
```diff
@@ -0,0 +1,59 @@
...
+
+:class:`.FileAccess` manages access to a single EuXFEL format HDF5 file,
+including caching index information. There should only be one ``FileAccess``
+object per file on disk, even if multiple ``DataCollection`` and ``KeyData``
+objects refer to it.
+
+Modules
+-------
+
+- ``cli`` contains command-line interfaces (so far only
```
philsmt, line 31: It feels a bit like nitpicking, but given the purpose of this section is to learn your way around the code, it may be confusing. We define five `console_scripts` entry points currently, and only one (the mentioned `make-virtual-cxi`) is actually in this package, the others are in `lsxfel`, `export` and `validation`.

We could either actually move this code there, or add something like "miscellaneous" or "auxiliary" here?

philsmt: Definitely not, but in that case I'm happy with the text 😄

PR #155, opened 2021-03-10
Fix detector get_array with ROI

takluyver:
Raw xtdf detector (AGIPD/DSSC/LPD) data has an extra dimension which is only used for AGIPD. `det.get_array()` checks for this and shifts the specified ROI so it applies the pixel dimensions, not the meaningless length=1 dimension.

However, we were doing this shift inside a loop, so each data chunk (typically a sequence file) would add another dimension to the ROI, and before long we'd be trying to select a 4D array on 5 dimensions, causing an error. This fixes that.

We missed this because the tests for this bit of code only used a single sequence file. I've added a test with 2 sequence files per DSSC module, which failed before this change.

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -409,9 +409,11 @@ class XtdfDetectorBase(MultimodDetectorBase):
                 if dset.ndim >= 2 and dset.shape[1] == 1:
                     # Ensure ROI applies to pixel dimensions, not the extra
                     # dim in raw data (except AGIPD, where it is data/gain)
-                    roi = np.index_exp[:] + roi
+                    sel_args = (data_positions, np.s_[:]) + roi
+                else:
+                    sel_args = (data_positions,) + roi
 
-                data = f.file[data_path][(data_positions,) + roi]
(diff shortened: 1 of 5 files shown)


philsmt:
Sneaky one, but nice fix. LGTM
TIL you can add `np.index_exp` to `tuple`s 😄 

Should we make an issue to check that all tests run on more than one sequence file?

philsmt:
I saw you tagging it with 1.5 as milestone, how about a point release `1.4.1`?

takluyver:
Yup, `np.index_exp[]` actually returns a tuple.

It might be a good idea to check more things with >1 sequence file. We do have some tests that do that, but there weren't previously any for multi-module detectors. The range of different possibilities (different detector types, raw/proc, ROI, reduced data, selecting pulses by index or by pulseId...) means it's probably not practical to exhaustively test every possibility.

I may do this as 1.4.1, as there have only been a couple of small changes since 1.4. I don't think it makes much difference what we call the version.

PR #156, opened 2021-03-15
Create smaller files when testing writing selected frames

takluyver:
And delete them when the test passes. This reduces the disk space required to run tests, and dramatically reduces the size of the temp files hanging around after running the tests.


tmichela:
LGTM

PR #158, opened 2021-03-29
Add support for opening runs serially with RunDirectory()

JamesWrigley:
By default `RunDirectory()` will open multiple files in parallel with
`multiprocessing.Pool()`. If using extra_data in a daemonized
`multiprocessing.Process`, which does not allow creating child processes, this
will throw an exception.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -132,19 +132,23 @@ class DataCollection:
     @classmethod
     def from_paths(
             cls, paths, _files_map=None, *, inc_suspect_trains=True,
-            is_single_run=False
+            is_single_run=False, parallelize=True
     ):
         files = []
         uncached = []
+
+        def handle_open_file_attempt(fname, fa):
+            if isinstance(fa, FileAccess):
+                files.append(fa)
+            else:
+                print(f"Skipping file {fname}", file=sys.stderr)
+                print(f"  (error was: {fa})", file=sys.stderr)
+
         for path in paths:
             cache_info = _files_map and _files_map.get(path)
-            if cache_info:
+            if cache_info and ('flag' in cache_info):
                 filename, fa = cls._open_file(path, cache_info=cache_info)
-                if isinstance(fa, FileAccess):
-                    files.append(fa)
-                else:
-                    print(f"Skipping file {filename}", file=sys.stderr)
-                    print(f"  (error was: {fa})", file=sys.stderr)
+                handle_open_file_attempt(filename, fa)
             else:
                 uncached.append(path)
 
@@ -153,14 +157,15 @@ class DataCollection:
                 # prevent child processes from receiving KeyboardInterrupt
                 signal.signal(signal.SIGINT, signal.SIG_IGN)
 
-            nproc = min(available_cpu_cores(), len(uncached))
-            with Pool(processes=nproc, initializer=initializer) as pool:
-                for fname, fa in pool.imap_unordered(cls._open_file, uncached):
-                    if isinstance(fa, FileAccess):
-                        files.append(fa)
-                    else:
-                        print(f"Skipping file {fname}", file=sys.stderr)
-                        print(f"  (error was: {fa})", file=sys.stderr)
+            # Open the files either in parallel or serially
+            if parallelize:
+                nproc = min(available_cpu_cores(), len(uncached))
+                with Pool(processes=nproc, initializer=initializer) as pool:
+                    for fname, fa in pool.imap_unordered(cls._open_file, uncached):
+                        handle_open_file_attempt(fname, fa)
+            else:
+                for path in uncached:
+                    handle_open_file_attempt(*cls._open_file(path))
 
         if not files:
             raise Exception("All HDF5 files specified are unusable")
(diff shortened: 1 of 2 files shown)


JamesWrigley:
Alternative design I just thought of: automatically fall back to opening the files serially if a `Pool` cannot be created?

tmichela:
> Alternative design I just thought of: automatically fall back to opening the files serially if a `Pool` cannot be created?

That might be a better option, or check if it's running in the main process (`multiprocessing.current_process().name == 'MainProcess'`). The fact that files are opened in multiple processes is a technical details and I don't think users should have to think about that when opening a run, so I would not expose an argument for this (unless you see a useful usecase?)

takluyver:
I think it's a good idea to have the parameter, whether or not we also try to auto-detect what to do. Even if it's something not many people will use, it's useful to have the ability to control it. E.g. maybe people use it inside parallel workers run by something other than multiprocessing, and don't want another layer of parallelism. Or maybe they find that opening files in parallel is slower in some situation.

I'm generally a bit wary of things that try to guess e.g. whether to parallelise or not. If the guessing goes wrong, it's more to understand than an 'always does X unless you tell it differently' rule. *Maybe* this is a case where we can guess well enough, though.

JamesWrigley:
I think I lean towards keeping it as a parameter for the reasons @takluyver mentioned, being able to control the parallelism of code running in workers is something I often want to do.

tmichela:
Alright, I don't have particularly strong feeling either way. My reasoning was mainly that it would be ignored in 99% of the cases anyway since we read the cache files by default. Also I still hope we can get the overview files (@JamesWrigley [for reference](https://github.com/European-XFEL/EXtra-data/pull/69)) in a reasonable time frame :smirk:

extra_data/tests/test_reader_mockdata.py
```diff
@@ -679,6 +680,16 @@ def test_open_file(mock_sa3_control_data):
...
 
+def open_run_daemonized_helper(path):
+    RunDirectory(path, parallelize=False)
+
+def test_open_run_daemonized(mock_fxe_raw_run):
+    p = Process(target=open_run_daemonized_helper, args=(mock_fxe_raw_run,), daemon=True)
+    p.start()
+    p.join()
+
+    assert p.exitcode == 0
```
tmichela, line 691: You can maybe add the test that fails when the flag is set, for completeness?

Else LGTM

extra_data/reader.py
```diff
@@ -1215,7 +1223,7 @@ def H5File(path):
     return DataCollection.from_path(path)
 
 
-def RunDirectory(path, include='*', file_filter=locality.lc_any):
```
takluyver, line 1218: The option should probably be exposed in the `open_run()` function as well.

extra_data/tests/test_reader_mockdata.py
takluyver: I don't think it's so important to test that something fails :wink: . But I think it would be good to have a brief explanatory comment on this test, to make it clear what we care about once it doesn't have the context of this PR.

takluyver:
I've just noticed something unfortunate, which is not strictly related to this PR, but maybe this PR is a good place to fix it, or at least discuss how to fix it.

At the moment, the `from_paths` method assumes that if it gets `cache_info` for a given file, creating the FileAccess will be fast, so it just does that directly in the same process. It only uses multiprocessing where the cache data is missing.

This assumption was valid until recently. But now we're extending the cache with validity information (`INDEX/flag`), which requires opening the file even when the old cache is valid. This should only need to be done once, because then it rewrites the cache with the validity information added, but it means we have a slow opening step again. I noticed this when I tried to inspect a 12 TB FXE run (p 2729, r 44).

I see two possible fixes:

- Here: if `parallel=True`, do all the file opening in parallel, even when there's cache info. We could also debate having a third state (`parallel=None`?) to let the method guess what to do, though that's not essential.
- In the cache machinery: if the cache is 'old' (without the flag data), declare it entirely invalid and don't use anything from it. Maybe this is throwing the baby out with the bathwater, but maybe if we have to open the file it's not much slower to re-read the other metadata we cache.

tmichela:
Just finalizing this PR:
-  adding the `parallelize` argument to open_run
- invalidate cache if `flag` is missing

@takluyver

extra_data/reader.py
```diff
@@ -1443,18 +1449,23 @@ def RunDirectory(
         In newer files, trains where INDEX/flag are 0 are suspect. For older
         files which don't have this flag, out-of-sequence train IDs are suspect.
         If True (default), it tries to include these trains.
+    parallelize: bool
+        Enable or disable loading files in parallel. Particularly useful if
```
takluyver, line 1453:
```suggestion
        Enable or disable opening files in parallel. Particularly useful if
```

extra_data/reader.py
```diff
@@ -1500,6 +1511,10 @@ def open_run(
...
```
takluyver, line 1515:
```suggestion
        Enable or disable opening files in parallel. Particularly useful if
```

extra_data/tests/test_reader_mockdata.py
```diff
@@ -789,6 +790,16 @@ def test_open_file(mock_sa3_control_data):
         assert 'METADATA/dataSources/dataSourceId' in file_access.file
 
 
+def open_run_daemonized_helper(path):
+    RunDirectory(path, parallelize=False)
+
+def test_open_run_daemonized(mock_fxe_raw_run):
+    p = Process(target=open_run_daemonized_helper, args=(mock_fxe_raw_run,), daemon=True)
```
takluyver, line 797:
```suggestion
    # Daemon processes can't start their own children, check that opening a run is still possible.
    p = Process(target=open_run_daemonized_helper, args=(mock_fxe_raw_run,), daemon=True)
```

takluyver:
Thanks, LGTM

PR #159, opened 2021-03-30
Fix float train IDs when require_all is used during select

philsmt:
By chance, I stumbled upon a bug I introduced with #113. Turns out, the implicit conversion of a list passed to `np.union1d` yields an array with type `np.float64`, causing all train IDs to become float. This PR fixes this by starting with an empty `uint64` array instead.

Also, I added a conversion back into list, as `DataCollection.train_ids` is typically one.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -759,7 +759,8 @@ class DataCollection:
                     groups = ['']
 
                 for group in groups:
-                    source_tids = []
+                    # Empty list would be converted to np.float64 array.
+                    source_tids = np.empty(0, dtype=np.uint64)
 
                     for f in self._source_index[source]:
                         # Add the trains with data in each file.
@@ -774,6 +775,8 @@ class DataCollection:
             # Filtering may have eliminated previously selected files.
             files = [f for f in files
                      if np.intersect1d(f.train_ids, train_ids).size > 0]
+
+            train_ids = list(train_ids)  # Convert back to a list.
 
         else:
             train_ids = self.train_ids
--- a/extra_data/tests/test_reader_mockdata.py
+++ b/extra_data/tests/test_reader_mockdata.py
@@ -530,6 +530,11 @@ def test_select_require_all(mock_sa3_control_data, select_str):
     subrun = run.select(select_str, require_all=True)
     np.testing.assert_array_equal(subrun.train_ids, run.train_ids[1::2])
 
+    # The train IDs are held by ndarrays during this operation, make
+    # sure it's a list of np.uint64 again.
+    assert isinstance(subrun.train_ids, list)
+    assert all([isinstance(x, np.uint64) for x in subrun.train_ids])
+
 
 def test_deselect(mock_fxe_raw_run):
     run = RunDirectory(mock_fxe_raw_run)


tmichela:
an you add or update a test that check the type is correct in the returned `DataCollection`?

I wonder if it would be more efficient to do something like this, instead of doing a union for each file:
```python

source_tids = []
for f in self._source_index[source]:
    # Add the trains with data in each file.
    source_tids.append( f.train_ids[f.get_index(source, group)[1] > 0])

source_tids = np.unique(np.stack(tids))
# Remove any trains previously selected, for which this
# selected source and key group has no data.
train_ids = np.intersect1d(train_ids, source_tids).tolist()
```

tmichela:
LGTM nonetheless

philsmt:
> Can you add or update a test that check the type is correct in the returned `DataCollection`?

I wasn't sure. When I checked why test didn't fail, I realized that `np.testing.assert_array_equal` does not care for type, and it seems there's no function in `numpy` for testing to do that.

If we'd do that more explicitly, I'd like to also define whether it's supposed to be a list.

> I wonder if it would be more efficient to do something like this, instead of doing a union for each file:
> 
> ```python
> source_tids = []
> for f in self._source_index[source]:
>     # Add the trains with data in each file.
>     source_tids.append( f.train_ids[f.get_index(source, group)[1] > 0])
> 
> source_tids = np.unique(np.stack(tids))
> # Remove any trains previously selected, for which this
> # selected source and key group has no data.
> train_ids = np.intersect1d(train_ids, source_tids).tolist()
> ```

From the top of my head without actual code or profiling, I'd wager you're doing more work here in stacking and finding uniques across all train IDs, rather than filtering out iteratively (with possibly less in each iteration). But it may heavily depend on the case.

tmichela:
> From the top of my head without actual code or profiling, I'd wager you're doing more work here in stacking and finding uniques across all train IDs, rather than filtering out iteratively (with possibly less in each iteration). But it may heavily depend on the case.

yeah, nevermind me :sweat_smile: I've just benchmarked it and the actually have very similar performances...

tmichela:
> If we'd do that more explicitly, I'd like to also define whether it's supposed to be a list.

We always store it as a list AFAIK :thinking: At least reading the train ids from file is stored in a list, and `select_train_ids` always returns a list as well

takluyver:
I think we have an unfortunate inconsistency that `DataCollection.train_ids` (the main public API) is a list but `FileAccess.train_ids` (lower level) is an array.

philsmt:
> I think we have an unfortunate inconsistency that `DataCollection.train_ids` (the main public API) is a list but `FileAccess.train_ids` (lower level) is an array.

Yup, that's why I suspect more places where it might end up being an `ndarray`. It's definition-by-convention, so I'd hesitate to actually test for it. But I don't mind making it explicit it should be a list. Or `ndarray`.

philsmt:
Okay, so `DataCollection.train_ids` is supposed to be a list and I'll explicit tests for it?

takluyver:
Yup, that sounds right. In fact, we do even [document that it's a list](https://extra-data.readthedocs.io/en/latest/reading_files.html#extra_data.DataCollection.train_ids).

PS did you see my suggested change above? I think it may have got lost during the array/list discussion. :slightly_smiling_face:

philsmt:
> Yup, that sounds right. In fact, we do even [document that it's a list](https://extra-data.readthedocs.io/en/latest/reading_files.html#extra_data.DataCollection.train_ids).

Huh, skipped over it. Although seeing it now, "list" might sometimes be seen as a generic term rather than the specific type.

> PS did you see my suggested change above? I think it may have got lost during the array/list discussion. 🙂

Certainly, just pushed this back for a day. Thanks! 😃

takluyver:
Thanks, this LGTM :-)

Issue #27, opened 2020-02-27
Provide a means to access metadata fields


takluyver:
The new DAQ version records a few extra fields in the `METADATA` section of files, including start time and DAQ version. It should be possible to get this information through EXtra-data.

Complication: we allow combining data from more than one run. What should happen where the metadata differs?

takluyver:
This will be kind of tricky to provide in a nice way, because internally EXtra-data knows about a collection of files providing data, which are not necessarily in the same run - e.g. if you `.union()` two runs together. So we can't just pull the metadata from a random file and say it applies to the whole `DataCollection`.

It would make logical sense to take the minimum for start time, the maximum for end time, and check that other things are consistent across all files. But reading from every file in a run to get a simple piece of metadata would make it ridiculously slow.

So I guess we'll have to bake in some knowledge of runs as groups of files with common metadata.

takluyver:
Just to note, we did get a support request from a scientist who wanted the start times of runs. He was looking at data which didn't contain this metadata, so this wouldn't immediately have helped (he'll use control data timestamps instead), but there are use cases for it.

PR #164, opened 2021-04-09
Method to get RUN values

### LINKED ISSUE #26: Provide a means to access RUN data

takluyver:
Closes #26

To keep things fast & simple, this just selects an arbitrary file for the relevant source to get the RUN data from. Within a run, the value should be the same in all files.

If you combine runs with `.union()`, you'll get the value from an arbitrary one of the runs. If that was likely to come up often, I'd want to load the value from each run and error if they're inconsistent. But I think merging runs is relatively unusual, and checking for it has a cost in either performance or complexity, so for now I'm ignoring it.

```
In [2]: run = open_run(2702, 173)                                               

In [3]: run.get_run_value('HED_OPT_PAM/CAM/SAMPLE_CAM_4', 'triggerMode')        
Out[3]: 'ExtTrigMult'

In [4]: run.get_run_value('HED_OPT_PAM/CAM/SAMPLE_CAM_4', 'triggerMode.value')  
Out[4]: 'ExtTrigMult'
```

### DIFF
--- a/docs/reading_files.rst
+++ b/docs/reading_files.rst
@@ -128,9 +128,9 @@ below, e.g.::
    .. automethod:: select_trains
 
 The run or file object (a :class:`DataCollection`) also has methods to load
-data by sources and keys. Some of them are directly equivalent to the options
-above, but :meth:`.get_dataframe` and :meth:`.get_virtual_dataset` offer extra
-capabilities.
+data by sources and keys. :meth:`get_array`, :meth:`get_dask_array` and
+:meth:`get_series` are directly equivalent to the options above, but other
+methods offer extra capabilities.
 
 .. class:: DataCollection
    :noindex:
@@ -157,6 +157,10 @@ capabilities.
 
       .. seealso::
         :doc:`parallel_example`
+
+   .. automethod:: get_run_value
+
+      .. versionadded:: 1.5
 
 .. _data-by-train:
 
--- a/docs/reading_files.rst
+++ b/docs/reading_files.rst
@@ -160,7 +160,11 @@ methods offer extra capabilities.
 
    .. automethod:: get_run_value
 
-      .. versionadded:: 1.5
+      .. versionadded:: 1.6
+
+   .. automethod:: get_run_values
+
+      .. versionadded:: 1.6
 
 .. _data-by-train:
 
--- a/extra_data/__init__.py
+++ b/extra_data/__init__.py
@@ -36,6 +36,9 @@ program. If not, see <https://opensource.org/licenses/BSD-3-Clause>
 __version__ = "1.4.1"
 
 
+from .exceptions import (
+    SourceNameError, PropertyNameError, TrainIDError, MultiRunError
+)
 from .keydata import KeyData
 from .reader import *
 from .stacking import *
--- a/extra_data/exceptions.py
+++ b/extra_data/exceptions.py
@@ -27,3 +27,13 @@ class TrainIDError(KeyError):
 
     def __str__(self):
         return "Train ID {!r} not found in this data".format(self.train_id)
+
+
+class MultiRunError(ValueError):
+    def __str__(self):
+        return (
+            "The requested data is only available for a single run. This "
+            "EXtra-data DataCollection may have data from multiple runs, e.g. "
+            "because you have used .union() to combine data. Please retrieve "
+            "this information before combining."
+        )
--- a/extra_data/file_access.py
+++ b/extra_data/file_access.py
@@ -148,6 +148,7 @@ class FileAccess:
         self._index_cache = {}
         # {source: set(keys)}
         self._keys_cache = {}
+        self._run_keys_cache = {}
         # {source: set(keys)} - including incomplete sets
         self._known_keys = defaultdict(set)
 
@@ -311,6 +312,29 @@ class FileAccess:
                 res.add(key.replace('/', '.'))
 
         self.file[group].visititems(add_key)
+        self._keys_cache[source] = res
+        return res
+
+    def get_run_keys(self, source):
+        """Get the keys in the RUN section for a given control source name
+
+        Keys are found by walking the HDF5 file, and cached for reuse.
+        """
+        try:
+            return self._run_keys_cache[source]
+        except KeyError:
+            pass
+
+        if source not in self.control_sources:
+            raise SourceNameError(source)
+
+        res = set()
+
+        def add_key(key, value):
+            if isinstance(value, h5py.Dataset):
+                res.add(key.replace('/', '.'))
+
+        self.file['/RUN/' + source].visititems(add_key)
         self._keys_cache[source] = res
         return res
 
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -622,6 +622,36 @@ class DataCollection:
             return val.decode('utf-8', 'surrogateescape')
         return val
 
+    def get_run_values(self, source) -> dict:
+        """Get a dict of all RUN values for the given source
+
+        This includes keys which are also in CONTROL.
+
+        Parameters
+        ----------
+
+        source: str
+            Control device name, e.g. "HED_OPT_PAM/CAM/SAMPLE_CAM_4".
+        """
+        if not self.is_single_run:
+            raise MultiRunError
+
+        if source not in self.control_sources:
+            raise SourceNameError(source)
+
+        # Arbitrary file - should be the same across a run
+        fa = self._source_index[source][0]
+        res = {}
+        def visitor(path, obj):
+            if isinstance(obj, h5py.Dataset):
+                val = obj[0]
+                if isinstance(val, bytes):
+                    val = val.decode('utf-8', 'surrogateescape')
+                res[path.replace('/', '.')] = val
+
+        fa.file['RUN'][source].visititems(visitor)
+        return res
+
     def union(self, *others):
         """Join the data in this collection with one or more others.
 
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1087,6 +1087,8 @@ class DataCollection:
                         else:
                             entry_info = ""
                         dt = ds.dtype
+                        if h5py.check_string_dtype(dt):
+                            dt = 'string'
                         print(f"      - {k}\t[{dt}{entry_info}]")
 
         print()
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1072,7 +1072,22 @@ class DataCollection:
             print('  -', s)
             if any(p.match(s) for p in details_sources_re):
                 # Detail for control sources: list keys
-                keys_detail(s, sorted(self.keys_for_source(s)), prefix='    - ')
+                ctrl_keys = self.keys_for_source(s)
+                keys_detail(s, sorted(ctrl_keys), prefix='    - ')
+
+                run_keys = self._source_index[s][0].get_run_keys(s)
+                run_only_keys = run_keys - ctrl_keys
+                if run_only_keys:
+                    print('    + extra RUN keys (1 entry per run):')
+                    for k in run_only_keys:
+                        ds = self._source_index[s][0].file[f"/RUN/{s}/{k.replace('.', '/')}"]
+                        entry_shape = ds.shape[1:]
+                        if entry_shape:
+                            entry_info = f", entry shape {entry_shape}"
+                        else:
+                            entry_info = ""
+                        dt = ds.dtype
+                        print(f"'      - '{k}\t[{dt}{entry_info}]")
 
         print()
 
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -574,6 +574,42 @@ class DataCollection:
         """
         return self._get_key_data(source, key).dask_array(labelled=labelled)
 
+    def get_run_value(self, source, key):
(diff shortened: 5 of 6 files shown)


takluyver:
Regarding combining runs: we're going to face the same question for reading metadata fields like run number & start time (#27), and I'm less convinced that 'arbitrary one' is the right answer there. So we might decide we need some mechanism for tracking groups of files associated with the same run anyway.

philsmt:
How unfortunate we missed the 1.5 milestone for this one, as it happens I've got a use case for this coming up soon 😞 

> Regarding combining runs: we're going to face the same question for reading metadata fields like run number & start time (#27), and I'm less convinced that 'arbitrary one' is the right answer there. So we might decide we need some mechanism for tracking groups of files associated with the same run anyway.

Depending on the experiment, combining runs is actually fairly common, e.g. for lots of AMO-related experiments. Here you often need to collect statistics, but spread the same configuration across several runs if some parameter tends to be unstable.

I'd propose rather than introducing a new layer into EXtra-data to track the run a `DataCollection` belongs to and resolve conflicts, we could simply refuse providing a RUN value whenever it's not clear (based on a cheap check) all files belong to a particular run. Hence, when a user is combining runs, they have to sort out the conflict explicitly on their side.

takluyver:
I like that idea, at least as a starting point - we can do something smarter later if people need it. For the first cut, I'll assume that any DataCollection that comes from `.union()` may be multiple runs.

Releases aren't hard to make, so if you want this soon, we can make 1.6 after merging the PR.

philsmt:
Playing around with having keys only in `RUN` and not `CONTROL`, we might want to extend `lsxfel`? It currently lists only what can be found in `CONTROL`.

philsmt:
The ultimate solution should probably to list the keys found in either, and have access transparently choose the clostest `CONTROL` value to a given train and `RUN` if there's none in EXtra-data proper. But that's an entirely different beast.

For the interim, given `lsxfel` already has this header for `CONTROL` data
> N control sources: (1 entry per train)

It might make the most sense to add a section for `RUN` data (if present):

> M control sources: (1 entry per run)

Or more simple and to integrate the section name:

> M run sources:

takluyver:
How about this:

```
23 control sources: (1 entry per train)
  - HED_EXP_FDI/CAM/CAM_3
  - HED_IA1_TSC-1/CAM/TARGET_FRONT
  - HED_IA1_TSC-1/PROC/TARGET_FRONT
  - HED_IA1_TSC-2/CAM/TARGET_BACK
  - HED_IA1_TSC-2/PROC/TARGET_BACK
  - HED_OPT_IPM/ADC/1
  - HED_OPT_PAM/CAM/SAMPLE_CAM_4
    - Logger.file.maxBackupIndex.timestamp	[uint64]
    - Logger.file.maxBackupIndex.value	[uint32]
    - Logger.file.maxFileSize.timestamp	[uint64]
    ...
    - temperature.timestamp	[uint64]
    - temperature.value	[float64]
    + extra RUN keys (1 entry per run):
      - Logger.file.category.timestamp	[uint64]
      - Logger.file.category.value	[string]
      - Logger.file.filename.timestamp	[uint64]
      - Logger.file.filename.value	[string]
      - Logger.file.pattern.timestamp	[uint64]
      ...
```

philsmt:
Works great on the files I've created, thanks!

I'd probably not sort `RUN` below `CONTROL` (in particular given your example regarding DaqPolicy), but that's an aesthetic issue.

takluyver:
Would you put RUN above? Or mix the groups so they're in one alphabetic sequence? I'm guessing that RUN data are generally less important (given that we've got this far without exposing them).

philsmt:
I would probably put them on the same level (with either one first). We could remove the "1 entry per train" without `--detail` and just list sources, and show either section (if necessary) with `--detail`.
In reality, plenty of `CONTROL` data is actually `RUN` data, but nobody bothered to update the DaqPolicy.

takluyver:
I'm not quite following your description. You mean a separate section listing run sources like for control sources? I'm grouping them like this because there aren't (as far as I know) separate sources for run - the data in the RUN section is extra information attached to control sources. But I still want to make the distinction clear, because it affects how you can access and use them.

philsmt:
I understand, and either view of this is probably valid. I'll give my point another attempt:

Starting from Karabo, properties have either `DaqPolicy.OMIT` (only found in `RUN`) or `DaqPolicy.SAVE` (found in `RUN` once and `CONTROL` per train). Both is control data with the only distinction of meant to change over a run or not. So if looking at details, there is just the device with control data (i.e. is present in `CONTROL` and/or `RUN`, probably and). If looking with details, there are two sections with train-constant data and run-constant data (depending on which has any key).

Unfortunately, a property missing this is assigned `DaqPolicy.UNSPECIFIED`, which is equivalent to `DaqPolicy.SAVE`. And over the last year, little has changed on that front. Thus, it looks like most data is train-constant, even though it's run-constant.

takluyver:
That's how I see it too, but at the level of sources (aka Karabo devices), there is only one list of control sources - the RUN section may have extra properties within a source, but it doesn't have extra sources. So I'd like to display just one list of control sources, and split out the different keys under a given source when showing the details.

Is putting `(1 entry per train)` at the top of the section the sticking point? What about something like this?

```
23 control sources:
  - HED_EXP_FDI/CAM/CAM_3
  - HED_IA1_TSC-1/CAM/TARGET_FRONT
  - HED_IA1_TSC-1/PROC/TARGET_FRONT
  - HED_IA1_TSC-2/CAM/TARGET_BACK
  - HED_IA1_TSC-2/PROC/TARGET_BACK
  - HED_OPT_IPM/ADC/1
  - HED_OPT_PAM/CAM/SAMPLE_CAM_4
    - Control keys (1 entry per train):
      - Logger.file.maxBackupIndex.timestamp	[uint64]
      - Logger.file.maxBackupIndex.value	[uint32]
      - Logger.file.maxFileSize.timestamp	[uint64]
      ...
      - temperature.timestamp	[uint64]
      - temperature.value	[float64]
    - Extra run keys (1 entry per run):
      - Logger.file.category.timestamp	[uint64]
      - Logger.file.category.value	[string]
      - Logger.file.filename.timestamp	[uint64]
      - Logger.file.filename.value	[string]
      - Logger.file.pattern.timestamp	[uint64]
      ...
```

philsmt:
I tihnk that's what I meant all along, and in my head the `(1 entry per train)` applied to properties, not full sources. I probably confused myself by only looking at files with a single one. Thanks!

Love your idea!

While you're at it, the current implementation does not sort the `RUN` keys 😄

takluyver:
OK, I've implemented something similar to that:

```
23 control sources:
  - HED_EXP_FDI/CAM/CAM_3
  - HED_IA1_TSC-1/CAM/TARGET_FRONT
  - HED_IA1_TSC-1/PROC/TARGET_FRONT
  - HED_IA1_TSC-2/CAM/TARGET_BACK
  - HED_IA1_TSC-2/PROC/TARGET_BACK
  - HED_OPT_IPM/ADC/1
  - HED_OPT_PAM/CAM/SAMPLE_CAM_4
    - Control keys (1 entry per train):
      - Logger.file.maxBackupIndex.timestamp	[uint64]
      - Logger.file.maxBackupIndex.value	[uint32]
      - Logger.file.maxFileSize.timestamp	[uint64]
      - Logger.file.maxFileSize.value	[uint32]
      ...
      - temperature.timestamp	[uint64]
      - temperature.value	[float64]
    - Additional run keys (1 entry per run):
      - Logger.file.category.timestamp	[uint64]
      - Logger.file.category.value	[string]
      - Logger.file.filename.timestamp	[uint64]
      - Logger.file.filename.value	[string]
      - Logger.file.pattern.timestamp	[uint64]
      - Logger.file.pattern.value	[string]
      ...
```

I also wonder about hiding the `.timestamp` entries and adding a note at the bottom like 'Each key also has .timestamp (uint64)', to make the list more compact. But that is a separate question from this PR. It also depends a bit whether we want to preserve the idea that every key has an associated timestamp or not.

philsmt:
Thanks, seems to work nicely.

There's one issue I have, but I may break the EuXFEL file spec here myself. Naively, I've created most of my datasets with `create_dataset(key, data=value)`, which causes scalar datasets for, well, scalar values. This PR then throws:

```python
ValueError                                Traceback (most recent call last)
<ipython-input-7-ee1ffd82e7ba> in <module>
----> 1 run.get_run_value('SQS_REMI_DLD6/DET/BOTTOM', 'trigger.method')

~/code/EXtra-data/extra_data/reader.py in get_run_value(self, source, key)
    618             raise PropertyNameError(key, source)
    619
--> 620         val = ds[0]
    621         if isinstance(val, bytes):  # bytes -> str
    622             return val.decode('utf-8', 'surrogateescape')

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

/gpfs/exfel/sw/software/xfel_anaconda3/1.1/lib/python3.7/site-packages/h5py/_hl/dataset.py in __getitem__(self, args, new_dtype)
    764         if self.shape == ():
    765             fspace = self.id.get_space()
--> 766             selection = sel2.select_read(fspace, args)
    767             if selection.mshape is None:
    768                 arr = numpy.ndarray((), dtype=new_dtype)

/gpfs/exfel/sw/software/xfel_anaconda3/1.1/lib/python3.7/site-packages/h5py/_hl/selections2.py in select_read(fspace, args)
     99     """
    100     if fspace.shape == ():
--> 101         return ScalarReadSelection(fspace, args)
    102
    103     raise NotImplementedError()

/gpfs/exfel/sw/software/xfel_anaconda3/1.1/lib/python3.7/site-packages/h5py/_hl/selections2.py in __init__(self, fspace, args)
     84             self.mshape = ()
     85         else:
---> 86             raise ValueError("Illegal slicing argument for scalar dataspace")
     87
     88         self.mspace = h5s.create(h5s.SCALAR)

ValueError: Illegal slicing argument for scalar dataspace
```

Now I've seen the DAQ to use 1-len arrays instead of scalars, is this our assumption for any EuXFEL data?

takluyver:
We are currently assuming that RUN contains 1D, length 1 datasets rather than scalars. I'd agree that scalars make more sense in principle, and it wouldn't be too difficult to support them if we needed to, but I think I'd encourage you to match what the DAQ writes in this case. Consistency is nice. :slightly_smiling_face:

philsmt:
Ah, but there is CONTROL data with length > 1, e.g. `VECTOR_STRING` properties. It does not happen too often, but there are a few devices out there with it. It is true however that currently (might change in Karabo 2.11), a bound device cannot have `NDARRAY_ELEMENT` properties, i.e. multidimensional data.

So it is true we do not need to support scalar datasets, but CONTROL data may have any length.

takluyver:
Do you know where I might find some such data to check what shape it is? I would guess that both RUN and CONTROL become 2D, with the first dimension in RUN still being length 1.

takluyver:
Found one that fits with my expectation, but I'm interested if you have examples with strings.

```
% h5glance /gpfs/exfel/exp/SPB/202130/p900188/raw/r0455/RAW-R0455-AGIPD1MCTRL01-S00000.h5 RUN/SPB_DET_AGIPD1M-1/CC/MON_0/veto_pattern/
/gpfs/exfel/exp/SPB/202130/p900188/raw/r0455/RAW-R0455-AGIPD1MCTRL01-S00000.h5/RUN/SPB_DET_AGIPD1M-1/CC/MON_0/veto_pattern/
├timestamp	[uint64: 1]
└value	[uint32: 1 × 1000]
```

philsmt:
Ah, but I see the problem... here's an example file I found

```
RAW-R0077-DA03-S00000.h5/RUN (3 attributes)
├SQS_AQS_LAS
│ └SPEC
│   └800
│     ├acqRate
│     │ ├timestamp   [uint64: 1]
│     │ └value       [float64: 1]
│     ├bkgSpectrum
│     │ ├timestamp   [uint64: 1]
│     │ └value       [uint16: 1 × 2068]
```

So that's also on me. `VECTOR` data must have two axes 😞

takluyver:
I'm working on the assumption that `run_dset.shape == ctrl_dset[:1].shape`.

philsmt:
Works like a charm for both regular raw data with both `CONTROL` and `RUN`, as well as the new files with plenty of `RUN` values without `CONTROL`. Happy with the API, will save a lot of typing if you want just this one control value you doesn't change.

When going over a novice workflow for the upcoming users, I've realized however that opening a run with both `raw` and `proc` is somewhat awkward. I've come up with #174 to address this. Happy to merge this first and base #174 on master, or merge here first.

LGTM the rest!

takluyver:
I'd prefer to land this one and retarget #174 to master, as it sounds like you're happy with that as well. There's another branch I'm waiting to make a PR from once this lands. :slightly_smiling_face:

PR #165, opened 2021-04-09
Read timestamps from INDEX

### LINKED ISSUE #28: Provide a means to get train timestamps
The new DAQ version adds a timestamp field in the INDEX section, with a timestamp matching each train ID. We should have a way to access these.

takluyver:
File format 1.0 contains `INDEX/timestamp` with a timestamp for each train - though it's not exactly clear to me whether this comes from the time server or is created on the DAQ. This adds a `run.train_timestamps()` method to read the timestamps into an array or a pandas series. The values are converted into numpy datetimes.

Closes #28.

### DIFF
--- a/docs/reading_files.rst
+++ b/docs/reading_files.rst
@@ -67,6 +67,8 @@ to refer to all data associated with that 0.1 second window.
 
    .. automethod:: get_data_counts
 
+   .. automethod:: train_timestamps
+
    .. automethod:: info
 
 .. _data-by-source-and-key:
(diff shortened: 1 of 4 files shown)


tmichela:
>  though it's not exactly clear to me whether this comes from the time server or is created on the DAQ

I understand that it is created when a trainId arrives to the DAQ, so it is created at the DAQ level. In principle when a trainId arrives from the timeserver (unless the flag is 0)

tmichela:
LGTM

takluyver:
Thanks! So if this happens to construct the sequence from different DAQs, there could be small inconsistencies in the timestamps it returns - a graph of timestamp against train ID might not be a nice smooth line. Hopefully that's already covered in the documentation by referring to them as *approximate* timestamps.

PR #166, opened 2021-04-15
Update some details in README

takluyver:
We heard from someone who tried to run pytest without the implicit context of being in the source directory, so I thought I'd clarify that. Then I noticed that the detail about the required Python versions was out of date.


takluyver:
The tests seem to be hanging on `test_serve_files` in `test_streamer.py`.

takluyver:
And again with #169 merged :crossed_fingers:

PR #169, opened 2021-04-19
Select source to send in test_serve_files

takluyver:
Hopefully sending one source instead of all the sources makes this fast enough that we won't see timeouts.

### DIFF
--- a/extra_data/export.py
+++ b/extra_data/export.py
@@ -103,8 +103,9 @@ def serve_files(path, port, source_glob='*', key_glob='*',
     ----------
     path: str
         Path to the HDF5 file or file folder.
-    port: int
-        Local TCP port to bind socket to.
+    port: str or int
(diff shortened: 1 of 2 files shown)


takluyver:
I can't work out what's going on with these tests. My best guess is it's something to do with the machine it runs on, because both the repetitions (for the different file format versions) seem to fail on the same machine, and when it passes, the offending test runs much quicker (~1 second, not nearly 30). That suggests it's not always slow and only randomly too slow for the 30 second timeout.

takluyver:
I've adapted the `karabo-bridge-serve-files` to accept a full ZMQ endpoint string as an alternative to a port, and used a Unix socket (`ipc://...`) for the tests. Hopefully that will get rid of the issues with local network connections.

tmichela (approved):
:sweat: Thanks, LGTM!

PR #170, opened 2021-04-21
Make FileAccess.has_source_key() faster

takluyver:
I don't understand why, but `f.get(path, getclass=True)` is weirdly slow, at least in some circumstances. `f.get(path)`, which opens the object, is significantly faster.

This uses the same low-level functions as the core of `f.get(path)`, but skips retrieving some metadata about the dataset, which may save a bit of time.

### DIFF
--- a/extra_data/file_access.py
+++ b/extra_data/file_access.py
@@ -334,7 +334,11 @@ class FileAccess:
         else:
             raise SourceNameError(source)
 
-        if self.file.get(path, getclass=True) is h5py.Dataset:
+        # self.file.get(path, getclass=True) works, but is weirdly slow.
+        # Checking like this is much faster.
+        if (path in self.file) and isinstance(
+                h5py.h5o.open(self.file.id, path.encode()), h5py.h5d.DatasetID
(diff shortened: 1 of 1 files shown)


takluyver:
I filed an h5py issue about the surprising performance: https://github.com/h5py/h5py/issues/1881

dallanto:
For my understanding: firstly you replace the type check (class access) with a file-open-access, and second, to do so, you employ ``h5py.h5o.open(...)`` instead of ``self.file.get(path)`` (without ``getclass=true``). The speed benefit is a combination of both?

takluyver:
In my tests, `self.file.get(path)` was as fast as what I've done here. But I could imagine that in some other situation - a different file layout, or a different version of HDF5, it might be slower, because it gets a little bit more metadata about the dataset, so I made the same low-level `open()` call that `self.file.get(path)` uses.

Although we've seen how my guess of what might be faster or slower can be very wrong...

dallanto:
O.k. fair enough. I'd say LGTM : )

PR #171, opened 2021-04-21
Don't default to libver latest when writing virtual CXI files

takluyver:
This allows h5py with HDF5 1.12 to write files which can be read by HDF5 1.10.

A single argument to `libver` sets the default version it will try to use; it may still use newer parts of the file format if we do something that requires a new feature. We could specify a tuple `libver=('v110', 'v110')`, in which case attempting to use features incompatible with 1.10 would raise an error while writing the file.

- h5py docs: https://docs.h5py.org/en/stable/high/file.html#version-bounding
- HDF5 docs: https://portal.hdfgroup.org/display/HDF5/H5P_SET_LIBVER_BOUNDS

cc @dallanto

### DIFF
--- a/extra_data/write_cxi.py
+++ b/extra_data/write_cxi.py
@@ -202,7 +202,11 @@ class VirtualCXIWriter:
 
         log.info("Writing to %s", filename)
 
-        with h5py.File(filename, 'w', libver='v110') as f:
+        # Virtual datasets require HDF5 >= 1.10. Specifying this up front should
+        # mean it fails before touching the file if run on an older version.
+        # We also specify this as the maximum version, to ensure we're creating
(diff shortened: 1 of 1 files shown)


RobertRosca:
It might be worth adding a comment with a link to this PR or with a short explanation for why that's there. If somebody stumbles across it in the future they could do git blame or search through the repo, or they could just assume it doesn't matter and change it, comment would make it a bit clearer why it's there.

takluyver:
Added a comment. I also decided to set the maximum version, so we won't create 1.12-only files without a deliberate decision to allow that.

RobertRosca:
Cool, thanks, LGTM :+1:

PR #172, opened 2021-04-29
Some fixes for bad train handling

takluyver:
We found a few places where we weren't correctly excluding or including bad trains (according to the `inc_suspect_trains` flag). In particular, `key.select_trains()` and `run.select(..., require_all=True)`. This should fix them.

cc @philsmt

### DIFF
--- a/extra_data/file_access.py
+++ b/extra_data/file_access.py
@@ -163,6 +163,10 @@ class FileAccess:
     def valid_train_ids(self):
         return self.train_ids[self.validity_flag]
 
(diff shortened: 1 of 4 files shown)


philsmt:
LGTM, ran it through the same code paths I had issues with on Wedesnday. Now it all works out of the box, thanks!

Really need to migrate `pasha`'s `ExtraDataFunctor` here and test it, searched the problem there for a couple of hours 😁

philsmt (approved):

PR #173, opened 2021-05-03
Select pulses per gain stage for LPD in parallel gain mode

takluyver:
cc @daviddoji - ~~I haven't actually tested this yet,~~ but if you've got time to try it and see if it works for your use case, that would be good. Selecting `pulses=np.s_[:1]` should give you the first pulse from each gain stage.

Todo:

- [x] Do the same for selection using pulse ID
- [x] Add tests


daviddoji:
LGTM

PR #174, opened 2021-05-04
Extend open_run to return a single run DataCollection of both raw and proc

philsmt:
The "single-run" logic introduced in #164 makes it somewhat awkward to obtain a `DataCollection` of both `raw` and `proc` for a single run.  Just using `open_run(..., data='raw').union(open_run(..., data='proc'))` with the same parameters no longer has the `is_single_run` flag. In an effort to no longer mindlessly copy over all raw data (and/or having the pre-calibrated data available, too!), this will be common use case.

While it can be set manually, this is confusing and unnecessarily verbose. This PR extends `open_run` with an option for `data` to create this union automatically and re-enable the `is_single_run` flag.

As an WIP, I've added the string `'all'` and the `Ellipsis`, but I don't have any strong opinions of this. Given this flag is only useful on our infrastructure and there won't be any new folder suddenly beyond the current ones, a string like `'all'` is probably safe?

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1473,8 +1473,9 @@ def open_run(
     run: str, int
         A run number such as 243, '243' or 'r0243'.
     data: str
-        'raw' or 'proc' (processed) to access data from one of those folders.
-        The default is 'raw'.
+        'raw', 'proc' (processed) or 'all' (both 'raw' and 'proc') to access
+        data from either or both of those folders. If 'all' is used, sources
+        present in 'proc' overwrite those in 'raw'. The default is 'raw'.
     include: str
         Wildcard string to filter data files.
     file_filter: callable
@@ -1486,6 +1487,24 @@ def open_run(
         files which don't have this flag, out-of-sequence train IDs are suspect.
         If True, it tries to include these trains.
     """
+    if data == 'all':
+        common_args = dict(
+            proposal=proposal, run=run, include=include,
+            file_filter=file_filter, inc_suspect_trains=inc_suspect_trains)
+
+        # Create separate data collections for raw and proc.
+        raw_dc = open_run(**common_args, data='raw')
+        proc_dc = open_run(**common_args, data='proc')
+
+        # Deselect to those raw sources not present in proc.
+        raw_extra = raw_dc.deselect(
+            [(src, '*') for src in raw_dc.all_sources & proc_dc.all_sources])
+
+        # Merge extra raw sources into proc sources and re-enable is_single_run.
+        dc = proc_dc.union(raw_extra)
+        dc.is_single_run = True
+        return dc
+
     if isinstance(proposal, str):
         if ('/' not in proposal) and not proposal.startswith('p'):
             proposal = 'p' + proposal.rjust(6, '0')
--- a/extra_data/tests/test_reader_mockdata.py
+++ b/extra_data/tests/test_reader_mockdata.py
@@ -694,12 +694,31 @@ def test_open_run(mock_spb_raw_run, mock_spb_proc_run, tmpdir):
         assert {f.filename for f in run.files} == paths
 
         # Proc folder
-        run = open_run(proposal=2012, run=238, data='proc')
+        proc_run = open_run(proposal=2012, run=238, data='proc')
 
-        proc_paths = {f.filename for f in run.files}
+        proc_paths = {f.filename for f in proc_run.files}
         assert proc_paths
         for path in proc_paths:
             assert '/raw/' not in path
+
+        # All folders
+        all_run = open_run(proposal=2012, run=238, data='all')
+
+        # Raw contains all sources.
+        assert run.all_sources == all_run.all_sources
+
+        # Proc is a true subset.
+        assert proc_run.all_sources < all_run.all_sources
+
+        for source, files in all_run._source_index.items():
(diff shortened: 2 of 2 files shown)


extra_data/reader.py
```diff
@@ -1449,6 +1449,16 @@ def open_run(
         files which don't have this flag, out-of-sequence train IDs are suspect.
         If True, it tries to include these trains.
     """
+    if data is Ellipsis or data == 'all':
+        args = (proposal, run)
+        kwargs = dict(include=include, file_filter=file_filter,
+                      inc_suspect_trains=inc_suspect_trains)
+
+        dc = open_run(*args, data='raw', **kwargs).union(
+            open_run(*args, data='proc', **kwargs))
```
tmichela, line 1458: I'm afraid you're testing that in a very specific case. In practice there are keys duplication between raw and proc data, which will be conflicting when merging the 2 data collections.

philsmt: Ouch, how ironic the very problem I'm trying to avoid with REMI is biting this solution now. I'll take a look whether one could add an `ignore_conflicts` flag to `DataCollection.union()`.

takluyver: You could use `raw_extra = raw.deselect(proc)`

extra_data/reader.py
```diff
@@ -1449,6 +1449,16 @@ def open_run(
         files which don't have this flag, out-of-sequence train IDs are suspect.
         If True, it tries to include these trains.
     """
+    if data is Ellipsis or data == 'all':
```
takluyver, line 1452: I'd lean towards just one way (`data='all'`) to indicate this - `data=...` is cool but a bit cryptic.

(I think we used `data='proc'` rather than `corrected=True` because we anticipated one day adding an option to combine the two :+1: )

extra_data/reader.py
takluyver: Just remembering, I think `.deselect()` works for this where it can drop entire files for the resulting DataCollection, as with the big detectors which are in their own files, but it won't work if it deselects some sources but still has other data from the same files, because `.union()` will effectively reselect these. I think this was part of why we postponed this before.

(I'm not sure I've explained that very clearly. I can have another go if it doesn't make sense)

philsmt: Yes, I've just noticed this. When calibrated data adds a virtual source, `.deselect()` will throw an exception.
But since we filter on the source level here anyway, I might as well exclude those explicitly.

takluyver:
(Neat, Github automatically updated the target branch when I merged #164)

philsmt:
Okay, removed `Ellipsis` and added @takluyver 's beautiful solution to the duplicate sources problem. Works for me with both detector-type data (with _real_ calibrations) and my REMI data.

As a result from today's discussion, I will attempt to push for no longer copying raw data to `proc` and soft links in the interim. If this happens (and it will take time!), this will become really useful.

takluyver:
I think this should work where either the proc data contains entirely new sources (as we've agreed it should where we're designing it now), or where the sources which get corrected correspond to complete files (as for AGIPD, DSSC & LPD data, at least). It would still have issues if it needed to include part but not all of a raw file. I'm inclined to say that's good enough for the time being.

It would be good to add a quick test for this mechanism - it can be an extra piece inside `test_open_run` which already has everything it would need in place.

philsmt:
> It would still have issues if it needed to include part but not all of a raw file. I'm inclined to say that's good enough for the time being.

True, but I think this case does not exist in the wild. Traditional detectors have their own DA over xtdf or just a single source. As it happens, even in the REMI case, it's only the digitizer.
In the long run, the optimal solution might be a conflict resolution flag.

> It would be good to add a quick test for this mechanism - it can be an extra piece inside `test_open_run` which already has everything it would need in place.

Sure, coming in hopefully later today.

philsmt:
Ignore the failed CI please, I commited a test script by accident 🙈

extra_data/tests/test_reader_mockdata.py
```diff
@@ -701,6 +701,19 @@ def test_open_run(mock_spb_raw_run, mock_spb_proc_run, tmpdir):
...
+
+        for source, files in run._source_index.items():
+            for file in files:
+                if '/DET' in source:
+                    # AGIPD data is in proc.
+                    assert '/raw/' not in file.filename
+                else:
+                    # Non-AGIPD data is in raw.
+                    # (CAM, XGM)
+                    assert '/proc/' not in file.filename
```
takluyver, line 715: Any particular reason to check `not in` in both cases? I forget now why I did one `in` and one `not in` in the code just above.

Is it worth checking that both kinds of sources are actually there? At the moment, this would pass if there were no sources at all, I think.

philsmt: Ha, fair point with the sources check! But yes, I wanted to mirror your assertions.

philsmt: Added a test of sources for `open_run(data='all')`. Strictly speaking, the mock SPB run is a special case where `raw` contains all sources and `proc` only a select few, but this should neatly cover the two real world examples of

a) copy of all `raw` data in `proc` for traditiona detectors and
b) additional sources in `proc` for REMI data

takluyver:
Thanks, this LGTM

PR #175, opened 2021-05-04
Method to read METADATA for a given run

### LINKED ISSUE #27: Provide a means to access metadata fields

takluyver:
Return a dict of metadata for a single run. Building on the machinery from #164, this raises an error if you use it after union-ing data together.

Closes #27

### DIFF
--- a/extra_data/file_access.py
+++ b/extra_data/file_access.py
@@ -292,6 +292,25 @@ class FileAccess:
             counts = np.uint64((ix_group['last'][:ntrains] - firsts + 1) * status)
         return firsts, counts
 
+    def metadata(self) -> dict:
+        """Get the contents of the METADATA group as a dict
+
+        Not including the lists of data sources
+        """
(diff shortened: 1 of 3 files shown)


philsmt:
Thanks, this is also sorely needed. Any particular reason you choose to omit a verb in the method name?

takluyver:
No particularly good reason, though I've also done the same for the `train_timestamps` method in #165. I felt that all the `get_xyz` wasn't necessarily clearer. Both names are up for discussion, though, if you think the consistency of a `get_` prefix would be useful.

philsmt:
Ah, that precedence is good enough for me. Thanks, LGTM for the implementation.

PR #177, opened 2021-05-04
Remove some old h5py workarounds, bump required version

takluyver:
I spend time improving h5py, I may as well take advantage of the improvements.

These workarounds were for things fixed in h5py 2.10 - see the [what's new in 2.10](https://docs.h5py.org/en/stable/whatsnew/2.10.html) notes.


takluyver:
Thanks Egor!

PR #178, opened 2021-05-05
Include 'suspect' trains by default

takluyver:
As we've discussed internally, it seems that 'suspect' train IDs are more common than we had thought, and we've heard from a couple of scientists who were confused why data was suddenly missing. We had hoped it was just occasional problem trains from AGIPD, but other things are affected too.

This preserves all the machinery to exclude the suspect trains, but changes the default to include them again.

Q: should there be a different default for `extra-data-make-virtual-cxi`? Should we add a command-line option to control it?


egorsobolev:
I vote for command-line options. LGTM.

PR #181, opened 2021-05-10
Add new writer machinery

egorsobolev:
This implements the new writer machinery as we discussed recently.

Last week I needed to write files to euxfel. Thomas has already started this work on the writer2 branch, but it has stalled for now. I have finished the implementation.

In general, the interface follows the original concept, as I understood it. But I changed the declaration of the datasets. Instead of a two-level `dict`, they are declared as attributes of a new class:
```python
from extra_data.writer2 import FileWriter, DS

ctrl_grp = 'MID_DET_AGIPD1M-1/x/y'
inst_grp = 'MID_DET_AGIPD1M-1/x/y:output'

class MyFileWriter(FileWriter):
    nb = DS(ctrl_grp, 'param.numberOfBins', (), np.uint64)
    v = DS(inst_grp, 'azimuthal.profile', (1000,), float)

    class Meta:
        max_train_per_file = 10
        break_into_sequence = True
```
The `Meta` subclass sets the options of writing.
Then write data by trains:
```python
with MyFileWriter(new_file) as wr:
    # add data (funcion kwargs interface)
    wr.add(nb=100)

    for tid in trains:
        # create/compute data for 3 pulses
        v = np.random.randn(3, 1000)
        # add data (class attribute interface)
        wr.v = v
        # write train
        wr.write_train(tid, 0)
```
In 'CONTROL' section, a single entry per train is allowed, and data repeats in the following trains.

In 'INSTRUMENT' section, multiple entries per train are allowed, the number of entries may vary in different trains. There is the check that all datasets in the same source have the same number of entries in the same train.

cc @takluyver


takluyver:
Thanks Egor, and sorry that I haven't looked at this sooner.

I have to say up front that I don't really like the API based on subclassing FileWriter and adding special attributes to the subclass, the `Meta` class for options, or the special behaviour of assigning data to attributes. My preference is for something much more plain, with normal method calls, even though this would mean code using it was a bit longer. Of course, this is a matter of taste, and we don't have to do it my way. But it will be hard for me to provide detailed review on this when my preference is for something that looks very different. I'd like to ask @philsmt to do some review on the API as well, as he's probably happier with this kind of design than I am.

One concrete limitation I see is that it seems to be designed for writing a train at a time. I think we will often want to write an array with data for several trains in one go, especially for smaller data. Writing 50 single numbers for each of 100 trains is probably much less efficient than writing 50 arrays of 100 numbers each. But I'm not sure how to add that ability without making the automatic handling of sequence files significantly more complicated.

egorsobolev:
Thanks, Thomas.

I suspect that users also may have their own tastes. Fortunately, this design makes it easy to add various interfaces and satisfy different tastes. I added an interface that looks as close as possible to the `writer2` branch:

```python
import numpy as np
from extra_data.writer2 import FileWriter, DatasetDescr

ctrl_grp = 'MID_DET_AGIPD1M-1/x/y'
inst_grp = 'MID_DET_AGIPD1M-1/x/y:output'

sources = {
    ctrl_grp: {
        'param.numberOfBins': DatasetDescr('nb', (), np.uint64),
    },
    inst_grp: {
        'azimuthal.profile': DatasetDescr('v', (1000,), float)
    }
}

trains = range(10001, 10005)
with FileWriter.open('test.h5', sources, max_train_per_file=10) as wr:
    # add data (source name interface)
    wr.add_value_by_key(ctrl_grp, 'param.numberOfBins', 100)

    for tid in trains:
        # create/compute data for 3 pulses
        v = np.random.randn(3, 1000)
        # add data (source name interface)
        wr.add_value_by_key(inst_grp, 'azimuthal.profile', v)
        # write train
        wr.write_train(tid, 0)

```
I hope that is closer to your preferences ;)

As you can see I also added the method of passing the data by source name and key in the file. But even before it, there were two methods implemented as classical calls: `add_value(this, name, value)` and `add(this, **kwargs)`. 

Besides I've added the option to enable/disable the interface implemented as an assignment of data to class attributes. Indeed, this is little strange usage of attribute assignment but I found this funny ;)

Of cause, all these alternatives just small wrappers of a basic interface implemented in `add_value(this, name, value)`

@philsmt, could you please also take a look at this PR?

egorsobolev:
> One concrete limitation I see is that it seems to be designed for writing a train at a time. I think we will often want to write an array with data for several trains in one go, especially for smaller data. Writing 50 single numbers for each of 100 trains is probably much less efficient than writing 50 arrays of 100 numbers each. But I'm not sure how to add that ability without making the automatic handling of sequence files significantly more complicated.

Yes, the question on writing multi-train arrays at once a little more debatable and leads to some increase of complication anyway. I already thought about it, let me share some ideas.

It seems to me that the writing by trains corresponds to the vast majority of data processing that I can imagine. In fact, I wrote this implementation, because I needed to write azimuthal profiles in EuXFEL format. And of course, my processing algorithm is organized exactly by trains, mainly because EXtra-data naturally forces usage of such iterations.

In these algorithms, numbers appear by trains anyway. And the only way to improve performance with small data is buffering. I was hoping that the HDF library already buffers the write. If suddenly this is not the case, then we ourselves must implement buffering. It would be strange to leave the buffering on the user. I don't think that is an extreme complication.

Thus, this implementation covers a good portion of scenarios, and the buffering solves the performance problem with small data in these scenarios.

Thomas, do you know if the HDF library buffers the recording?

takluyver:
Thanks, the API in your new example is more what I was envisaging. I'd be more comfortable, though, if the plain API (the one I like ;-) was the base layer, and the fancy API with subclassing magic was built on top of it, rather than building the plain API as a wrapper around the fancy one, as it currently is. If something doesn't work as expected using the fancy API, I'd like to be able to drop down into the plain API so I can see what's going on without the extra layer of complexity.

> It seems to me that the writing by trains corresponds to the vast majority of data processing that I can imagine... And of course, my processing algorithm is organized exactly by trains, mainly because EXtra-data naturally forces usage of such iterations.

I don't think EXtra-data does really force train-by-train processing. It's a fairly obvious pattern with e.g. AGIPD data, where 1 train can already be quite a lot of data, but for a lot of other data it's natural to get an array for an entire run (or a sizeable chunk), and I've tried to emphasise this in the docs, because it's much more efficient where it's feasible.

> Thomas, do you know if the HDF library buffers the recording?

I believe it does so long as we keep the dataset open between writes. But there's still some overhead on each h5py operation - translating Python slicing to HDF5 calls, HDF5 validating its inputs, updating metadata, h5py checking if an error occurred... I did some work to reduce the overhead in h5py 3.x, but it's never going to be eliminated entirely (and I focused mainly on reading). My educated guess is that if we have a lot of small datasets, writing train by train will be significantly slower than writing arrays.

Writing train-by-train should certainly be an option - I don't want to force people to manually buffer their data if they're producing it train-by-train. But if they're producing it in chunks of, say, 100 trains at a time, I also don't want to force them to split it up into trains, or slow things down by doing that for them.

> all datasets in one source (indexed group) must have the same number of records in each train. Writing by trains makes it easy to keep the invariant that the written trains meet this requirement.

That makes sense, and I don't have a great answer ready for how we ensure this. I think it's worth thinking of something, though. My plan before was that you would supply an array of train IDs along with your data to be written, and if the pattern of train IDs conflicts with what's already in the index, you'd get an error. That's not particularly elegant, but...

egorsobolev:
>If something doesn't work as expected using the fancy API, I'd like to be able to drop down into the plain API so I can see what's going on without the extra layer of complexity.

C'mon, what extra layer of complexity? Just a small piece of code is located in `FileWriterMeta.__new__()` instead of `FileWriterBase.__init__()` as in the case of plain implementation and run at class declaration instead of instantiation time. A similar design perfectly works in many productions (e.g. Django is the engine of a huge number of sites) and described in cookbooks. Just look at this clear declarative style, no brackets, light constructor. I bet you'll love it too when it stops being unusual for you ;)

> I'd be more comfortable, though, if the plain API (the one I like ;-) was the base layer, and the fancy API with subclassing magic was built on top of it, rather than building the plain API as a wrapper around the fancy one, as it currently is.

Unfortunately, I don't see a nice way to make the plain version basic. Any try requires to twin the code, makes the implementations different, and will look ugly. It's not worth it. 

> My educated guess is that if we have a lot of small datasets, writing train by train will be significantly slower than writing arrays.

>Writing train-by-train should certainly be an option - I don't want to force people to manually buffer their data if they're producing it train-by-train.

Okay, then we have to implement buffering with direct writing of chunks. I'll make a benchmark to be sure. Let's do it in the next PR, especially considering that there is some kind of buffering in HDF. This one is already pretty big.

>My plan before was that you would supply an array of train IDs along with your data to be written, and if the pattern of train IDs conflicts with what's already in the index, you'd get an error. That's not particularly elegant, but...

Why not?... Better to issue an error early. It is going to be a fail if we have to resolve issues with broken files, especially if they are broken via errors in the user's script.

But I would emphasize, that if we will write arrays immediately then we broke the invariant that we have a correct portion of a file at least in the buffers. It means that if a user doesn't provide data for the rest datasets, then we have no option to finalize the writing in the correct state. What we will do in this case? Will we remove the file after hours of writing? 

I'd better hold on arrays provided for writing until the user fills all datasets in the source for the given number of trains.

takluyver:
> C'mon, what extra layer of complexity? Just a small piece of code is located in FileWriterMeta.__new__() instead of FileWriterBase.__init__()...

The extra layer of complexity that needs a metaclass, an `Options` class, descriptor objects, dynamically created class attributes, and so on. If you want to argue this is worth it because of how nice the code using it can look, that's fine, but you can't claim there's no extra complexity in making a fancy API.

> I bet you'll love it too when it stops being unusual for you ;)

I bet I won't. ;-) I've worked with enough APIs by now that I have a pretty good idea which bits of magic I like and which I don't.

> Unfortunately, I don't see a nice way to make the plain version basic.

I haven't worked this out in detail, but as @philsmt pointed out to me, a class definition is really just a dict literal with some magic. I believe it should be possible to make a plain API which works with dicts and/or keyword arguments, and make the metaclass/desciptors part a wrapper around that.

> But I would emphasize, that if we will write arrays immediately then we brake the invariant that we have a correct portion of a file at least in the buffers. It means that if a user doesn't provide data for the rest datasets, then we have no option to finalize the writing in the correct state. What we will do in this case? Will we remove the file after hours of writing?

I wouldn't delete the file - not least because it might be useful for investigating what went wrong. But if there's an error while writing a file, that file is not guaranteed to be valid. That seems like a reasonable compromise (and I imagine it's really hard to actually *guarantee* that a partly written file will always be valid, in any case).

egorsobolev:
> The extra layer of complexity that needs a metaclass, an `Options` class, descriptor objects, dynamically created class attributes, and so on.

But if you don't create these additional classes (descriptors, `Options` class) you will spread code from them into the main class. The simplification only in a sense of fewer classes. But I don't see advantages in this, I prefer to encapsulate the code. It adds modularity, makes the main code cleaner and everyone can easily add new descriptors with new behavior without touching the main class (e.g. detector descriptor, which works with multiple datasets). It is rather related to programming style than the interface kind.

The same with dynamically created class attributes. You anyway will create them in an instance using the same code. And also dynamically (this is the curse of python). Don't see any simplifications or advantages.

> I haven't worked this out in detail, but as @philsmt pointed out to me, a class definition is really just a dict literal with some magic. I believe it should be possible 

I worked out on it, there is no magic. Actually the metaclass "magic" adds some touch at the very end, and this is the only missed thing to make `FileWriterBase` functional. If you don't want to have metaclass you have to move the metaclass "magic" in `__init__()`. It leads to a single possible solution. We have to fork hierarchy after `FileWriterBase` into two parallel children branches: the first is already implemented, the second will implement this relocation of "magic".
```python
class PlainFileWriter(FileWriterBase):
    def __init__(self, fn, sources, **kwargs):
        self._meta = Options(type('Meta', (object,), kwargs)) # we can enable Options to take a kwargs dict directly

        self.datasets = {}
        self.dataset_names = {}
        source_names = {}
        for source_name, datasets in sources.items():
            for ds_key, ds_attr in datasets.items():
                if ds.name in datasets:
                    warnings.warn(
                        f"The dataset name '{ds.name}' is duplicated. "
                                "Only the last entry is actual.", RuntimeWarning)

                ds = ds_attr.get_dataset(source_name, ds_key)
                ds.chunks_autosize(self._meta.max_train_per_file)

                self.datasets[ds.name] = ds
                self.dataset_names[ds.canonical_name] = ds.name
                source_names.setdefault(ds.source_name, ds.stype)

        self.list_of_sources = list(
                    (Source.SECTION[src_type], src_name)
                    for src_name, src_type in source_names.items()
        )
        super().__init__(fn)
```
Thus, the APIs are not the same semantically anymore (`PlainFileWriter` and `FancyFileWriter`). There is no way to use one and then switch to another in the same or future code.

And we duplicate the code in metaclass and `PlainFileWriter.__init__()`. It is different in details but does a similar thing. We will have to develop both in parallel. Luckily it's not very big. But you cannot say that is a nice way.

I also don't see any advantages in support. It is hard to imagine that "you will see better what's going on without the extra layer of complexity" if all the difference in 20 lines. Where is another layer of complexity? I really don't see it.

Let's go with it if you like.

> But if there's an error while writing a file, that file is not guaranteed to be valid. 

Okay. We shouldn't even try to implement writing by filled parts?

philsmt:
I'm sorry for coming late to this discussion. Let me try add a few thoughts.

First, one of the general design points you've talked about, which is important to me:

> It seems to me that the writing by trains corresponds to the vast majority of data processing that I can imagine. In fact, I wrote this implementation, because I needed to write azimuthal profiles in EuXFEL format. And of course, my processing algorithm is organized exactly by trains, mainly because EXtra-data naturally forces usage of such iterations.

I see that you've agreed already to adapt it for writing multiple trains at once, but I'd just like to double-down on this. Quite likely it's a question of community and scientific cases, and I understand the context of AGIPD, but in most of my scientific use cases, I want to turn raw data into large(ish) reduced arrays in memory and do exploratory filtering. And in special cases like REMI, where the reconstruction reduced the data by a factor ~1000x, I can (and need to) have hours of data in memory and work with them - including writing it to disk. So my (!) primary requirement for such an API is to write a large array `(num_trains, ...)` or even `(num_pulses, ...)` (with accompanying offset/count tables) to disk in EuXFEL-format without caring for manually laying out all the necessary meta-structures.

Now, concerning the API style, I find using classes to define the structure visually quite appealing. The only thing I really dislike is the `Meta` class - it's a weird magic name to define some metadata in a weird place. Given these are probably optional to define, how about moving them to inheritance parameters:
```python
class MyFileWriter(FileWriter, max_train_per_file=10, break_into_sequence=True):
    nb = DS(ctrl_grp, 'param.numberOfBins', (), np.uint64)
    v = DS(inst_grp, 'azimuthal.profile', (1000,), float)
```
That way, it is a bit more similar to specifying additional arguments of `DS`.

The concern I have is that this API style is pretty in a specific notebook that wants to run processing and add a new kind of data (like your example running within Project Launchpad), but cumbersome in abstract processing, where you handle unknown datasets. Now, one might argue that such code is more sophisticated anyway, but I'd prefer to have a more generic (and thus, less fancy) way around. I apoplogize if this comes from not fully grasping your prior discussion and code yet, but shouldn't it be easy to have a basic layer like this:
```python
file_desc = {
    'my_source': DS(...),
    ...
}

with FileWriter(file_desc, new_file) as wr:
    ...
```
...and layer the pretty API on top of that. This is what I was referring to in my earlier comment to Thomas ("classes are just syntactical sugar around dicts"). I realize that is a bit of an oversimplification, but you're not _really_ using them for much more here. Naturally, the question to layer which API on which might be somewhat academical, it's the maintainability and testability that should decide. What does matter is to use using either API clear to the user. I wonder if we could merge the `PlainFileWriter` and `FileWriterBase` somewhat into a single class `FileWriter` (i.e. the classical API) and then use a more declarative term for the pretty API which includes the metaclass, say something like `FileFormat`, `DataDescription`. I'd have to check, but since we're using metaclass magic anyway, it should not be a problem to have a complexer `__init__` with the property-style definitions.

egorsobolev:
I added buffering. It seems helping:
```shell
EXtra-data/timing % python buffered_write.py
Medium data: 100 trains by 1000 frames of vectors [float: 1000]
 - buffered: nrep=5, mean=6.40 s, std=0.368 s
 - direct:   nrep=5, mean=6.39 s, std=0.206 s
Small data: 10000 trains by 1 frame of 10 scalars [uint64]
 - buffered: nrep=15, mean=1.08 s, std=0.0391 s
 - direct:   nrep=3,  mean=20.62 s, std=0.628 s
```
Does it close the question on writing scalars and other small data by train?

JamesWrigley:
I have some small comments on the API from using it recently:
* It would be nice for the API to support writing multiple trains at once. I think it's not critical, though I suspect that if we don't, people will write their own.
* Having to use `DatasetDescr` for use with `FileWriter.open()` but `Dataset` for the class-style definition is confusing, could it not use `Dataset` for both?
* To me it's quite unintuitive that the keys in the sources dict for `FileWriter.open()` are the dataset key and the `name` argument for `DatasetDescr` is the attribute name, because the attribute name doesn't really describe the `Dataset`. It does make it easier to tell at a glance what datasets each group has, but I think it would make more sense if those were swapped so the sources dict would be structured as:
```python
sources = {
    ctrl_grp: {
        'nb': DatasetDescr('param.numberOfBins', (), np.uint64),
    },
    inst_grp: {
        'v': DatasetDescr('azimuthal.profile', (1000,), float)
    }
}

```
* I think `Dataset` should handle the `str` type so that the users don't have to explicitly pass `h5py.string_dtype()` as the dtype if they want to create string datasets.

takluyver:
> I've added buffering. It seems to help... Does it close the question on writing small data by train?

It helps, but with your added 'h5py' example, it looks like writing an array directly is still significantly faster than working train-by-train, even with the buffering. This is going to be especially the case if the data you want to write is in arrays to start with: you don't want to iterate over small pieces of the array for the writing machinery to copy into an internal buffer.

I'm not going to insist on this, because I can see how it's a bad fit with the design of automatically starting a new sequence file, but it still seems desirable to me to have a way to directly write an array for >1 train at once.

> I think Dataset should handle the str type so that the users don't have to explicitly pass h5py.string_dtype() as the dtype if they want to create string datasets.

I might suggest making it very mildly inconvenient to write variable-length strings, because I've come to realise how comparatively rough they are both in HDF5 and h5py. HDF5 stores every vlen string individually, and basically makes an array of pointers to them, while h5py constructs a Python object for each individual string on read. You give up a lot of the efficiency you have when working with fixed size data types.

egorsobolev:
Sorry, I didn't mean that I want to avoid implementing the multi-train writing. It's already here:
```python
import numpy as np
from extra_data.writer2 import FileWriter, DS, trs_

ctrl_grp = 'MID_DET_AGIPD1M-1/x/y'
inst_grp = 'MID_DET_AGIPD1M-1/x/y:output'

class MyFileWriter(FileWriter):
    nb = DS(ctrl_grp, 'param.numberOfBins', (), np.uint64)
    v = DS(inst_grp, 'azimuthal.profile', (1000,), float)

    class Meta:
        max_train_per_file = 40
        break_into_sequence = True

trains = range(10001, 10101)
with MyFileWriter('test_{seq:02d}.h5') as wr:
    # add data (source name interface)
    wr.add_trains(trains[:99], [0] * 99)

    wr.v = trs_([3]*99, np.random.rand(99*3, 1000))
    wr.add_value([1]*33, 'nb', [1000]*33)
    wr.add_value([1]*33, 'nb', [1000]*33)
    wr.add_value([1]*33, 'nb', [1000]*33)
```
Example of writing by trains:
```python
trains = range(10001, 10101)
with MyFileWriter('test_by_train_{seq:02d}.h5') as wr:
    for t in trains:
        # add train
        wr.add_train(t, 0)
        # add data
        wr.add_train_data(v = np.random.rand(3, 1000))
        wr.nb=1000
```
Unfortunately, writing by train works even slower for small data:
```shell
EXtra-data/timing% python buffered_write.py
Medium data: 100 trains by 1000 frames of 1 vector [float: 1000]
 - h5py:     nrep=5, mean=2.12 s, std=0.0339 s
 - buffered: nrep=5, mean=5.85 s, std=0.0349 s
 - direct:   nrep=5, mean=5.94 s, std=0.247 s
 - at once:  nrep=5, mean=5.90 s, std=0.107 s
Small data: 10000 trains by 1 frame of 10 scalars [uint64]
 - h5py:     nrep=25, mean=0.05 s, std=0.0126 s
 - buffered: nrep=15, mean=6.18 s, std=0.266 s
 - direct:   nrep=3,  mean=26.67 s, std=0.161 s
 - at once:  nrep=25,  mean=0.07 s, std=0.0148 s
```
Other changes:

- a train must be added to the file before passing data, also in the case of writing by trains
- do not need to finalize every train
- CONTROL data aren't copied to the next train
- removed interface with addressing datasets by pair (source, key) because this involves the details of file format into the implementation. The internal (attribute) names allow isolating these details in `Dataset` descriptors which can be easily overridden.

egorsobolev:
Thus, two major issues resolved:
- [x] buffering
- [x] writing multi-train data

What do you think about the new implementation?

egorsobolev:
Thanks, @philsmt, @JamesWrigley

You both mentioned writing of multi-train data. I added this yesterday, please look post above.

Today I polished the plain interface. I agree with @JamesWrigley and have swapped the dataset names and keys. I also removed an additional descriptor which was introduced specially for this inverted position of these two parameters. I hope @takluyver  also agrees with this change.

Moreover, I added aliases for source names. Now one doesn't need to break encapsulation and use external variables to reuse long names. See examples:

Class interface:
```python
class MyFileWriter(FileWriter):
    nb = DS('@ctrl', 'param.numberOfBins', (), np.uint64)
    rlim = DS('MID_DET_AGIPD1M-1/x/y', 'param.radiusRange', (2,), float)
    tid = DS('@inst', 'azimuthal.trainId', (), np.uint64)
    v = DS('@inst', 'azimuthal.profile', (1000,), float)

    class Meta:
        aliases = {
            'ctrl': 'MID_DET_AGIPD1M-1/x/y',
            'inst': 'MID_DET_AGIPD1M-1/x/y:output',
        }

with FileWriter('filename-{seq:02d}.h5') as wr:
    ...
```

Plain interface with one-level `dict`:
```python
datasets = {
    'nb': DS('@ctrl', 'param.numberOfBins', (), np.uint64),
    'rlim': DS('MID_DET_AGIPD1M-1/x/y', 'param.radiusRange', (2,), float),
    'tid': DS('MID_DET_AGIPD1M-1/x/y:output', 'azimuthal.trainId', (), np.uint64),
    'v': DS('@inst', 'azimuthal.profile', (1000,), float),
}
aliases = {
    'ctrl': 'MID_DET_AGIPD1M-1/x/y',
    'inst': 'MID_DET_AGIPD1M-1/x/y:output',
}
with FileWriter.open('filename-{seq:02d}.h5', datasets, aliases=aliases, buffering=True) as wr:
    ...
```
Plain interface with two-level `dict`:
```python
datasets = {
    'MID_DET_AGIPD1M-1/x/y': {
        'nb': DS('', 'param.numberOfBins', (), np.uint64),
        'rlim': DS('', 'param.radiusRange', (2,), float),
    },
    'MID_DET_AGIPD1M-1/x/y:output': {
        'tid': DS('', 'azimuthal.trainId', (), np.uint64),
        'v': DS('', 'azimuthal.profile', (1000,), float),
    }
}
aliases = {
    'ctrl': 'MID_DET_AGIPD1M-1/x/y',
    'inst': 'MID_DET_AGIPD1M-1/x/y:output',
}
with FileWriter.open('filename-{seq:02d}.h5', datasets, buffering=True) as wr:
    ...
```
If the first parameter is not an empty string it will be added to the source name over the slash. 
If the first parameter starts with '@' then it will be resolved first then added.
One-level and two-level styles may be mixed.

@philsmt, I checked that it is really possible to add keyword parameters to the list of base classes. But take a look at the examples, there are already about four long parameters in this `Meta` and now I added the `dict`. This `class Meta` attribute (the name may be different) was specially added so as not to overload other places and complicate the reading of the code, but at the same time follow the principle of encapsulation. Do you anyway insist to add this option?

egorsobolev:
Remained minor suggestions:
- mix class parameters and parameters from `class Meta` attribute (I would prefer do not to provide this, see remarks above)
- allow `str` as `dtype` parameter in Datasets. (Thomas suggest to don't simplify usage of variable-length strings)

Please remind me if I forgot anything. And comment on these in the list.

There is also a discussion about what interface must be first. I don't have a better solution for now. I would probably comment on it later.

egorsobolev:
I've added run-time parameterization:
```python
from extra_data.writer2 import FileWriter, DS, AdjVecDS, AdjDS

class AzFileWriter(FileWriter):
    gv = AdjDS('@prm', 'geom.fragmentVectors', ('nmod', 'nfrag', 3, 3), float)
    gs = AdjDS('@prm', 'geom.fragmentSize', ('nmod', 'nfrag', 2), int)
    nb = DS('@prm', 'param.numberOfBins', (), np.uint64)
    rlim = DS('@prm', 'param.radiusRange', (2,), float)

    n = AdjVecDS('@res', 'azimuthal.radial', 'nbin', float)
    v = AdjVecDS('@res', 'azimuthal.intensity', 'nbin', float)
    class Meta:
        aliases = {
            'prm': "{det_name}/x/y"
            'res': "{det_name}/x/y:output"
        }

def calc(det_name, nmod, nfrag, nbin):
    ...
    prm = dict(det_name=det_name, nmod=nmod, nfrag=nfrag, nbin=nbin)
    with AzFileWriter('filename-{seq:02d}.h5', **prm) as wr:
        ...
```
As you can see, now both source name and dataset shape may be given as parameters. Thus, there is no problem adding some abstraction and still have a compact nice declarative style. Going this way you even can put the definition of the writer in a separate module, keeping your computational code absolutely clear.

If one needs more abstraction he may create his own dataset descriptor, which is designed to control the entire machinery (not fully finished at the moment).

The plain interface is also available. Still implemented in standard Python way by using type() to create a class in a functional way.

egorsobolev:
I optimized this a little bit and added profiling. See new benchmarks:
```shell
EXtra-data/timing% python3 bench_writer2.py
Data(*)   |Variant   |Nrep  |mean(T), s  |std(T), s   |T_bench, s  |T_wr2, s    |T_wr2_h5, s |T_pure_h5,s
----------+----------+------+------------+------------+------------+------------+------------+------------
scalar    |hdflib    |    30|        0.03|      0.0129|     0.00215|           0|           0|      0.0287
scalar    |buffer    |     9|        3.34|      0.0579|       0.053|        3.23|      0.0538|           0
scalar    |direct    |     3|       27.85|      0.0372|      0.0664|         5.4|        22.4|           0
scalar    |entire    |    30|        0.08|      0.0166|    0.000104|      0.0331|      0.0428|           0
vector    |hdflib    |    30|        0.60|      0.0299|    0.000903|           0|           0|       0.594
vector    |buffer    |    30|        0.65|      0.0248|    0.000602|      0.0289|       0.624|           0
vector    |direct    |     3|       13.95|         1.9|     0.00101|      0.0303|        13.9|           0
vector    |entire    |    30|        0.64|      0.0494|    3.94e-05|      0.0023|       0.639|           0

(*) scalar - small data: 10000 trains by 1 frame of 10 scalars [uint64]
(*) vector - medium data: 100 trains by 1000 frames of 1 vector [float: 1000]
```
The scripts also generate detailed profiler reports.

Although writing of scalars by trains still better to avoid, writing by multi-train arrays looks good. Bigger data looks good even in writing by trains.

egorsobolev:
Thanks all for the discussion and suggestions! I have finished.

Features:
- writing multi-train arrays at once
- writing data by trains
- check of data consistency
- buffering (direct writing is also an option)
- automatic sequence splitting
- aliases for source name
- parameterization of datasets and sources
- constructor in plain style

Look at previous posts for details. There is also [an architecture description](../blob/writer2-alt/docs/architecture_writer2.rst) and benchmarks with profiler reports.

@takluyver and @philsmt, could you please review?

takluyver:
After some discussion, Philipp and I are suggesting that this should be developed as a separate EXtra-writer package, rather than merging it into EXtra-data.

This is a substantial amount of code and API, for a more specialised use case, and it (quite reasonably) uses nothing from the existing EXtra-data code. So making a separate package seems the natural and obvious way to develop it. That also means it can have its own release schedule and its own documentation, with no concerns about how it fits in with the reading APIs in EXtra-data.

egorsobolev:
This has been split into a separate package https://github.com/European-XFEL/EXtra-writer.

PR #183, opened 2021-05-11
Speed up _check_source_conflicts()

takluyver:
Opening proposal 2926, run 74, with around 21000 files, was taking a very long time (long enough that I didn't wait for it to finish; the [report on Redmine](https://in.xfel.eu/redmine/issues/96112) mentions hours).

Using the poor man's sampling profiler (hitting Ctrl-C and looking at the traceback) I found that `_check_source_conflicts` was the culprit. This checks whether any source has the same train ID in different files. It was going through every source, and building up a set of train IDs seen so far, comparing each new file to that. This got slow with large numbers of files. This PR makes two changes:

- Rather than checking each new file and adding it to an array of train IDs, make one big array with train IDs from all files, and check it for duplicates. This avoids doing repeated numpy operations and reallocating an array.
- As one set of files typically contains several sources, cache the result on a frozenset of file paths, and reuse it for the next source related to the same set of files.

With this branch, I can open the run in 10-15 seconds.

```shell
time lsxfel /gpfs/exfel/exp/SQS/202121/p002926/raw/r0074
```

```python
%%time
extra_data.open_run(2926, 74)
```

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -964,15 +964,21 @@ class DataCollection:
         """Check for data with the same source and train ID in different files.
         """
         sources_with_conflicts = set()
+        files_conflict_cache = {}
+
+        def files_have_conflict(files):
+            fset = frozenset({f.filename for f in files})
+            if fset not in files_conflict_cache:
+                if self.inc_suspect_trains:
+                    tids = np.concatenate([f.train_ids for f in files])
+                else:
+                    tids = np.concatenate([f.valid_train_ids for f in files])
+                files_conflict_cache[fset] = len(np.unique(tids)) != len(tids)
+            return files_conflict_cache[fset]
+
         for source, files in self._source_index.items():
-            got_tids = np.array([], dtype=np.uint64)
-            for file in files:
-                if file.has_train_ids(got_tids, self.inc_suspect_trains):
-                    sources_with_conflicts.add(source)
-                    break
-                f_tids = file.train_ids if self.inc_suspect_trains \
-                            else file.valid_train_ids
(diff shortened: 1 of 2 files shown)


takluyver:
The test failures are related to the handling of bad train IDs. One of the sample files we make to test with has a duplicated train ID. Previously, `_check_source_conflicts` would only spot repeated train IDs *between* files, not *within* them. The new implementation complains about repeated train IDs in the same file.

We could recreate the previous behaviour exactly, but I'm inclined to say that a duplicate train ID inside a file is problematic enough (and hopefully rare enough) that it should fail the check. You would still be able to open it by passing `inc_suspect_trains=False`, so long as one of them is flagged as invalid.

takluyver:
For now, I've updated the tests to match the new behaviour (failing on duplicated train IDs unless `inc_suspect_trains=False`). We can knock that commit off if we choose to go the other way.

tmichela (approved):
> We could recreate the previous behaviour exactly, but I'm inclined to say that a duplicate train ID inside a file is problematic enough (and hopefully rare enough) that it should fail the check. You would still be able to open it by passing inc_suspect_trains=False, so long as one of them is flagged as invalid.

Let's hope we don't end up with the same kind of issue #178 fixed :upside_down_face: 
But I don't recall ever seeing this, so... :crossed_fingers: LGTM

tmichela:
Just for the sake of it, I checked an few 100 runs  I know with trainId issues, and did not any duplicate Train ID in the same file

takluyver:
Thank-you! :grinning: :+1:

PR #184, opened 2021-05-18
Split DataCollection & KeyData into chunks by train

takluyver:
Currently, we make it easy to get all of the data for a given source & key in one go, or to work train by train. But you may want an intermediate position, where you work on chunks of, say, 10 trains at a time. This can be more efficient than reading individual trains, but use less memory than loading data from a whole run. It can also be useful for parallel processing.

Currently, calling code has to decide where to cut the data and use `.select_trains()` to get subsets. This adds a `.split_trains()` method which does that for you, given a number of parts to make and/or a maximum number of trains per part. If both parameters are given, it makes *at least* X parts so that there are *at most* Y trains per part.

### DIFF
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -116,6 +116,8 @@ class KeyData:
         This returns an iterator yielding new :class:`KeyData` objects.
         The parts will have similar sizes, e.g. splitting 11 trains
         with ``trains_per_part=8`` will produce 5 & 6 trains, not 8 & 3.
+        Selected trains count even if they are missing data, so different
+        keys from the same run can be split into matching chunks.
 
         Parameters
         ----------
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -113,7 +113,8 @@ class KeyData:
 
         Either *parts* or *trains_per_part* must be specified.
 
-        The returned parts will have similar sizes, e.g. splitting 11 trains
+        This returns an iterator yielding new :class:`KeyData` objects.
+        The parts will have similar sizes, e.g. splitting 11 trains
         with ``trains_per_part=8`` will produce 5 & 6 trains, not 8 & 3.
 
         Parameters
@@ -124,8 +125,8 @@ class KeyData:
             specified, this is a minimum, and it may make more parts.
             It may also make fewer if there are fewer trains in the data.
         trains_per_part: int
-            A maximum number of trains in each part. Typically the parts will
-            have fewer trains than this.
+            A maximum number of trains in each part. Parts will often have
+            fewer trains than this.
         """
         for s in split_trains(len(self.train_ids), parts, trains_per_part):
             yield self.select_trains(s)
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -4,7 +4,9 @@ import numpy as np
 
 from .exceptions import TrainIDError
 from .file_access import FileAccess
-from .read_machinery import contiguous_regions, DataChunk, select_train_ids
+from .read_machinery import (
+    contiguous_regions, DataChunk, select_train_ids, split_trains,
+)
 
 class KeyData:
     """Data for one key in one source
@@ -105,6 +107,28 @@ class KeyData:
 
     def __getitem__(self, item):
         return self.select_trains(item)
+
+    def split_trains(self, parts=None, trains_per_part=None):
+        """Split this data into chunks with a fraction of the trains each.
+
+        Either *parts* or *trains_per_part* must be specified.
+
+        The returned parts will have similar sizes, e.g. splitting 11 trains
+        with ``trains_per_part=8`` will produce 5 & 6 trains, not 8 & 3.
+
+        Parameters
+        ----------
+
+        parts: int
+            How many parts to split the data into. If trains_per_part is also
+            specified, this is a minimum, and it may make more parts.
+            It may also make fewer if there are fewer trains in the data.
+        trains_per_part: int
+            A maximum number of trains in each part. Typically the parts will
+            have fewer trains than this.
+        """
+        for s in split_trains(len(self.train_ids), parts, trains_per_part):
+            yield self.select_trains(s)
 
     def data_counts(self):
         """Get a count of data entries in each train.
--- a/docs/reading_files.rst
+++ b/docs/reading_files.rst
@@ -17,8 +17,6 @@ run number to look up the standard data paths on the Maxwell cluster.
 
 .. autofunction:: open_run
 
-   .. versionadded:: 0.5
-
 You can also open a single file. The methods described below all work for either
 a run or a single file.
 
@@ -162,8 +160,6 @@ methods offer extra capabilities.
           Examples using xarray & pandas with EuXFEL data
 
    .. automethod:: get_virtual_dataset
-
-      .. versionadded:: 0.5
 
       .. seealso::
         :doc:`parallel_example`
--- a/docs/reading_files.rst
+++ b/docs/reading_files.rst
@@ -133,6 +133,10 @@ below, e.g.::
 
    .. automethod:: select_trains
 
+   .. automethod:: split_trains
+
+      .. versionadded:: 1.7
+
 The run or file object (a :class:`DataCollection`) also has methods to load
 data by sources and keys. :meth:`get_array`, :meth:`get_dask_array` and
 :meth:`get_series` are directly equivalent to the options above, but other
@@ -230,6 +234,10 @@ data, so you use them like this::
    .. automethod:: deselect
 
    .. automethod:: select_trains
+
+   .. automethod:: split_trains
+
+      .. versionadded:: 1.7
 
    .. automethod:: union
 
--- a/extra_data/read_machinery.py
+++ b/extra_data/read_machinery.py
@@ -128,10 +128,12 @@ def select_train_ids(train_ids, sel):
 
 def split_trains(n_trains, parts=None, trains_per_part=None) -> [slice]:
     if trains_per_part is not None:
+        assert trains_per_part >= 1
         n_parts = math.ceil(n_trains / trains_per_part)
         if parts is not None:
             n_parts = max(n_parts, min(parts, n_trains))
     elif parts is not None:
+        assert parts >= 1
         n_parts = min(parts, n_trains)
     else:
         raise ValueError("Either parts or trains_per_part must be specified")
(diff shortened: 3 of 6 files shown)


egorsobolev:
Oh, nice ;)

I would remove the `parts` parameter. This provokes unsafe programming. One needs to know the run length to use it. It is possible only if a user writes a script to process a certain run. Then it is unsafe to apply it for other runs. Or he will compute the number of parts at runtime reading the run length first. Then let him better compute `trains_per_part`.

Is it possible to split by frames? It is better in the case if the number of frames varies between trains.

egorsobolev:
Aha, I cannot get chunks with the requested number of trains. It always adjusts this number. I'm used to having this option. And why do you restrict it, in principle? 

What do you think about splitting exactly as was requested:
```python
[ slice(i, min(i + trains_per_chunk, n_trains)) for i in range(0, n_trains, trains_per_chunk) ]
```
and all balancing computations provide in separate functions?

extra_data/keydata.py
```diff
@@ -106,6 +108,29 @@ def select_trains(self, trains):
...
+
+        parts: int
+            How many parts to split the data into. If trains_per_part is also
+            specified, this is a minimum, and it may make more parts.
+            It may also make fewer if there are fewer trains in the data.
+        trains_per_part: int
+            A maximum number of trains in each part. Parts will often have
+            fewer trains than this.
+        """
+        for s in split_trains(len(self.train_ids), parts, trains_per_part):
```
tmichela, line 131: I'd probably select only train IDs containing data for `KeyData` before splitting in parts, and maybe a `require_all` argument for the same purpose on `DataCollection`?

takluyver:
> I would remove the parts parameter. This provokes unsafe programming. One needs to know the run length to use it.

I assume you mean because the parts could be so large that you can't load them into memory? The same issue applies in theory to all the methods like `.get_array()` which let you retrieve a particular source/key for an entire run. But there are plenty of sources and keys where it's pretty reasonable to assume that you a run's worth of data can be loaded with no problem, and it can be very useful for quick analysis to do so.

The obvious use case for splitting into N similar parts is for parallelisation - N might be the number of nodes or cores you can use, or some fraction of that. It doesn't necessarily mean you're going to load each chunk fully into memory - you could iterate through trains within each chunk, or make smaller chunks within each worker, if memory usage is a concern.

> Is it possible to split by frames? It is better in the case if the number of frames varies between trains.

It's not, at least with this PR. That's a much more complex thing to do in a general way, partly because the file format doesn't really say much about data with multiple entries per train (does entry 7 for this source correspond to entry 7 for that source?), and partly because EXtra-data's selection mechanisms deal with trains, not entries.

> What do you think about splitting exactly as was requested:

In the use cases where I've seen something like this, (in calibration & pasha), we've wanted balanced chunks rather than exact ones. This seems likely in general where you're breaking something up to be processed in multiple workers. And of course you can't make every chunk exact in general, because the last one may be smaller, so code has to be prepared to handle a chunk of < n trains either way.

It's also comparatively easy to make slices of 100 trains each and use them with `.select_trains()` if you do want that, as in your example code. You can even leave out the `min()`, because slicing beyond the end of the sequence is OK. There's a tiny bit more to think about in making balanced chunks, which makes the helper method a bit more useful.

extra_data/keydata.py
takluyver: Yes, I think that probably makes sense for KeyData at least.

For DataCollection, what do you think about leaving this to `.select()`? I think filtering for trains with data only makes sense when you define which data you're looking at (otherwise, some source which wasn't recording could mean no trains are included), so you'd want to select first anyway. It would look like this:

```python
for chunk in run.select('*XGM/*', require_all=True).split_trains(8):
    ...
```

tmichela: yup, I agree :+1:

tmichela:
> I would remove the `parts` parameter. This provokes unsafe programming. One needs to know the run length to use it. It is possible only if a user writes a script to process a certain run. Then it is unsafe to apply it for other runs. Or he will compute the number of parts at runtime reading the run length first. Then let him better compute `trains_per_part`.
> 
> Is it possible to split by frames? It is better in the case if the number of frames varies between trains.

It probably really depends on the use cases. For SFX data, splitting by number of trains makes sens because the memory is the limit.
I've personally have been more often in more situation where the memory available isn't the bottleneck, and in those cases I'd use the `parts` argument to distribute evenly the data to a pool of workers.

I also think a split per frame would be nice. But do we have yet data with variable number of event per train? I'd expect this to be the case once data reduction is in place (for example)

egorsobolev:
> But there are plenty of sources and keys where it's pretty reasonable to assume that you a run's worth of data can be loaded with no problem, and it can be very useful for quick analysis to do so.

I just suggested making `trains_per_part` mandatory. That is not a big deal to pass `trains_per_part=n_trains` and got splitting by parts. But the user will pay attention to a potential issue, I hope. I think it is good practice but I'm not going to insist.

> we've wanted balanced chunks rather than exact ones

Actually similar, `parts=1` enables the balancing without limiting of parts. Thus, `parts=None` may be dedicated to exact chunking. But you are right I can do it without a dedicated method. Thus, it is an optional suggestion on your taste.

Aside from these optional suggestions, I'm happy with the implementation ;) The question on splitting by frames is just to bump this topic.

philsmt:
Thanks for this PR. It makes sense to expose this as part of the regular interface.

I can vouch for the `split_trains` functions, it's the same implementation Thomas contributed to `pasha` and it's been working well for a while 😄

Concerning the API discussion you're having, I have to agree with Thomas and the way it is implemented right now. It primarily tackles the "complicated" cases of doing such a splitting operation when trying to fit into a predefined numbers of `parts`. I understand your point of rather having a fixed number of trains per part, but the algorithm as is does not generate additional `parts`, it just balances their contents. If you have this very specific concern, I think you're still well suited by just doing it yourself with a list comprehension of `select_trains()`s or so.

egorsobolev:
> Concerning the API discussion you're having, I have to agree with Thomas and the way it is implemented right now.

Thomas implementation LGTM. I suppose my suggestion may only add completeness to this but it's entirely up to Thomas.

takluyver:
Thanks all. :slightly_smiling_face: 

We could always extend this with an option like `balance=False`. But I'm not sure how often it would be useful. Either way, you have to ensure that either `trains_per_part` exactly divides the number of trains you have, or the code that gets a chunk is prepared for it having fewer trains.

I still need to implement @tmichela's suggestion to filter out trains with no data for `KeyData.split_trains()`.

philsmt:
I'm not 100% sold that we should exclude trains with no data for `KeyData` by default. I can imagine use cases where you look at a particular set of trains, and have optional sources that you take when they're there, but you can continue without (e.g. by interpolating). Having the option to get empty data would make such code much more elegant.

However, I'm not sure how exactly the semantics are currently for `KeyData`. Methods like `ndarray()` will definitely return only trains with data, but what does `trains()` do?

takluyver:
Interesting point. At present, `kd.trains()` will only yield something where there is data for a train, skipping over trains with no data entirely. To include trains with missing data, you'd do something like this:

```python
for tid in kd.train_ids:
    _, arr = kd.train_from_id(tid)
    if arr.size == 0:
        ... # interpolate?
```

Maybe if this goes in as is, we should have a separate method to select only trains with data for a KeyData object, similar to `dc.select(..., require_all=True)`?

takluyver:
And/or maybe `KeyData.trains()` should have an option like `inc_missing=True`?

philsmt:
Hmm, however way we go, we should probably make this behaviour more explicit. It seems like now:
* `DataCollection` and `KeyData` track trains with and without data
* All their data retrieval methods return only trains with data

...which makes this case so complicated, as it's not data retrieval 😁

You're probably right that it's more unlikely you'd want to track empty trains once you're at a `KeyData` object, but again, then we should make this probably more explicit.
How about `KeyData` always implies `require_all=True`, i.e. when you want to work with empty data, you have to go via `DataCollection`?

takluyver:
> How about KeyData always implies require_all=True, i.e. when you want to work with empty data, you have to go via DataCollection?

How far would it imply that, though? If we always remove trains with missing data from `KeyData.train_ids`, that could break existing code (e.g. things like the `for tid in kd.train_ids` snippet above). But if we include them in `.train_ids` and then exclude them for `.split_trains()`, that could be surprising (it looks like you have 100 trains, then when you split them up, you only have 50... or none at all).

:thinking:

philsmt:
It would've, but I didn't consider a user iterating over train IDs directly. I suppose it's the place we're in.

Then it might indeed be beneficial to have flags, possibly even for other data retrieval methods in the future (e.g. provide a fill value for `ndarray()`). Ultimately the library is supposed to make data analysis easier, and this is one of these annoying boilerplate things.

Which leaves the question of default behaviour. Based on this recent point of iterating train IDs (which is a public property), I'd include empty trains by default.

takluyver:
OK, so merge this as it is, and think separately about different ways to deal with missing data in KeyData?

@tmichela , is that OK with you?

tmichela:
>  I can imagine use cases where you look at a particular set of trains, and have optional sources that you take when they're there, but you can continue without 

But this use case would already be covered by using `DataCollection`, or labeled arrays.

In my opinion, this method only makes sens if we can filter empty trains first, as it is the only use case I encountered so far. I don't care that much if that's the default behavior or not, but considering that it is the behavior of `.trains()` I'd be inclined to have it as default.

(You can merge this already if you prefer)

takluyver:
In #193, I've added a `.drop_missing()` method (though the name is still under discussion) to go from a `KeyData` object which includes trains with no data, to one which only has trains with data. So with both PRs as they stand:

```python
# Behaviour Philipp wants:
run[src, key].split_trains(parts=16)

# Behaviour Thomas M wants:
run[src, key].drop_missing().split_trains(parts=16)
```

I accept this distinction is not super obvious, but one thing I like about not keeping those trains by default is that if you get two KeyData objects from the same run and use `.split_trains()` with the same options, you'll get matching chunks for the two parts of data, which you can zip together. Of course, you could do the same with a DataCollection, but I bet people will sometimes want to use a couple of KeyData objects instead.

philsmt:
I don't have strong feelings anymore about it. My only remaining argument is that, since `KeyData` exposes `train_ids` as a public property the same way `DataCollection` does, I'd find it counter-intuitive to have methods not return said trains, unless there's an obvious reason (i.e. for `.ndarray()`).
But I concede that for most practical purposes, you _probably_ want only those trains with data.

philsmt (approved):

takluyver:
I checked with @tmichela separately about this, and he was happy with the plan I outlined above.

Issue #187, opened 2021-05-27
Problem when trying to stream data


FilipeMaia:
Hi,

I was trying to stream data using the following command:
`karabo-bridge-serve-files /gpfs/exfel/exp/SPB/201901/p002316/raw/r0174 4545`

And I got the following error, which I don't quite know where it's coming from:
```
Streamer started on: tcp://max-display002.desy.de:4545
Traceback (most recent call last):
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1/bin//karabo-bridge-serve-files", line 8, in <module>
    sys.exit(main())
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1/lib/python3.7/site-packages/extra_data/export.py", line 190, in main
    use_infiniband=args.use_infiniband, sock=args.socket_type
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1/lib/python3.7/site-packages/extra_data/export.py", line 141, in serve_files
    for tid, data in _iter_trains(data, merge_detector=append_detector_modules):
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1/lib/python3.7/site-packages/extra_data/export.py", line 66, in _iter_trains
    for tid, train_data in data.trains():
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1/lib/python3.7/site-packages/extra_data/reader.py", line 1397, in __iter__
    yield tid, self._assemble_data(tid)
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1/lib/python3.7/site-packages/extra_data/reader.py", line 1372, in _assemble_data
    self._set_result(res, source, key, ds[pos])
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1/lib/python3.7/site-packages/h5py/_hl/dataset.py", line 780, in __getitem__
    selection = sel.select(self.shape, args, dataset=self)
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1/lib/python3.7/site-packages/h5py/_hl/selections.py", line 82, in select
    return selector.make_selection(args)
  File "h5py/_selector.pyx", line 272, in h5py._selector.Selector.make_selection
[...]

PR #188, opened 2021-05-27
Validate indexes for CONTROL data

takluyver:
Previously, we were only checking the indexes for INSTRUMENT data, not CONTROL. I think we may have assumed that CONTROL was unlikely to go wrong. On #187, we found a run where there is a problem - CONTROL data missing, INDEX still records it as though it was there - which `extra-data-validate` did not pick up.

This extends the same checks we do on INSTRUMENT data to CONTROL, with one difference. We only record the "Index referring to data outside dataset" problem once per source & per file, rather than for every key. When I recorded it for every key, I got a list of some 27000 problems with the run. Once per source should be plenty to point out the problem, and then you can use `h5glance` to find more detail if necessary.

```
Checking run directory: /gpfs/exfel/exp/SPB/201901/p002316/raw/r0174
Progress: 207/207 [##################################################]
84 problems in 7 files (last: RAW-R0174-DA02-S00002.h5)
Validation failed! 84 problems:

Index referring to data (11) outside dataset (0)
  dataset: CONTROL/SPB_IRU_AGIPD1M1/CTRL/MC1/module2/dBoardTemperature2/timestamp
  file: /gpfs/exfel/exp/SPB/201901/p002316/raw/r0174/RAW-R0174-DA02-S00006.h5

Index referring to data (11) outside dataset (0)
  dataset: CONTROL/SPB_IRU_AGIPD1M/TSENS/Q2_T_BLOCK/force/timestamp
  file: /gpfs/exfel/exp/SPB/201901/p002316/raw/r0174/RAW-R0174-DA02-S00006.h5

...
```

### DIFF
--- a/extra_data/validation.py
+++ b/extra_data/validation.py
@@ -101,6 +101,7 @@ class FileValidator:
                         ),
                         dataset=ds_path,
                     )
+                    break  # Recording every key separately can make a *lot* of errors
 
             self._check_index(f'/INDEX/{src}')
 
--- a/extra_data/validation.py
+++ b/extra_data/validation.py
@@ -72,13 +72,13 @@ class FileValidator:
                     dataset=ds_path,
                 )
 
-    def _get_index(self, source, group):
+    def _get_index(self, path):
         """returns first and count dataset for specified source.
         
         This is slightly different to the same method in FileAccess as it does
         cut the dataset up to the trainId's dataset length.
         """
-        ix_group = self.file.file['/INDEX/{}/{}'.format(source, group)]
+        ix_group = self.file.file[path]
         firsts = ix_group['first'][:]
         if 'count' in ix_group:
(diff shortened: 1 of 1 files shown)


tmichela:
I wonder if we should try to summarize similar problems to avoid returning N times the same output with just a different source name, ex:

```console
Index referring to data (11) outside dataset (0) for 82 sources, example:
  dataset: CONTROL/SPB_IRU_AGIPD1M1/CTRL/MC1/module2/dBoardTemperature2/timestamp
  file: /gpfs/exfel/exp/SPB/201901/p002316/raw/r0174/RAW-R0174-DA02-S00006.h5
```
and add a verbose argument.

But that's a different issue.

LGTM

tmichela (approved):

takluyver:
Thanks - I've opened #190 for summarising the issues.

PR #193, opened 2021-06-11
Add KeyData.drop_empty_trains() method to drop trains with no data

takluyver:
This is roughly equivalent to `.select(..., require_all=True)`, for a single source:

```python
beamview_dims = run['SA3_XTD10_IMGFEL/CAM/BEAMVIEW2:daqOutput', 'data.image.dims'].drop_missing()
```

The naming is roughly imitating pandas (`dropna`, `drop_duplicates`)

I've also added `.data_counts(labelled=False)` to find where there's data without having to load pandas.

### DIFF
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -89,6 +89,12 @@ class KeyData:
             run[source, key][:10]  # Select data for first 10 trains
         """
         tids = select_train_ids(self.train_ids, trains)
+        return self._only_tids(tids)
+
+    def __getitem__(self, item):
+        return self.select_trains(item)
+
+    def _only_tids(self, tids):
         files = [f for f in self.files
                  if f.has_train_ids(tids, self.inc_suspect_trains)]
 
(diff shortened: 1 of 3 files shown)


philsmt:
Thanks, that's useful!

I don't have strong feelings for it and I see the similarity to `pandas`, but just want to drop it in there for discussion: How about taking the _positive_ spin with something like `.require_trains()` or `.require_data()`?

takluyver:
Could do. The plus side of something like `.require_data()` is making the similarity to `.select(..., require_all=True)` more obvious. One possible downside is that it could imply 'require that there is some data', i.e. throw an error if there's no data available for this source. :shrug: 

I also have a minor concern about how this could affect performance. If you have a source which misses a lot of trains - e.g. it records data on every second train - then de-selecting the missing trains would currently mean that `.ndarray()` and similar methods would read data in many smaller pieces. We might need to make the chunking machinery smarter for that case.

takluyver:
@tmichela told me he favoured `drop_missing`:

> I think they both make sens in their own context (semantically).
> in the dataCollection case, we select a subset of data (sources), and to me it makes more sens to tell "take only trains if you have data for all sources you selected". Here with multiple source, "require_all" is clearer as it means you need all sources for a train to keep it.
> For the keydata, we effectively remove something from the selection. require_data feels weird because it does not indicate clearly what happens compared to drop_xxx. If you want the inverted logic, mabe something like keep_data, or trains_with_data, only_data, select_entries, ... all sound a bit off... :thinking:

I think I agree that `drop_missing` is clearer about what actually happens, though I could live with either.

extra_data/keydata.py
```diff
@@ -133,31 +142,26 @@ def split_trains(self, parts=None, trains_per_part=None):
...
-        train_ids = ser.index.intersection(np.asarray(self.train_ids))
-        return ser.loc[train_ids]
+        if labelled:
+            import pandas as pd
+            return pd.Series(counts, index=train_ids)
+        else:
+            # self.train_ids is always sorted. The train IDs from chunks
+            # should be in order, but sometimes trains are written out of order.
+            # Reorder the counts to match self.train_ids.
+            assert len(train_ids) == len(self.train_ids)
```
kakhahmed, line 161: Wouldn't this fail after dropping missing trains?

takluyver: `self._data_chunks` should have chunks contaiing only the selected trains. It wouldn't hurt to add a test for it, though.

takluyver: I added a test :slightly_smiling_face:

kakhahmed: Thank you

takluyver:
Further naming ideas from @philsmt: `drop_empty` or `drop_empty_trains` .

takluyver:
I've renamed the method to `drop_empty_trains`, which I found clearer.

PR #196, opened 2021-06-29
Remove special get_series() behaviour for big detector data

takluyver:
`run.get_series(src, key)` is mostly the same as `run[src, key].series()`. But the former had a special case for 1D fields in XTDF detector data, where it would load the pulse IDs and use them to make a multi-level index (along with train IDs). The latter just makes an index of train IDs, even if there are multiple values per train ID.

We suspected no-one was using the special case (there are only a couple of fields it could work for, and we haven't seen a use case), so we added a deprecation warning in version 1.4, released in February. We asked people to contact da-support if they were using it. As far as I know, no-one has got in touch. So this removes the special case, and makes `get_series()` use `KeyData.series()`.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -453,76 +453,11 @@ class DataCollection:
         return self._get_key_data(source, key).data_counts()
 
     def get_series(self, source, key):
-        """Return a pandas Series for a particular data field.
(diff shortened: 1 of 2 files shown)


tmichela:
LGTM

PR #197, opened 2021-06-29
Deprecate h5index and *_to_cbf functions

takluyver:
* `h5index` was basically a script I used when creating the mock data to check that its structure matched the real data. It worked, but it's very basic & limited and there are no tests for it. I'd rather people didn't notice that it's included in extra_data and then start relying on it.
* The `*_to_cbf` functions are the only part which depends on [FabIO](https://pypi.org/project/fabio/). `numpy_to_cbf` is trivial - you could just use the function from fabio directly. `hdf5_to_cbf` could conceivably be useful, but if you want to convert >1 image it's probably inefficient (reopening the HDF5 file for each one). I'd like to wait for @tmichela to approve deprecating those, since they've been in the code longer than I've been at EuXFEL.

### DIFF
--- a/extra_data/h5index.py
+++ b/extra_data/h5index.py
@@ -1,7 +1,13 @@
 import csv
 import h5py
 import sys
+import warnings
 
+warnings.warn(
+    "extra_data.h5index is deprecated and likely to be removed. "
+    "If you are using it, please contact da-support@xfel.eu.",
+    stacklevel=2,
+)
 
 def hdf5_datasets(grp):
     """Print CSV data of all datasets in an HDF5 file.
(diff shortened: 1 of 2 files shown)


tmichela:
you forgot to add the warning to [hdf5_to_cbf](https://github.com/European-XFEL/EXtra-data/pull/197/files#diff-d63b28ba6c30a77424f76e4f6785b0b32235f176e428b9c10821b091f626280eR145).

I'd probably be more generous and deprecate everything in `utils.py` (appart from `find_infiniband_ip` as it is used).

takluyver:
My thinking was that, as `hdf5_to_cbf` calls `numpy_to_cdf`, so you'll see the warning from there. It won't helpfully point to the place where you called hdf5_to_cbf, but hopefully that's easy to find.

I might deprecate at least `hdf5_paths` as well. The cbf functions are a particular target because they'd let us drop a dependency on Fabio.

PR #206, opened 2021-08-02
SourceData interface

### LINKED ISSUE #55: Objects for individual sources or datasets?

takluyver:
An object representing one source, with several keys:

```python
xgm = run['SPB_XTD9_XGM/DOOCS/MAIN']

'beamPosition.ixPos.value' in xgm  # Check for key
xgm_keys = xgm.keys()  # Find all (selected) keys
beam_x = xgm['beamPosition.ixPos.value']  # Get a KeyData object
```

Closes #55

### DIFF
--- a/extra_data/sourcedata.py
+++ b/extra_data/sourcedata.py
@@ -61,17 +61,25 @@ class SourceData:
             inc_suspect_trains=self.inc_suspect_trains,
         )
 
-    def keys(self):
+    def keys(self, inc_timestamps=True):
         """Get a set of key names for this source
 
         If you have used :meth:`select` to filter keys, only selected keys
         are returned.
+
+        For control sources, each Karabo property is stored in the file as two
+        keys, with '.value' and '.timestamp' suffixes. By default, these are
+        given separately. Pass ``inc_timestamps=False`` to ignore timestamps and
+        drop the '.value' suffix, giving names as used in Karabo.
 
         Only one file is used to find the keys. Within a run, all files should
         have the same keys for a given source, but if you use :meth:`union` to
         combine two runs where the source was configured differently, the
         result can be unpredictable.
         """
+        if not inc_timestamps:
+            return {k[:-6] for k in self.keys() if k.endswith('.value')}
+
         if self.sel_keys is not None:
             return self.sel_keys
 
--- a/extra_data/sourcedata.py
+++ b/extra_data/sourcedata.py
@@ -22,6 +22,10 @@ class SourceData:
         self.section = section
         self.inc_suspect_trains = inc_suspect_trains
 
+    def __repr__(self):
+        return f"<extra_data.SourceData source={self.source!r} " \
+               f"for {len(self.train_ids)} trains>"
+
     def _has_exact_key(self, key):
         if self.sel_keys is not None:
             return key in self.sel_keys
--- a/extra_data/sourcedata.py
+++ b/extra_data/sourcedata.py
@@ -1,5 +1,7 @@
 from typing import List
 
+import h5py
+
 from .exceptions import PropertyNameError
 from .file_access import FileAccess
 from .keydata import KeyData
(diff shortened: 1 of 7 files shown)


philsmt:
Thank you, loving this interface pattern. I'm definitely using `KeyData` much more than the classic interface already.

It's not strictly touch the code changed in this PR, but I feel the specific syntax here makes it even more natural to expect keys to work without the attached `.value`:

`'beamPosition.ixPos.value' in xgm  # Check for key`

I realize this remains debatable whether a Karabo viewpoint is valid on files. I actually meant to discuss at some point a general strategy on how to handle this conundrum, but it creeps up here again. I understand if you prefer to postpone it as part of a bigger problem, though.

takluyver:
Thanks; I think it makes sense to try to do something about `.value` here.

takluyver:
OK, both `key in sourcedata` and `sourcedata[key]` should now allow control keys without the `.value` suffix.

Since `run[source, key]` and various methods like `run.get_array()` now go through the sourcedata object, this should indirectly mean that you can leave off `.value` in a whole lot of places. We still treat the name with `.value` as canonical, though, for e.g. `sourcedata.keys()`.

tmichela:
Nice 👍 

can we make a quick mention in the documentation, maybe [here](https://extra-data.readthedocs.io/en/latest/reading_files.html#getting-data-by-source-key)?

takluyver:
Yes, of course, good point!

tmichela:
LGTM

extra_data/sourcedata.py
```diff
@@ -0,0 +1,81 @@
...
+            key,
+            train_ids=self.train_ids,
+            files=self.files,
+            section=self.section,
+            dtype=ds0.dtype,
+            eshape=ds0.shape[1:],
+            inc_suspect_trains=self.inc_suspect_trains,
+        )
+
+    def keys(self):
```
philsmt, line 64: I'm sorry striking the same notch (this saying probably doesn't work in English), but what do you think about adding an optional flag here to collapse `.value` and `.timestamp`? Something like `omit_suffix` or `collapse_value_and_timestamp`.

While it might only be useful in generic programming, I could imagine it something people want to do with the output from this function, and it might be convenient to provide it.

takluyver: That seems reasonable enough. What about something that defaults to True for including both as separate keys, like `inc_timestamps=False` or `value_and_timestamp=False`?

philsmt: Sure, either way is fine with me!

takluyver:
I've implemented it with `inc_timestamps=False` as the option to hide the timestamp/value suffixes. I'm not 100% convinced on that name if anyone has a better idea, though. It's ignored for instrument sources.

I've also used this option when listing keys in lsxfel, so you no longer see separate timestamp & value listed there.

philsmt:
Thanks, LGTM!

PR #208, opened 2021-08-03
keep run_metadata if its a union of a single run

kakhahmed:
This PR is related to https://git.xfel.eu/gitlab/detectors/pycalibration/merge_requests/535

Creating a union out of Different `H5File`  `DataCollections` of the same run, acts as if it is not a single run anymore.

As I wanted to create a Union of certain files without loading all run files and then selecting specific sources. I created these changes to keep run_metadata of a union as long as it is of the same run.

@takluyver

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -593,17 +593,27 @@ class DataCollection:
         files = set(self.files)
         train_ids = set(self.train_ids)
 
+        dc_run_num = self.run_metadata()["runNumber"] if self.is_single_run else None  # noqa
+
         for other in others:
             files.update(other.files)
             train_ids.update(other.train_ids)
+            # Set self.is_single_run to False
+            # in case of Unioning multiple run DataCollections.
 
+            if not (
+                other.is_single_run and
+                other.run_metadata()["runNumber"] == dc_run_num
+            ):
+                self.is_single_run = False
         train_ids = sorted(train_ids)
-        selection = union_selections([self.selection] +
-                                     [o.selection for o in others])
+        selection = union_selections(
+            [self.selection] + [o.selection for o in others])
 
         return DataCollection(
             files, selection=selection, train_ids=train_ids,
             inc_suspect_trains=self.inc_suspect_trains,
+            is_single_run = self.is_single_run,
         )
 
     def _expand_selection(self, selection):
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -592,17 +592,24 @@ class DataCollection:
         """
         files = set(self.files)
         train_ids = set(self.train_ids)
+        is_single_run = self.is_single_run
+        run_num = None
+        proposal_num = None
 
-        run_num = self.run_metadata()[
-            "runNumber"] if self.is_single_run else None
-
-        proposal_num = self.run_metadata()[
-            "proposalNumber"] if self.is_single_run else None
+        # Account for old runs (file format 0.5),
+        # which doesn't have these information.
+        if (
+            "runNumber" in self.run_metadata().keys() and
+            "proposalNumber" in self.run_metadata().keys()
+        ):
+            if self.is_single_run:
+                run_num = self.run_metadata()["runNumber"]
+                proposal_num = self.run_metadata()["proposalNumber"]
 
         for other in others:
             files.update(other.files)
             train_ids.update(other.train_ids)
-            # Set self.is_single_run to False
+            # Set other.is_single_run to False
             # in case of Unioning multiple run DataCollections.
 
             if not (
@@ -610,7 +617,7 @@ class DataCollection:
                 other.run_metadata()["runNumber"] == run_num and
                 other.run_metadata()["proposalNumber"] == proposal_num
             ):
-                self.is_single_run = False
+                is_single_run = False
 
         train_ids = sorted(train_ids)
         selection = union_selections(
@@ -619,7 +626,7 @@ class DataCollection:
         return DataCollection(
             files, selection=selection, train_ids=train_ids,
             inc_suspect_trains=self.inc_suspect_trains,
-            is_single_run = self.is_single_run,
+            is_single_run = is_single_run,
         )
 
     def _expand_selection(self, selection):
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -593,7 +593,11 @@ class DataCollection:
         files = set(self.files)
         train_ids = set(self.train_ids)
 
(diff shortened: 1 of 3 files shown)


philsmt:
Hi Karim, nice seeing you here 😄 

That's a neat idea to improve this very naive behaviour right now. But please add a check for the proposal number, too.

kakhahmed:
> That's a neat idea to improve this very naive behaviour right now. But please add a check for the proposal number, too.

Yes, that makes sense. I have added a check for the proposal number as well.

extra_data/reader.py
```diff
@@ -593,17 +593,33 @@ def union(self, *others):
         files = set(self.files)
         train_ids = set(self.train_ids)
 
+        run_num = self.run_metadata()[
+            "runNumber"] if self.is_single_run else None
+
+        proposal_num = self.run_metadata()[
+            "proposalNumber"] if self.is_single_run else None
```
takluyver, line 600: Older runs (the file format we call 0.5, before it actually had a version number) don't have this information stored in files. This is making the tests fail.

I guess the simplest thing is just to catch that so that `union` on old data always behaves like they were two separate runs, and the new feature is only for newer data. It's a bit unfortunate to have mysteriously different behaviour with different data, but we can probably live with it.

extra_data/reader.py
```diff
@@ -593,17 +593,33 @@ def union(self, *others):
...
             train_ids.update(other.train_ids)
+            # Set self.is_single_run to False
+            # in case of Unioning multiple run DataCollections.
+
+            if not (
+                other.is_single_run and
+                other.run_metadata()["runNumber"] == run_num and
+                other.run_metadata()["proposalNumber"] == proposal_num
+            ):
+                self.is_single_run = False
```
takluyver, line 613: This shouldn't modify `self` - it should only affect the new DataCollection we return.

kakhahmed: yes, true. Fixed it.

extra_data/reader.py
kakhahmed: added a check for these two keys for the current run format

extra_data/reader.py
```diff
@@ -592,18 +592,41 @@ def union(self, *others):
...
+
+        # Account for old runs (file format 0.5),
+        # which doesn't have these information.
+        if (
+            "runNumber" in self.run_metadata().keys() and
+            "proposalNumber" in self.run_metadata().keys()
+        ):
+            if self.is_single_run:
+                run_num = self.run_metadata()["runNumber"]
+                proposal_num = self.run_metadata()["proposalNumber"]
```
takluyver, line 607: We should check that it's a single run before getting the metadata, because `run_metadata` will fail if it's not (and we don't need the metadata then anyway). We can check it for all the pieces we're unioning together at once: `all(dc.is_single_run for dc in (self,) + others)`.

`self.run_metadata()` isn't cached - it calls HDF5 to get the values every time. So it's probably worth calling it once and storing the dict in a local variable.

extra_data/reader.py
```diff
@@ -592,18 +592,41 @@ def union(self, *others):
...
         for other in others:
             files.update(other.files)
             train_ids.update(other.train_ids)
+            # Set other.is_single_run to False
+            # in case of Unioning multiple run DataCollections.
+
+            if not (
+                other.is_single_run and
+                other.run_metadata()["runNumber"] == run_num and
+                other.run_metadata()["proposalNumber"] == proposal_num
```
takluyver, line 618: You've handled if `self` comes from an older file without this metadata, but you're still assuming that `other` has these details.

takluyver (changes requested):
BTW, if you want to run the tests locally while developing, do something like `pytest -k union`. I usually use `-k` to select just a few tests, because the full suite is quite slow.

Once you've had one PR merged, the CI will also run automatically when you make a PR (for new contributors, I have to click a button to start the CI).

kakhahmed:
Is it safe to consider a data collection union of format version 0.5 not a single run? and have this feature update only for format version 1.0?

Updated the PR and the tests for union.

takluyver:
Yes, I think it's reasonable to say that union-ing together parts of one run only keeps 'single run' status if it's using the newer format (which I think went into use around the start of 2020).

extra_data/reader.py
```diff
@@ -592,18 +592,46 @@ def union(self, *others):
...
+        # which doesn't have runNumber and proposalNumber metadata.
+
+        # DataCollection union of format version = 0.5 is not considered
+        # a single run.
+        single_run_cond = []
+        for dc in (self,) + others:
+            single_run_cond += [
+                dc.is_single_run,
+                dc.run_metadata()["dataFormatVersion"] != '0.5'
+            ]
```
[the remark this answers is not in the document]
kakhahmed: suggestion added

extra_data/reader.py
```diff
@@ -592,18 +592,46 @@ def union(self, *others):
...
+
+        is_single_run = all(single_run_cond)
```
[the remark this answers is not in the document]
kakhahmed: sure, called it `check_same_run`

extra_data/reader.py
```diff
@@ -592,18 +592,46 @@ def union(self, *others):
...
 
         for other in others:
             files.update(other.files)
             train_ids.update(other.train_ids)
 
+            if is_single_run:
+                other_md = other.run_metadata()
+                if (
+                    other_md["runNumber"] != run_num or
+                    other_md["proposalNumber"] != proposal_num
```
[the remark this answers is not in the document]
kakhahmed: Thanks, removed.

kakhahmed:
> I think this is functionally OK now, but I've commented on some places where the code could be a bit clearer.
> 
> I'd be tempted to use sets for this - i.e. build a set of run numbers (& one of proposal numbers), using -1 for missing (either because it's multiple runs or the old format), and then check if there's 1 value in the set and it's not -1. (I'm not saying you need to do that, just thinking out loud :-)

Why not add this here though. Although I don't really understand what do you mean :D

takluyver:
I was thinking something like:

```python
run_nos, prop_nos = set(), set()
for dc in (self,) + others:
    runmd = dc.run_metadata() if dc.is_single_run else {}
    run_nos.add(runmd.get('runNumber', -1))
    prop_nos.add(runmd.get('proposalNumber', -1))

single_run = (len(run_nos) == len(prop_nos) == 1) and (run_nos != {-1}) and (prop_nos != {-1})
```

I don't think this is worth worrying about, though. Thanks for the PR. :smiley:

Issue #210, opened 2021-08-05
Trim dependencies?


takluyver:
EXtra-data has a lot of dependencies when users install it in a fresh environment, and many of them are not strictly required.

- h5py & numpy are clearly required
- psutil is used in a couple of places - is there a sensible alternative to the `cpu_affinity()` method we use to see how many cores a process can use?
- karabo-bridge is only used for streaming data -> move it to a 'bridge' extra?
- matplotlib, pandas and xarray are all used by specific features; it's possible to read data without any of these. But some of those methods are quite important, e.g. `.get_array()`.
- We're aiming to drop the dependency on fabio soon anyway

cc @philsmt

takluyver:
We could also consider removing the `QuickView` class - a simple 'frame switcher' for showing images with matplotlib. It has hardly been changed since 2017, which means either it's a triumph of elegant simplicity, or it's hardly getting used.

On the other hand, matplotib is so widely used that there may not be much point avoiding a dependency on it.

takluyver:
I think we can replace psutil's `cpu_affinity()` method with [`os.sched_getaffinity()`](https://docs.python.org/3.6/library/os.html#os.sched_getaffinity).

It looks like the stdlib function is Unix only, whereas psutil has a Windows implementation for it. But EXtra-data probably isn't used that much on Windows, so I think we can fall back to something like `min(os.cpu_count(), 4)`.

cydanil:
I think it's probably safe to remove `matplotlib`.  
It's only used with the `docs/xpd_examples` notebooks and within `extra_data/utils.py` [for the `QuickView` class](https://github.com/European-XFEL/EXtra-data/blob/master/extra_data/utils.py#L21), which I have a feeling isn't used (and could be deprecated?)

takluyver:
We got rid of the matplotlib dependency as well. The remaining direct dependencies are h5py, numpy, pandas and xarray. The last two could strictly be made optional, but they're familiar packages and used in some highly visible methods that are highlighted in example docs, so I'm inclined to leave them as full dependencies.

There's some discussion in Python packaging about a way to have optional dependencies that are installed by default. If that does happen, we might make use of it.

PR #211, opened 2021-08-05
Make dependencies for karabo_bridge streaming optional

takluyver:
Including making psutil optional. Part of #210.

- [x] What to call the extra? `bridge`? `stream`? `streaming`?
- [x] Document extra

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Install dependencies
       run: |
         python3 -m pip install --upgrade pip
-        python3 -m pip install .[test] --constraint .github/dependabot/constraints.txt
+        python3 -m pip install .[bridge,test] --constraint .github/dependabot/constraints.txt
 
     - name: Test with pytest
       run: |
--- a/docs/streaming.rst
+++ b/docs/streaming.rst
@@ -12,11 +12,22 @@ To stream the data from a file or run unmodified, use the command::
 
 The number (4545) must be an unused TCP port above 1024. It will bind to
 this and stream the data to any connected clients.
-
 Command-line options are explained on the
 :ref:`command reference <cmd-serve-files>` page.
 
-We provide Karabo bridge clients as Python and C++ libraries.
+.. note::
+
+   If you install EXtra-data yourself, streaming data requires some optional
+   dependencies. To ensure you have these, run::
+
+      pip install extra_data[bridge]
+
+   These dependencies are installed in the EuXFEL Anaconda installation on the
+   Maxwell cluster.
+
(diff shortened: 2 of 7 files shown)


tmichela:
What about the cli `karabo-bridge-serve-files`? I guess it will be installed even with the non optional 'bridge' install? that might confuse users 🤔

philsmt:
If I remember correctly, entry points can be made conditional on an extra feature to be installed, but this is apparently not recommended anymore (https://packaging.python.org/specifications/entry-points/). I don't see a proposed alternative

We could just throw an exception if the CLI is called and karabo_bridge is not installed. The idea was that the relative intersection of users using EXtra-data and `karabo-bridge-serve-files` at the same time is smaller than users using `karabo-bridge-serve-files` and `karabo_bridge`.

takluyver:
Maybe that's why we didn't make it optional before. :thinking: 

We could put the files-to-karabo-bridge code in its own tiny package which depends on karabo-bridge and EXtra-data, so neither of them have to depend on the other. I'm a bit reluctant to spin out such small packages, though - my experience is that it's more maintenance work for us and more confusion for users.

tmichela:
I'm also not too excited about packaging something that small...
`karabo_bridge` deps are quite small, maybe that's best to just leave it like that? It's deps are likely to be already present in most environements (numpy, pyzmq, msgpack), with the exception of `msgpack_numpy`, but it itself has no extra dependencies, and we could actually deprecate it by not supporting the format `1.0`. I'm pretty certain that there is no-one using it.

takluyver:
Numpy is a dependency anyway, but psutil, msgpack & pyzmq together (all compiled packages) make for a fairly tempting target. @philsmt noticed users being mildly surprised at how many dependencies EXtra-data & EXtra-geom had - I don't think they were actually complaining, but I thought it might be time to see if we could trim things back a bit.

Would it be reasonable, if `karabo-bridge-serve-files` is used and it can't import what it needs, to exit with a descriptive message?

```
karabo-bridge-serve-files requires additional dependencies:
    pip install karabo-bridge psutil
```

tmichela:
fine with me 👍

philsmt:
This helps a lot, thanks!

takluyver:
Any preference what the extra is called? That's the name you'd need in square brackets for installation, e.g. `pip install extra_data[bridge]` .

tmichela:
either `bridge` or `stream` work for me

takluyver:
OK, let's stick with `bridge` unless @philsmt wants to make a case for some other name. We're already using 'bridge' in the command name and in the client library, so that makes the connection fairly clear.

PR #213, opened 2021-08-06
Remove _to_cbf functions, fabio dependency

takluyver:
This can wait a little while to give anyone still using them a chance to notice the deprecation (from 1.7, released 3rd August 2021).


takluyver:
It's been nearly 2 months since we added the deprecation warning, and no-one has been in touch to say that they're using these. That's what I was expecting, as we've never really documented or advertised them. So I plan to merge this and release EXtra-data 1.8 without these functions next week.

PR #217, opened 2021-09-06
Fix listing of RUN keys in lsxfel and run.info()

takluyver:
While investigating something else, I noticed that #206 broke listing the RUN keys with things like `lsxfel --detail '*/XGM/*'`. This fixes it again.


tmichela:
LGTM

PR #218, opened 2021-09-09
Pre-allocate arrays to read multi-module detector data

takluyver:
This should improve performance by avoiding concatenating large arrays, which involves copying a lot of data in memory. cc @kakhahmed .

Some of the code for this is shared with the virtual CXI writing, as they're now doing a similar task - identifying blocks of source data which can be directly mapped to blocks of the output space.

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -24,6 +24,7 @@ __all__ = [
 log = logging.getLogger(__name__)
 
 MAX_PULSES = 2700
+NO_PULSE_ID = 9999
 
 
 def multimod_detectors(detector_cls):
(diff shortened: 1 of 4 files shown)


takluyver:
If you use pulse selection, this selects pulse indices from the file with a numpy array ('fancy indexing'). I suspected that this was inefficient, so I spent a while playing with identifying regular selections and using an h5py MultiBlockSlice (representing an HDF5 hyperslab selection) to process them, in [branch xtdf-slice-pulses](https://github.com/European-XFEL/EXtra-data/compare/refactor-multimod-array...xtdf-slice-pulses).

However, it doesn't seem to make a big difference, in testing with some LPD data @daviddoji pointed me to. So I'm leaving it out for now as extra complexity.

daviddoji:
LGTM!

PR #220, opened 2021-09-23
Use a temporary array to read detector data with pulse selection

takluyver:
This aims to speed up reading only selected frames from XTDF detector data, using the LPD/AGIPD/DSSC `.get_array()` method with `pulses=`. This was prompted by @daviddoji's use case reading LPD data.

I believe this is working around a [performance issue in HDF5](https://forum.hdfgroup.org/t/performance-reading-data-with-non-contiguous-selection/8979), which appears to have been around for years. It looks like HDF5 doesn't realise that it can copy large blocks of data, and falls back to copying point by point. By using a temporary array, we can persuade it that the source & destination are sufficiently similar so it works more efficiently. Then we copy the selected data to the output array with [`numpy.compress()`](https://numpy.org/doc/stable/reference/generated/numpy.compress.html).

This involves some extra temporary memory use for the intermediate array. It should never be more than the size of a single file, and it should be possible for the operating system to do virtual memory tricks and only allocate memory for the frames we're actually reading, not the gaps between them.

The timings below for reading LPD parallel gain data with different numbers of pulses selected (e.g. `.get_array('image.data', pulses=np.s[:1])`). There are 100 pulses per train, but parallel gain mode records all 3 gain stages as separate frames, so we're actually reading 3n of 300 frames in each case. Times are per train; I used 10 or 25 trains to get a better average. I ran each one a couple of times to ensure data was cached.

Pulses selected | Parent branch | This branch
----------------------|---------------------|---------------
1 | 0.56 s | 0.026 s
3 | 1.7 s | 0.030 s
10 | 5.6 s | 0.057 s
50 | 27.8 s | 0.20 s
100 (all) | 0.201 s | 0.204 s

Without this change, you can see that reading even a single frame is slow, and the time scales linearly with the number of frames to read (except when we read all of them). With this change, reading a subset of frames per train is much faster.

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -8,7 +8,7 @@ import xarray
 
 from .exceptions import SourceNameError
 from .reader import DataCollection, by_id, by_index
-from .read_machinery import roi_shape
+from .read_machinery import roi_shape, DataChunk
 from .writer import FileWriter
 from .write_cxi import XtdfCXIWriter, JUNGFRAUCXIWriter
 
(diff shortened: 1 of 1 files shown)


daviddoji:
LGTM.
Great improvement @takluyver.

takluyver:
Thanks David! I'll hold off merging for now, because this goes into the branch for #218, and I don't want to complicate that PR.

PR #226, opened 2021-10-13
Expose public KeyData.train_id_coordinates() method

takluyver:
Looking at some code used at FXE, I noticed that there was a pattern of loading data as a NumPy array, along with an array of train IDs to use with the data. You can already get an [xarray DataArray](http://xarray.pydata.org/en/stable/user-guide/data-structures.html#dataarray) which has the train IDs incorporated as labels, but not everyone's going to use Xarray.

```python
jf = run['FXE_XAD_JF1M/DET/JNGFR02:daqOutput', 'data.adc']

# This already works
jf_xarr  = jf.xarray()
jf_data = jf_xarr.values
jf_train_ids = jf.coords['trainId'].values

# New option
jf_data = jf.ndarray()
jf_train_ids = jf.train_ids_for_data()
```

### DIFF
--- a/docs/reading_files.rst
+++ b/docs/reading_files.rst
@@ -136,6 +136,8 @@ below, e.g.::
        :doc:`dask_averaging`
          An example using Dask with EuXFEL data
 
+   .. automethod:: train_ids_for_data
+
    .. automethod:: select_trains
    
    .. automethod:: split_trains
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -197,8 +197,20 @@ class KeyData:
 
         return out
 
-    def _trainid_index(self):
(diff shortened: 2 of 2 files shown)


philsmt:
I have to admit I liked the original name better, but let's not open that can of worms. LGTM

takluyver:
I kind of liked the original name (trainid_index) too. But anyone who looks into the files might reasonably expect that 'trainid index' corresponds to the `INDEX/trainId` dataset, which this doesn't (the `.train_ids` attribute is closer to that).

Other naming ideas welcome. I agree we shouldn't get too hung up on naming, but on the other hand, picking decent names for things matters, and it's easy to change it now.

philsmt:
Thanks, LGTM

takluyver:
Thanks!

PR #228, opened 2021-10-16
A robust way to retrieve reduced key values 

philsmt:
Quite often, `CONTROL` data is assumed constant for a given run or other unit of analysis, e.g. for the detector conditions when applying corrections. A common pattern here is to omit proper checks that this is actually the case. This PR adds the method `KeyData.single_value` which performs such checks automatically in a configurable way and a corresponding entry point in `DataCollection.get_single_value`.

The implementation is centered around [`np.allclose`](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html) which defines a relative and absolute tolerance combined into a single check. The reduction method to obtain the value to check the data against can be configured, but the median is used by default. It may be passed either as `Callable`, the string of a global symbol in `numpy` such as `median` or `mean` or the special case `first`.

Technically both the `str` version and `first` are redundant with just passing a `Callable`, e.g. `np.mean` instead of `'mean'` and `lambda x: x[0]` instead of `first`, but it seemed convenient to provide. I don't have strong feelings about this.

If the reduced value is within the specified tolerances, it is returned. If not, a `ValueError` is raised. I was not sure whether it makes sense to add a derived exception class here to distinguish it from other `ValueError`s.

### DIFF
--- a/docs/reading_files.rst
+++ b/docs/reading_files.rst
@@ -146,6 +146,11 @@ below, e.g.::
    
       .. versionadded:: 1.7
 
+   .. automethod:: as_single_value
+   
+      .. versionadded:: 1.9
+
+
 The run or file object (a :class:`DataCollection`) also has methods to load
 data by sources and keys. :meth:`get_array`, :meth:`get_dask_array` and
 :meth:`get_series` are directly equivalent to the options above, but other
--- a/extra_data/exceptions.py
+++ b/extra_data/exceptions.py
@@ -37,3 +37,16 @@ class MultiRunError(ValueError):
             "because you have used .union() to combine data. Please retrieve "
             "this information before combining."
         )
+
+
+class NoDataError(ValueError):
+    def __init__(self, source, key=None):
+        self.source = source
+        self.key = key
+
+    def __str__(self):
+        if self.key is not None:
+            return 'This data is empty for key {!r} of source {!r}'.format(
+                self.key, self.source)
+        else:
+            return 'This data is empty for source {!r}'.format(self.source)
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -169,7 +169,7 @@ class KeyData:
             idxs = np.argsort(train_ids)
             return counts[idxs]
 
-    def single_value(self, rtol=1e-5, atol=0.0, reduce_by='median'):
+    def as_single_value(self, rtol=1e-5, atol=0.0, reduce_by='median'):
         """Retrieve a single reduced value if within tolerances.
 
         The relative and absolute tolerances *rtol* and *atol* work the
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -169,6 +169,42 @@ class KeyData:
             idxs = np.argsort(train_ids)
             return counts[idxs]
 
+    def single_value(self, rtol=1e-5, atol=0.0, reduce_by='median'):
+        """Retrieve a single reduced value if within tolerances.
+
+        The relative and absolute tolerances *rtol* and *atol* work the
+        same way as in ``numpy.allclose``. The default relative tolerance
+        is 1e-5 with no absolute tolerance. The data for this key is compared
+        against a reduced value obtained by the method described in *reduce_by*.
+
+        This may be a callable taking the key data, the string value of a
+        global symbol in the numpy packge such as 'median' or 'first' to use
+        the first value encountered. By default, 'median' is used.
+
+        If within tolerances, the reduced value is returned.
+        """
+
+        data = self.ndarray()
+
+        if callable(reduce_by):
+            value = reduce_by(data)
+        elif isinstance(reduce_by, str) and hasattr(np, reduce_by):
+            value = getattr(np, reduce_by)(data, axis=0)
+        elif reduce_by == 'first':
+            value = data[0]
+        else:
+            raise ValueError('invalid reduction method (may be callable, '
+                             'global numpy symbol or "first")')
+
+        if not np.allclose(data, value, rtol=rtol, atol=atol):
+            adev = np.max(np.abs(data - value))
+            rdev = np.max(np.abs(adev / value))
+
+            raise ValueError(f'data values are not within tolerance '
+                             f'(absolute: {adev:.3g}, relative: {rdev:.3g})')
+
+        return value
+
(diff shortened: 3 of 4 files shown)


takluyver:
Thanks for this! I've been meaning to implement something like this since you suggested it, but you've beaten me to it. :slightly_smiling_face: 

I'm ambivalent about adding `DataCollection` methods for new functionality where KeyData is a good way to access it; I think that will be the main interface in the future. It doesn't hurt to have a DataCollection method as well, though.

I was thinking of having a parameter for how much of the data it checks: all of it (as you've implemented), a subset at the beginning and end, to check for jitter & drift without loading all the data, or just take the first version and assume it's constant (although `get_run_value` covers that for control data). This could save time & memory, especially for multi-dimensional data. On the other hand, I imagine the main use of this will be for one-number-per-train data, so maybe it's best to keep it simple for now and see how it goes in the wild.

extra_data/keydata.py
```diff
@@ -169,6 +169,42 @@ def data_counts(self, labelled=True):
...
+        against a reduced value obtained by the method described in *reduce_by*.
+
+        This may be a callable taking the key data, the string value of a
+        global symbol in the numpy packge such as 'median' or 'first' to use
+        the first value encountered. By default, 'median' is used.
+
+        If within tolerances, the reduced value is returned.
+        """
+
+        data = self.ndarray()
```
takluyver, line 187: Do we want to check for a length 0 array and raise our own error? At present, you'd get `nan` with a RuntimeWarning for median/mean, and an IndexError for 'first'.

philsmt: Oh good catch, that is definitely not consistent. I could see two options:
* Raise an error
* Return `None`

I gravitate towards the first, but then again regular read methods also don't throw an error an `None` could not be returned in any other case. It boils down to EAFP vs LBYL.

philsmt: Hmmm, any preference on the exception type? A simple one might be `RuntimeError`, but that's maybe a bit too generic. Functions like `np.max()` throw a `ValueError` when called on an empty array. I don't see any fitting type in `extra_data.exceptions`. Maybe a `NoDataError` to be precise?

takluyver: Python built in `min()` and `max()` also throw ValueError. The [`statistics` module](https://docs.python.org/3/library/statistics.html) defines its own StatisticsError, inheriting from ValueError. `functools.reduce` is something of an outlier, throwing TypeError.

So I'd either use ValueError or subclass it as NoDataError. I don't have a particular preference between those options.

philsmt:
After some discussion within DA, I changed the name to `KeyData.as_single_value()`. Shall I keep the name for `DataCollection.get_single_value`?

takluyver:
I'm ambivalent about the name on `DataCollection` - I think `get_single_value` has a similar issue to `single_value`, but I also don't particularly think we need a new DataCollection method, as I plan to give all examples using the KeyData API. So the clarity of the name if we do have one there isn't so important.

Don't forget to add `as_single_value` to the reference docs. The DataCollection method, if we do add one, could be undocumented.

Other than that, LGTM.

philsmt:
Then let's just get rid of the method in `DataCollection`.

I've added a documentation entry and cleaned up the commits.

docs/reading_files.rst
```diff
@@ -146,6 +146,11 @@ below, e.g.::
    
       .. versionadded:: 1.7
 
+   .. automethod:: as_single_value
```
takluyver, line 149: I've roughly grouped the reading methods before the selection methods, if you want to move this up - maybe just after `dask_array`, to keep the various array-like methods together?

philsmt: I don't mind, but I put it down there exactly because we concluded it's not just a reading method 🙃

extra_data/exceptions.py
```diff
@@ -37,3 +37,15 @@ def __str__(self):
...
+class NoDataError(ValueError):
+    def __init__(self, source, key=None):
+        self.source = source
+        self.key = key
+
+    def __str__(self):
+        if self.key is not None:
+            return f'{self.source}, {self.key}'
+        else:
+            return self.source
```
takluyver, line 51: I'd give this a slightly more descriptive message, something like "No data in run/selection for source ABC, key DEF"

extra_data/tests/test_keydata.py
```diff
@@ -169,3 +169,43 @@ def test_drop_empty_trains(mock_sa3_control_data):
...
+
+    # Monkeypatch some actual data into the KeyData object
+    data = np.arange(flux.shape[0])
+    monkeypatch.setattr(flux, 'ndarray', lambda: data)
+
+    with pytest.raises(ValueError):
+        # Try some tolerances that have to fail.
+        flux.as_single_value()
+        flux.as_single_value(atol=1)
+        flux.as_single_value(rtol=0.1)
```
takluyver, line 191: N.B. you're only actually testing the first one of these. If you want to test all of them, they each need their own `with pytest.raises` block.

takluyver: (Not precisely the first, of course - it tests that *any* of them raise ValueError, rather than *all* of them.)

extra_data/exceptions.py
philsmt: I tried to model it after the builtin `KeyError` or `AttributeError`, which only state the thing missing and imply the reason by their type. But we can change it if you feel strongly about it.

extra_data/tests/test_keydata.py
philsmt: Oops...

docs/reading_files.rst
takluyver: Not *just* a reading method, but it's more like `.ndarray()` than `.select_trains()` - it gives you back data rather than another KeyData object.

Dask arrays kind of mess up this categorisation, though, so I don't feel strongly about it.

takluyver:
LGTM. Adjusting the error message is optional - if I haven't persuaded you, feel free to merge this straight away.

extra_data/exceptions.py
philsmt: I see your point, without knowledge that `SourceNameError` and `PropertyNameError` exist one might not know that a `NoDataError` signals existance but emptiness. In addition, those errors are already fairly chatty.

PR #242, opened 2021-11-04
Updating examples in documentation

takluyver:
- More focused examples in quickstart (cc @lucagelisio )
- Using the new source & key data APIs in examples:`run[src]` & `run[src, key]`
- New example about aligning data by train IDs, using a run already in the XMPL folder. This is a common need, and there are a few different options depending on precisely what you need.

I also want to rework the parallel processing examples, to show off `split_trains` among other things, but I decided to leave that for a separate PR.

RTD should build a preview shortly after I open the PR - look for a link in the 'checks' section.


tmichela:
LGTM!

takluyver:
Thanks @tmichela !

PR #249, opened 2021-11-23
Add repr for DataCollection objects

takluyver:
We have repr methods for KeyData & SourceData already, so this seems like an obvious addition. :slightly_smiling_face:


tmichela:
LGTM

PR #250, opened 2021-11-23
Deprecate QuickView class

takluyver:
We've never documented this, and I don't remember hearing from anyone using it. This is the only bit of EXtra-data that uses matplotlib, so getting rid of it lets us drop that dependency.

I've no objection to keeping it if we think people are using it, or even if we just think it's useful enough to be worth documenting.


tmichela:
LGTM

PR #257, opened 2021-11-29
Fix .data_counts(labelled=False) when source doesn't have some trains

### LINKED ISSUE #256: AssertionError in .data_counts(labelled=False)
@yohey-uemura discovered this over the weekend. When writing this method, I had overlooked the possibility of having a selected train which doesn't correspond to any data for the given source. For instance, in this run, LPD08 starts from train 1232492275, but other sources in the run start from 1232492274.

```python
run = open_run(2958, 172)
run[
    'FXE_DET_LPD1M-1/DET/8CH0:xtdf', 'image.cellId'
].data_counts(labelled=False)
```

```
AssertionError                            Traceback (most recent call last)
<ipython-input-14-749f3505f4df> in <module>
      1 run[
      2     'FXE_DET_LPD1M-1/DET/8CH0:xtdf', 'image.cellId'
----> 3 ].data_counts(labelled=False)

/gpfs/exfel/sw/software/xfel_anaconda3/1.1.2/lib/python3.7/site-packages/extra_data/keydata.py in data_counts(self, labelled)
    168             # should be in order, but sometimes trains are written out of order.
    169             # Reorder the counts to match self.train_ids.
--> 170             assert len(train_ids) == len(self.train_ids)
    171             assert np.isin(train_ids, self.train_ids).all()
    172             idxs = np.argsort(train_ids)

AssertionError: 
```

takluyver:
I had missed that the train IDs found in all chunks for a source may still miss some selected train IDs, where the relevant DA didn't record those trains at all (especially at the beginning or end of a run). This should fix that.

Closes #256.

### DIFF
--- a/.github/dependabot/constraints.txt
+++ b/.github/dependabot/constraints.txt
@@ -1,3 +1,4 @@
+coverage==6.1.2
 cycler==0.11.0
 h5py<3.7.0
 karabo-bridge==0.6.1
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -164,13 +164,16 @@ class KeyData:
             import pandas as pd
             return pd.Series(counts, index=train_ids)
         else:
-            # self.train_ids is always sorted. The train IDs from chunks
-            # should be in order, but sometimes trains are written out of order.
-            # Reorder the counts to match self.train_ids.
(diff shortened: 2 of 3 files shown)


tmichela:
LGTM

PR #261, opened 2021-12-07
Fix union when we've already got a non-single-run data collection

takluyver:
@daviddoji relayed an error with code like this:

```python
run = open_run(prop, run_nums[0])
for run_num in run_nums[1:]:
    run = run.union(open_run(proposal, run_num))
```

After the first `.union()`, the DataCollection object is not a single run. Calling `.union()` the second time failed because it tried to get `self.run_metadata()`. This should fix it by always checking `.is_single_run` before calling `.run_metadata()`.

cc @kakhahmed

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -584,30 +584,24 @@ class DataCollection:
         """
         files = set(self.files)
         train_ids = set(self.train_ids)
-        md = self.run_metadata()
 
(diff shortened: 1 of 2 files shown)


kakhahmed:
Thanks for the fix, didn't think of that situation.

LGTM

takluyver:
Thanks!

PR #262, opened 2021-12-08
CorrectionData interface

philsmt:
There is still work left on adding the smaller detectors, finishing touches on the documentation and testing. I'd like to finally get it out however to start the likely lengthy review process. I will expand this description over time.

This PR adds a new `CalibrationData` interface to interact with CalCat for the purpose of obtaining metadata and the datasets of calibration constants. It is centered around the needs of data analysis and by design does not support constant injection or any other database operations. It should also be useful to the offline calibration notebooks for corrections, with other code internal to `pycalibration` handling the injection parts.

It attempts to be somewhat similar to the recent `KeyData` and `SourceData` interfaces by having an immutable state and various methods for (meta)data retrieval. In addition, its data retrieval methods typically allow to overwrite or extend the internal immutable state just for this call, say to query with almost the same set of operating condition parameter except one. The `CalibrationData` itself may be used in principle for any detector (instance), but there is a subclass for each detector type that encodes knowledge about the used operating conditions and where to get there values. In almost all cases, these classes should be used instead.

```python
dssc_caldata = DSSC_CalibrationData('SCS_DET_DSSC1M-1', memory_cells=400)
metadata = dssc_caldata.metadata(calibrations=['Offset'])
data = dssc_caldata.ndarray(calibrations=['Offset'])
```

The detector-specific classes also encode how to obtain the operating conditions from the control devices and other sources, which is meant to be the primary means of obtaining a `CalibrationData` instance in practice:

```
dc = open_run(...)
dssc_caldata = DSSC_CalibrationData.from_data(dc)
```

It uses two auxialiary types `OperatingCondition` and `CCVMetadata`, which are basically just `dict`s with some extra magic to serve their specific purpose.

### DIFF
--- a/extra_data/caldata.py
+++ b/extra_data/caldata.py
@@ -1,0 +1,1016 @@
+
+"""Interfaces to calibration constant data."""
+
+from datetime import datetime, date, time, timezone
+from enum import IntFlag
+from functools import lru_cache
+from os import getenv
+from pathlib import Path
+import re
+import socket
+
+import numpy as np
+import h5py
+
+from . import SourceNameError, PropertyNameError, MultiRunError, NoDataError
+
+from calibration_client import CalibrationClient
+from calibration_client.modules import Detector, PhysicalDetectorUnit, \
+    Parameter, Calibration, CalibrationConstantVersion
+
+
+__all__ = [
+    'BadPixels',
+    'CalCatError',
+    'AGIPD_CalibrationData',
+    'LPD_CalibrationData',
+    'DSSC_CalibrationData',
+    'JUNGFRAU_CalibrationData',
+    'PNCCD_CalibrationData',
+    'EPIX_CalibrationData',
+]
+
+
+class BadPixels(IntFlag):
+    """The European XFEL Bad Pixel Encoding."""
+
+    OFFSET_OUT_OF_THRESHOLD = 1 << 0
+    NOISE_OUT_OF_THRESHOLD = 1 << 1
+    OFFSET_NOISE_EVAL_ERROR = 1 << 2
+    NO_DARK_DATA = 1 << 3
+    CI_GAIN_OF_OF_THRESHOLD = 1 << 4
+    CI_LINEAR_DEVIATION = 1 << 5
+    CI_EVAL_ERROR = 1 << 6
+    FF_GAIN_EVAL_ERROR = 1 << 7
+    FF_GAIN_DEVIATION = 1 << 8
+    FF_NO_ENTRIES = 1 << 9
+    CI2_EVAL_ERROR = 1 << 10
+    VALUE_IS_NAN = 1 << 11
+    VALUE_OUT_OF_RANGE = 1 << 12
+    GAIN_THRESHOLDING_ERROR = 1 << 13
+    DATA_STD_IS_ZERO = 1 << 14
+    ASIC_STD_BELOW_NOISE = 1 << 15
+    INTERPOLATED = 1 << 16
+    NOISY_ADC = 1 << 17
+    OVERSCAN = 1 << 18
+    NON_SENSITIVE = 1 << 19
+    NON_LIN_RESPONSE_REGION = 1 << 20
+    WRONG_GAIN_VALUE = 1 << 21
+
+
+class OperatingCondition(dict):
+    """Dictionary for operating condition parameters.
+
+    A subclass of the regular Python dict to handle the semantics of
+    operating conditions for calibration constants in the form of
+    multiple layers to specify values and required/optional values.
+
+    The multiple layers allow to overwrite default or earlier specified
+    values at later times when actually using an operating condition.
+    The actual parameters contained in the concretized operating
+    conditions however (and thus in the object as dict itself) are only
+    those actually added via set_required and set_optional, with
+    addtional keys contained in one or more layers ignored.
+
+    Any given parameter of an operating condition may either be required
+    or optional. A required key is always present and accordingly must
+    have a default value. An optional key will be present if it is
+    specified in any of the layers or has a default value. If however
+    its value is set to None explicitly in any layer, the default value
+    is ignored.
+
+    A parameter name has two representations, the canonical database
+    form (e.g. Sensor Bias Voltage) and a key form, which must be a
+    valid name for a Python symbol. By default, the key form is the all
+    lower-case version of the database name with spaces replaced by
+    underscores (e.g. sensor_bias_voltage). The key form is expected in
+    the parameter layers.
+
+    Example:
+
+    .. code-block:: python
+
+       cond = OperatingCondition(
+           dict(key_a=1, key_b=2),
+           dict(key_a=3, key_c=None)
+       )
+
+       cond.set_required('Key A', -1)
+       cond.set_optional('Key B')
+       cond.set_optional('Key C', 5)
+
+       assert cond == {'Key A': 1, 'Key B': 2}
+    """
+
+    def __init__(self, *param_layers):
+        """Initialize a new operating condition.
+
+        Args:
+            *param_layers (list of dict): Parameters layers with values
+                with the lowest layer (passed first) taking precedence
+                for any given parameter, which must be referred to by
+                their key name.
+        """
+
+        self._param_layers = param_layers
+
+    def set_required(self, db_name, default, key_name=None):
+        """Set a required parameter.
+
+        Args:
+            db_name (str): Parameter name in database form.
+            default (Any): Default value for this parameter if not found
+                in any parameter layer.
+            key_name (str, optional): Parameter name in key form, if
+                omitted obtained from database form.
+
+        Returns:
+            None
+        """
+
+        assert default is not None, \
+            'Required condition parameter must have a default'
+        self.set_optional(db_name, default, key_name)
+
+    def set_optional(self, db_name, default=None, key_name=None):
+        """Set an optional parameter.
+
+        Args:
+            db_name (str): Parameter name in database form.
+            default (Any, optional): Default value for this parameter if
+                not found in any parameter layer. If omitted, the
+                parameter is not added at all in this case.
+            key_name (str, optional): Parameter name in key form, if
+                omitted obtained from database form.
+
+        Returns:
+            None
+        """
+
+        if key_name is None:
+            key_name = db_name.lower().replace(' ', '_')
+
+        for layer in self._param_layers:
+            if key_name in layer:
+                if layer[key_name] is not None:
+                    self[db_name] = layer[key_name]
+                return
+        else:
+            if default is not None:
+                self[db_name] = default
+
+    def api_encode(self, caldata):
+        """Encode operating condition to CalCat API format.
+
+        Args:
+            caldata (CalibrationData): Calibration data instance used to
+                interface with database.
+
+        Returns:
+            (dict) Operating condition for use in CalCat API.
+        """
+
+        return {'parameters_conditions_attributes': [
+            {'parameter_id': caldata.parameter_id(k), 'value': str(v)}
+            for k, v in self.items()
+        ]}
+
+
+class CCVMetadata(dict):
+    """Dictionary for CCV metadata."""
+
+    def set_pdu_ccv(self, modno, calibration_name, entry):
+        """Set a PDU CCV entry.
+
+        Convenience method to handle insertion of keys if missing.
+        """
+
+        pdu_metadata = self.get(modno, dict())
+        pdu_metadata[calibration_name] = entry
+        self[modno] = pdu_metadata
+
+    def __str__(self):
+        """Pretty-print CCV metadata using pandas."""
+
+        import pandas as pd
+
+        res = {pdu_idx: {cal_name: ccv_data['ccv_name']
+                         for cal_name, ccv_data in pdu_data.items()}
+               for pdu_idx, pdu_data in self.items()}
+
+        return str(pd.DataFrame.from_dict(res, orient='index'))
+
+
+class CalCatError(RuntimeError):
+    """CalCat API error."""
+
+    def __init__(self, response):
+        super().__init__(response['info'])
+
+
+class CalibrationData:
+    """Calibration data for detectors.
+
+    European XFEL uses a web app and database to store records about the
+    characterization of detectors and the data necessary to their
+    correction, collectively called CalCat. The default installation is
+    available at https://in.xfel.eu/calibration.
+
+    A detector is identified by a name (e.g. SPB_DET_AGIPD1M-1) and
+    consists of one or more detector modules. The modules are a virtual
+    concept and may be identified by their number (e.g. 3), the Karabo
+    data aggregator in EuXFEL's DAQ system they're connected to
+    (e.g. AGIPD05) or a virtual device name describing their relative
+    location (e.g. Q3M2).
+
+    A detector module is mapped to an actual physical detector unit
+    (PDU), which may be changed in case of a physical replacement. When
+    characterization data is inserted into the database, it is attached
+    to the PDU currently mapped to a module and not the virtual module
+    itself.
+
+    Characterization data is organized by its type just called
+    calibration (e.g. Offset or SlopesFF) and the operating condition it
+    was taken in, which is a mapping of parameter keys to their values
+    (e.g. Sensor bias voltage or integration time). Any unique
+    combination of calibration (type) and operating condition is a
+    calibration constant (CC). Any individual measurement of a CC is
+    called a calibration constant version (CCV). There may be many CCVs
+    for any given CC.
+
+    Note that while a connection to CalCat is possible from anywhere,
+    the actual calibration data referred to is only available on the
+    European XFEL infrastructure.
+    """
+
+    get_detector_keys = ['id', 'name', 'identifier', 'karabo_name',
+                         'karabo_id_control', 'description']
+    get_pdu_keys = ['id', 'physical_name', 'karabo_da', 'virtual_device_name',
+                    'detector_type_id', 'detector_id', 'description']
+
+    calibrations = set()
+
+    client = None
+
+    def __init__(self, detector_name, modules=None, client=None,
+                 **condition_params):
+        """Initialize a new CalibrationData object.
+
+        Args:
+            detector_name (str): Name of detector in CalCat.
+            modules (Iterable of int, optional): Module numbers to
+                query for or None for all available (default).
+            client (CalibrationClient, optional): Client for CalCat
+                communication, global one by default.
+            **condition_params: Operating condition parameters defined
+                on an instance level.
+        """
+
+        if client is None:
+            client = self.__class__.client
+
+            if client is None:
+                raise ValueError(f'need client passed by argument or prior '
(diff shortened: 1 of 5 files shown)


extra_data/caldata.py
```diff
@@ -0,0 +1,1016 @@
...
+        """
+
+        if key_name is None:
+            key_name = db_name.lower().replace(' ', '_')
+
+        for layer in self._param_layers:
+            if key_name in layer:
+                if layer[key_name] is not None:
+                    self[db_name] = layer[key_name]
+                return
```
philsmt, line 157: LGTM.com complains here about `This 'for' statement has a redundant 'else' as no 'break' is present in the body.`

Interesting it does not realize the `return` statement. A `break` would be identical in this case, but now I'm curious on an academic level: Do you think it should be generally preferred in such a case?

takluyver: I guess LGTM.com is right in that the `else` is equivalent to simply putting the code after the for loop. You could argue that using else makes the intent clearer, but I'm not sure I would argue that.

For this use case - do something side-effect-ish with one thing in this collection, or fall back to doing a default thing - I'd marginally lean towards a for/else with a break. I think that makes refactoring easier e.g. if you want to add another step afterwards which happens regardless of this step. But if the loop & the fallback were to retrieve a value, I'd lean towards `return` (and no `else`).

extra_data/caldata.py
```diff
@@ -0,0 +1,1016 @@
...
+    'LPD_CalibrationData',
+    'DSSC_CalibrationData',
+    'JUNGFRAU_CalibrationData',
+    'PNCCD_CalibrationData',
+    'EPIX_CalibrationData',
+]
+
+
+class BadPixels(IntFlag):
+    """The European XFEL Bad Pixel Encoding."""
```
takluyver, line 35: Have we confirmed this is exactly the same for all detectors? It's the kind of thing I could imagine starting off the same and then growing different additional fields in different detectors.

philsmt: As far as I've understood, yes, this is indeed what was agreed upon for all detectors (@kakhahmed for confirmation). There are actually additional enums like `BadPixelsFF` for AGIPD-only calculations, which are then afterwards translated to this one. That is the reason I moved them recently to AGIPD-specific parts of `pycalibration`.

Some flags are only used for a single detector right now, but they appear broad enough to be sensible.

kakhahmed: yes, `BadPixels` is already used for all detectors

takluyver: Thanks!

extra_data/caldata.py
```diff
@@ -0,0 +1,1016 @@
...
+    illuminated_calibrations = {'BadPixelsFF', 'SlopesFF'}
+
+    def __init__(self, detector_name, modules=None, client=None,
+                 **condition_params):
+        super().__init__(detector_name, modules=modules, client=client,
+                         **condition_params)
+
+    @classmethod
+    def from_data(cls, data, detector=None, modules=None, client=None,
+                  ctrl_device_id=None, **condition_params):
```
takluyver, line 757: I think all these methods that take `**condition_params` will accept and ignore any keyword argument, so if I make a typo like `intergation_time`, the value I pass will be silently ignored? I think we should error on unrecognised keyword arguments, to catch this kind of mistake.

extra_data/caldata.py
```diff
@@ -0,0 +1,1016 @@
...
+                         'karabo_id_control', 'description']
+    get_pdu_keys = ['id', 'physical_name', 'karabo_da', 'virtual_device_name',
+                    'detector_type_id', 'detector_id', 'description']
+
+    calibrations = set()
+
+    client = None
+
+    def __init__(self, detector_name, modules=None, client=None,
+                 **condition_params):
```
takluyver, line 256: If the `.from_data()` classmethod is meant as the primary public API to construct these classes, I might have `__init__` accept `condition_params` as a dict rather than keyword arguments. Not a big deal, but having fewer places taking arbitrary keyword args makes life easier if you want to add other specific arguments later on.

extra_data/caldata.py
```diff
@@ -0,0 +1,1016 @@
...
+    A subclass of the regular Python dict to handle the semantics of
+    operating conditions for calibration constants in the form of
+    multiple layers to specify values and required/optional values.
+
+    The multiple layers allow to overwrite default or earlier specified
+    values at later times when actually using an operating condition.
+    The actual parameters contained in the concretized operating
+    conditions however (and thus in the object as dict itself) are only
+    those actually added via set_required and set_optional, with
+    addtional keys contained in one or more layers ignored.
```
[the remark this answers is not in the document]
philsmt: For public documentation purposes, the idea of layers is to identify where a default was defined. An implementation may know a good default value (e.g. Acquisition rate of 4.5 MHz) applicable in most cases, but the user knows they want an older constant without this condition. The `OperatingCondition` type supports this case by defining the default value in layers, but layer supplying a `None` in any of the layers of higher priority.

However, with the rework attempt this class and layers will almost definitely disappear.

extra_data/caldata.py
```diff
@@ -0,0 +1,1016 @@
...
+        import pandas as pd
+
+        res = {pdu_idx: {cal_name: ccv_data['ccv_name']
+                         for cal_name, ccv_data in pdu_data.items()}
+               for pdu_idx, pdu_data in self.items()}
+
+        return str(pd.DataFrame.from_dict(res, orient='index'))
+
+
+class CalCatError(RuntimeError):
```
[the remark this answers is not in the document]
philsmt: 👍

extra_data/caldata.py
```diff
@@ -0,0 +1,1016 @@
...
+            **condition_params: Additional operating condition
+                parameters defined only for this query, ignored if
+                condition is passed.
+
+        Returns:
+            (CCVMetadata) CCV metadata result.
+        """
+
+        metadata = CCVMetadata()
+        self._retrieve_constant_version(
```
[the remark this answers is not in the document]
philsmt: Yes, the current code was a bit WIP, sorry!

extra_data/caldata.py
```diff
@@ -0,0 +1,1016 @@
...
+
+    def set_pdu_ccv(self, modno, calibration_name, entry):
+        """Set a PDU CCV entry.
+
+        Convenience method to handle insertion of keys if missing.
+        """
+
+        pdu_metadata = self.get(modno, dict())
+        pdu_metadata[calibration_name] = entry
+        self[modno] = pdu_metadata
```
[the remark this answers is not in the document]
philsmt: Yes, totally. I had this thought spinning in my mind there is a method doing what I want, but did not find it. Thanks!

extra_data/caldata.py
```diff
@@ -0,0 +1,1016 @@
...
+        self._detector = None
+        self._pdus = None
+
+    @classmethod
+    def new_client(cls, client_id, client_secret, user_email, installation='',
+                   base_url='https://in.xfel.eu/{}calibration'):
+        """Create a new calibration client object.
+
+        The client object is saved as a class property and is
+        automatically to any future CalibrationData objects created, if
```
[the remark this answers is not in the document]
philsmt: Yes, you're right. Changing this method to be a `staticmethod` and using `CorrectionData` directly.

extra_data/caldata.py
```diff
@@ -0,0 +1,1016 @@
...
+
+        if ctrl_device_id is None:
+            ctrl_device_id = caldata.detector['karabo_id_control'] + \
+                'MDL/FPGA_COMP'
+
+        try:
+            ctrl_src = data[ctrl_device_id]
+        except KeyError:  # Should be SourceNameError
+            raise ValueError(f'control source {ctrl_device_id} not found in '
+                             f'DataCollection')
```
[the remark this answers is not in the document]
philsmt: 👍

philsmt:
Thanks for the initial review. I've replied and resolved a few of the minor issues. The remaining ones center around how to deal with operating condition parameters and state. As just discussed with you privately, I will to address this by baking the operating condition into the `CorrectionData` object state. Then parameters can be hardcoded again, and one can use the ellipsis to overwrite default values with nothing.

takluyver:
I don't think ellipsis is a good choice for 'remove the default' - it looks more like 'this is implicit' or 'this comes from somewhere else' to my mind. I may have misunderstood what you were suggesting when we spoke earlier. :slightly_smiling_face: 

If you want a built-in object to mean 'don't use the default', I'd suggest it has to be either None or False. We could use some other sentinel to mean 'not passed, use from another source'. That could be an object we define (at the simplest, `unspecified = object()`), especially if we're not expecting people to explicitly pass it. Or possibly Ellipsis, or even just a short string.

extra_data/caldata.py
```diff
@@ -0,0 +1,1016 @@
...
+    detector or not, correspondingly called the illuminated and dark
+    conditions. Typically the illuminated condition is a superset of the
+    dark condition.
+
+    Not all implementations for semiconductor detectors inherit from
+    this type, but only those that make this distinction such as AGIPD
+    and LPD.
+    """
+
+    dark_calibrations = set()
```
D4vidH4mm3r, line 618: Comparison: in `calng`, each detector-specific subclass can define arbitrary methods to generate conditions and the attribute `_constants_need_conditions` explicitly maps from constants (or "calibrations") to condition methods. One can of course translate between that and `dark_calibrations` / `illuminated_calibrations`, so it's in the end a matter of modeling things nicely.
Personally, I find the explicit map easier to read (to figure out what goes into getting an `Offset`, I see that I need to look at `self.dark_condition`). Additionally, it avoids the introduction of a "dual condition" subclass.

philsmt: I'm still somewhat hopeful we can get rid of the split between dark and illuminated condition. It makes the interaction unnecessarily complicated and with tolerances, there's a better and native way to express this distinction anyway (e.g. dark constants are defined with `source_energy = 0 +/ inf`).

Apart from that, given there are de facto only those two conditions (and only for some detectors anyway), I'm not sure it warrants an entire additional layer of abstraction across all detectors (said mapping). With a subclass, I can make the distinction easily disappear later on.

extra_data/caldata.py
```diff
@@ -0,0 +1,1016 @@
...
+        Args:
+            metadata (CCVMetadata, optional): CCV metadata to load
+                constant data for, may be None to query metadata.
+            *args, **kwargs: If metadata is omitted, any additional
+            positional and keyword arguments are passed on to
+            .metadata().
+
+        Returns:
+            (dict of dict): Nested dictionary of PDU index and
+                calibration name to CCV data.
```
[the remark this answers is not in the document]
D4vidH4mm3r: Could `.ndarray()` be tied to a CCV object (modeling the situation in CalCat)? As it stands, a user can either first obtain a `CCVMetadata` via `.metadata` and then pass it for `.ndarray` or omit this, leaving the `CalibrationData` to implicitly create one. How about having the auxiliary class model instead a `CalibrationConstantVersion` and have it contain metadata and a way to access the actual data like `ndarray`?

philsmt:
I've pushed a preliminary version of the rework. The `CorrectionData` object is now designed to be immutable and contain the operating condition as its state. Any change to it should go through the `CorrectionData.replace()` method (modelled after `inspect.Signature.replace()`).

As it happens, this made large sections of the code considerablt easier lost more than 200 lines. The `OperatingCondition` type became superfluous and the condition parameters can now be hardcoded as function argument in exactly one function (the initializer).

Couple of minor changes related to past comments:
* `CorrectionData.ndarray()` now returns an `ndarray` for a specific module and calibration, there's a new method `ndarray_map` to get nested dicts of `ndarray`.
* The detector components actually only report "frames", which are somewhat akin to pulses yet not memory cells. As I need the latter for constants, I've added the method `_get_memory_cell_count` to the component types. I'm not sure yet it makes sense there, also happy to move it back to this module. 
* I've tried to complete the parameter detection for AGIPD, DSSC and JF. For DSSC however, I cannot replicate the extended operating condition yet.
* The `calibration_client` is wrapped now in a `CalCatApi` object for easier testability and caching request results independent of `CorrectionData` objects.

extra_data/calibration.py
```diff
@@ -0,0 +1,1090 @@
...
+
+    dark_parameters = ['Sensor Bias Voltage', 'Pixels X', 'Pixels Y',
+                       'Memory cells', 'Acquisition rate', 'Gain setting',
+                       'Gain mode', 'Integration time']
+    illuminated_parameters = dark_parameters + ['Source energy']
+
+    def __init__(self, detector_name, modules=None, client=None,
+                 event_at=None, snapshot_at=None,
+                 sensor_bias_voltage=300.0, memory_cells=202, pixels_x=512,
+                 pixels_y=128, acquisition_rate=4.5, gain_setting=None,
```
takluyver, line 710: `acquisition_rate` is not being stored on the instance at present.

extra_data/calibration.py
```diff
@@ -0,0 +1,1090 @@
...
+        Returns:
+            (dict) Operating condition for use in CalCat API.
+        """
+
+        return {'parameters_conditions_attributes': [
+            {'parameter_id': self.parameter_id(k), 'value': str(v)}
+            for k, v in condition.items()]}
+
+    @lru_cache()
+    def detector(self, detector_name):
```
takluyver, line 146: I came across [a blog post](https://rednafi.github.io/reflections/dont-wrap-instance-methods-with-functoolslru_cache-decorator-in-python.html) the other day pointing out that using `lru_cache` on regular methods like this prevents instances from being freed - the cache is kept on the class, and keeps references to each instance where the method has been called. The suggested workaround was to wrap the methods in `__init__`:

```python
self.calculate = lru_cache()(self._calculate)
```

Probably not a big deal here, because we're unlikely to create thousands of `CalCatApi` instances, but it might be best to avoid the issue anyway, in case we change how the code works or copy this pattern somewhere else.

extra_data/calibration.py
```diff
@@ -0,0 +1,1090 @@
...
+        Returns:
+            (Path or None) Location of caldb store or
+                None if not available.
+        """
+
+        if not hasattr(CorrectionData, '_caldb_root'):
+            if getenv('SASE'):
+                # ONC
+                CorrectionData._caldb_root = Path('/common/cal/caldb_store')
+            elif re.match(r'^max-(.+)\.desy\.de$', socket.getfqdn()):
```
takluyver, line 412: I'd be inclined to just check if the expected directories exist, rather than looking at environment variables and host names. E.g. inside a container, we may have a different hostname, even if the `/gpfs` paths are accessible.

I suspect we'll also want a way to override it, e.g. to use the test CalCat files at `/gpfs/exfel/d/cal_tst/caldb_store` on Maxwell.

extra_data/calibration.py
```diff
@@ -0,0 +1,1090 @@
...
+                metadata for.
+            condition (dict): Mapping of parameter name to value.
+            modules (Collection of int or None): List of module numbers
+                or None for all (default).
+            event_at (datetime, date, str or None): Time at which the
+                CCVs should have been valid or None for now (default).
+            snapshot_at (datetime, date, str or None): Time of database
+                state to look at or None for now (default).
+            metadata (dict or None): Mapping to fill for results or
+                None for a new dictionary (default).
```
takluyver, line 218: The 'pass in a container to put the output into' pattern seems a little bit ugly. I think I can see why it's there, to simplify combining the retrieved metadata for dark & illuminated conditions for AGIPD & LPD, but I think that can be done simply enough outside this method:

```python
for pdu_idx, pdu_data in new_metadata.items()
    prev_metadata.setdefault(pdu_idx, {}).update(pdu_data)
```

Or invert the nesting so the outer layer is constant types, and then it's just a plain `.update()`. :slightly_smiling_face:

extra_data/calibration.py
```diff
@@ -0,0 +1,1090 @@
...
+        detector_sources = [data[source] for source in source_to_modno.keys()]
+
+        if modules is None:
+            modules = sorted(source_to_modno.values())
+
+        creation_date = cls._determine_data_creation_date(data)
+
+        # Create new CorrectionData object.
+        corrdata = cls(detector_name, modules, client,
+                       creation_date, creation_date)
```
takluyver, line 565: This is using `creation_date` for both `event_at` and `snapshot_at`, which IIUC means it will only consider CCVs injected before the run was taken. I'd expect `snapshot_at=None` to be the default, so it would also consider CCVs injected afterwards.

Tracing through the code paths used to retrieve constants for the offline correction at the moment, I *think* it's setting `snapshot_at` for identifying the PDUs, and then unsetting it again before actually looking up the CCV. Specifically, here:

https://git.xfel.eu/karaboDevices/calibrationDBRemote/-/blob/3cce401db508d690e4998eab87a588b5bba52034/src/calibrationDBRemote/calibration_db_remote.py#L395

(And figuring that out reminds me that we should really move this forwards and get rid of calibrationDBRemote and iCalibrationDB ASAP)

takluyver:
If we're dropping Python 3.6 support - which I'm OK with - we'll also need to drop it from the CI: https://github.com/European-XFEL/EXtra-data/blob/664d1830de14ed1435216b305def5545337dcc99/.github/workflows/tests.yml#L14

That might be a good time to add 3.10. If you do, it needs quotation marks `"3.10"` otherwise it's a float and tries to install Python 3.1. I've been bitten by that a few times.

takluyver:
So looking at the code, I'm guessing it would be used something like this?

```python
consts = LPD_CalibrationData.from_data(run).ndarray_map()
consts[0]['Offset']  # Offset constant array for module 0
```

extra_data/calibration.py
```diff
@@ -0,0 +1,1090 @@
...
+    def calibrations(self):
+        """Compatibility with CorrectionData."""
+
+        return self.dark_calibrations | self.illuminated_calibrations
+
+    @property
+    def parameters(self):
+        """Compatibility with CorrectionData."""
+
+        return self.dark_parameters + self.illuminated_parameters
```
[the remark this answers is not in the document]
philsmt: Thanks for spotting this.

extra_data/calibration.py
philsmt: Very true, thanks for spotting the use case of test data.

extra_data/components.py
```diff
@@ -426,6 +427,23 @@ def _select_pulse_indices(pulses, counts):
...
+        detector_source: SourceData
+          Fast data source of a detector module.
+
+        Returns
+        -------
+        int
+          Number of used memory cells.
+        """
+
+        return detector_source['image.cellId'][0].shape[0]
```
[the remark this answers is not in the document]
philsmt: I'm wondering know if we can safely assume there can never be a run with changing memory cells. I suppose the DAQ would not take that without reconfiguring?

takluyver:
Just to record, we'll try this API out internally in calibration code ([internal MR](https://git.xfel.eu/detectors/pycalibration/-/merge_requests/738)) before we land it in EXtra-data, so this PR will stay as a draft for now.

philsmt:
I'm closing this PR now. It can still be useful for reference, but with EXtra we have a more fitting place once this API reaches maturity and becomes public.

PR #264, opened 2021-12-15
Use a metaclass to make FileAccess behave as expected

takluyver:
@philsmt - this came from our discussion yesterday.

This uses a metaclass so that attempting to instantiate `FileAccess()` will look in the registry for an existing instance, and not reinitialise it if it exists. Likewise, unpickling a FileAccess object won't overwrite the state (caches) of an object that already exists.

I had a go at implementing the other option with the `.get()` classmethod as well - branch [`file-access-no-reinit`](https://github.com/European-XFEL/EXtra-data/compare/file-access-no-reinit) - but I think I prefer this one. It's too easy with the `.get()` option to use the normal constructor instead, and get duplicate objects. I could make it an error to call `FileAccess()` directly, but I'm reluctant to do that in case someone is using it outside our code; it's not documented but it's not hard to find.

### DIFF
--- a/extra_data/file_access.py
+++ b/extra_data/file_access.py
@@ -94,7 +94,22 @@ def init_open_files_limiter():
 open_files_limiter = init_open_files_limiter()
 
 
-class FileAccess:
+class MetaFileAccess(type):
+    # Override regular instance creation to check in the registry first.
+    # Defining __new__ on the class is not enough, because an object retrieved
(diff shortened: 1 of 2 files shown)


philsmt:
Thanks, LGTM!

This is indeed the first time I see the `__reduce__` machinery for `pickle` in action. A nice example one may certainly go back to if needed.

PR #268, opened 2022-01-06
Virtual overview edits

tmichela:
@takluyver I just had a go with generating virtual overview files. it looks like it's working fine for me. I've added some of the changes we discussed last time:
- do no check for mtime
- write overview file to /usr

I'd ideally like to deploy autogeneration of these files before the next user run. Luckily the next Karabo release has an updated numpy version making easier to deploy extra-data in that environment \o/

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,3 @@
 [run]
 omit = */tests/*
-concurrency = "multiprocessing"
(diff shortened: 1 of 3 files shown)


takluyver:
Thanks, these changes made sense. We can carry on on #69.

PR #269, opened 2022-01-10
Write virtual overview files in temp location & rename

takluyver:
So a partially written overview file is never visible where EXtra-data looks for it to read.

cc @tmichela this relates to what we were just discussing.


tmichela:
LGTM 👍

PR #274, opened 2022-01-28
Track files by source

### LINKED ISSUE #273: data='all' incompatible with using virtual overview files for runs
I realised while talking to @tmichela yesterday that reading runs from a virtual dataset is going to fail if you use `data='all'`, because it can't properly deselect *some* sources from a file and then add them back later if it still has that file for other data.

This is essentially the same as https://github.com/European-XFEL/karabo_data/issues/131 . We mitigated it by dropping files where all sources were deselected (especially useful for big detector data), but didn't properly solve it. Virtual overview files mean all the sources from a run are in one file, so dropping files with no sources we want is much less effective.

We can't directly test this until we have some runs where the uncorrected files are not copied to proc. But I can illustrate an analogous situation, because when Fabio copied an MID run for my demo on Tuesday, he copied the full folder for `proc` but only the slow data for `raw`. So we have an example where raw is a subset of proc (the reverse of what will happen). If I make virtual overview files for those, then:

```
In [2]: from extra_data import H5File                                           

In [3]: r33_proc = H5File('xmpl-33-proc.h5')                                    

In [4]: r33_raw = H5File('xmpl-33-raw.h5')                                      

In [5]: r33_partial = r33_proc.deselect([(src, '*') for src in r33_raw.all_sources])                                                                    

In [7]: r33_raw.union(r33_partial)                                              
ValueError                                Traceback (most recent call last)
<ipython-input-7-39fb21d66ee8> in <module>
----> 1 r33_raw.union(r33_partial)

~/Code/EXtra-data/extra_data/reader.py in union(self, *others)
    616             files, selection=selection, train_ids=train_ids,
    617             inc_suspect_trains=self.inc_suspect_trains,
--> 618             is_single_run=same_run,
    619         )
    620 

~/Code/EXtra-data/extra_data/reader.py in __init__(self, files, selection, train_ids, ctx_closes, inc_suspect_trains, is_single_run)
    108 
    109         # Throw an error if we have conflicting data for the same source
--> 110         self._check_source_conflicts()
    111 
    112         self.control_sources = frozenset(self.control_sources)

~/Code/EXtra-data/extra_data/reader.py in _check_source_conflicts(self)
    978             raise ValueError("{} sources have conflicting data "
    979                              "(same train ID in different files): {}".format(
--> 980                 len(sources_with_conflicts), ", ".join(sources_with_conflicts)
    981             ))
    982 

ValueError: 30 sources have conflicting data (same train ID in different files): MID_XTD6_IMGPI/CAM/BEAMVIEW, MID_EXP_AGIPD1M/MOTOR/Q1M2, MID_EXP_AGIPD1M/MOTOR/Q2M1, ...
```

I'm working on a solution to this.

takluyver:
This allows you to deselect some sources from a file, and then union the result with data which also contains those sources. This will be necessary once we make virtual overview files and don't copy all the raw data to proc, because we need to deselect part of the raw data.

Previously, the set of open files and the selected sources were tracked separately, so an included source would look at all files in the pool and see if they had that source. This change tracks both the files and keys selected in the `SourceData` interface, so you can have a source that doesn't look at all open files. I tried a couple of ideas, but this one seems fairly elegant, and adds minimal complexity.

As a side effect, a DataCollection now keeps concrete SourceData objects internally, rather than creating them on demand.

Closes #273.

TBD: I broke a couple of small things which were never documented, but were used in tests - the private `._source_index` attribute, and passing an empty dict instead of an empty set to select keys. They'd both be easy to add a compatibility hack for, so I could do that just to be extra cautious. I think it's possible that some people discovered and used `._source_index`.

I hope no-one was manually instantiating `DataCollection(...)` with the selection parameter, because that would be less easy to accommodate.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -635,7 +635,9 @@ class DataCollection:
                 if source not in self.all_sources:
                     raise SourceNameError(source)
 
-                if in_keys is not None and not isinstance(in_keys, set):
+                # Empty dict was accidentally allowed and tested; keep it
+                # working just in case.
+                if in_keys is not None and not isinstance(in_keys, (set, dict)):
                     raise TypeError(
                         f"keys in selection dict should be a set or None (got "
                         f"{in_keys!r})"
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -41,7 +41,6 @@ from .read_machinery import (
     by_index,
     select_train_ids,
     split_trains,
-    union_selections,
     find_proposal,
     glob_wildcards_re,
 )
@@ -80,37 +79,13 @@ class DataCollection:
     for a single file or :func:`RunDirectory` for a directory.
     """
     def __init__(
-            self, files, selection=None, train_ids=None, ctx_closes=False, *,
+            self, files, sources_data=None, train_ids=None, ctx_closes=False, *,
             inc_suspect_trains=True, is_single_run=False,
     ):
         self.files = list(files)
         self.ctx_closes = ctx_closes
         self.inc_suspect_trains = inc_suspect_trains
         self.is_single_run = is_single_run
-
-        # selection: {source: set(keys)}
-        # None as value -> all keys for this source
-        if selection is None:
-            selection = {}
-            for f in self.files:
-                selection.update(dict.fromkeys(f.control_sources))
-                selection.update(dict.fromkeys(f.instrument_sources))
-        self.selection = selection
-
-        self.control_sources = set()
-        self.instrument_sources = set()
-        self._source_index = defaultdict(list)
-        for f in self.files:
-            self.control_sources.update(f.control_sources.intersection(selection))
-            self.instrument_sources.update(f.instrument_sources.intersection(selection))
-            for source in (f.control_sources | f.instrument_sources):
-                self._source_index[source].append(f)
-
-        # Throw an error if we have conflicting data for the same source
-        self._check_source_conflicts()
-
-        self.control_sources = frozenset(self.control_sources)
-        self.instrument_sources = frozenset(self.instrument_sources)
 
         if train_ids is None:
             if inc_suspect_trains:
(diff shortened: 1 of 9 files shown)


extra_data/reader.py
```diff
@@ -199,6 +205,20 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    @property
+    def selection(self):
+        # This was previously a regular attribute, which code may have relied on.
+        return {src: srcdata.sel_keys for src, srcdata in self._sources_data.items()}
+
+    @property
+    def _source_index(self):
+        warn(
+            "DataCollection._source_index will be removed. "
+            "Contact da-support@xfel.eu if you need to discuss alternatives.",
```
takluyver, line 217: If anyone is using this, they can just replace `run._source_index[source]` with `run[source].files`. We could just put that in the warning. I've suggested that people contact us partly so we can get a bit of a sense if anyone is using it, and partly so we can explain about not relying on things that `_start_with_underscore`.

extra_data/sourcedata.py
```diff
@@ -87,3 +95,109 @@ def keys(self, inc_timestamps=True):
...
+        else:
+            # Both the new and current keys are specific: take the intersection.
+            # The check above should ensure this never results in an empty set,
+            # but
+            new_keys = self.sel_keys & keys
+            assert new_keys
+
+        res = copy(self)
+        res.sel_keys = new_keys
+        return res
```
takluyver, line 176: These 3 methods that return a new `SourceData` object all use this pattern of making a copy and modifying it. Similar methods in DataCollection & KeyData work by creating a new object and passing all the unchanged fields back to it. I'm not sure which one I like better.

Using `copy()` tends to be fewer lines of code, and if we add a new attribute for `SourceData` objects, we don't have to change every such method to copy it across. But creating a new object each time is more explicit, and you can reuse any type conversions or checks in the `__init__` method. :shrug:

extra_data/tests/test_reader_mockdata.py
```diff
@@ -511,7 +511,7 @@ def test_select(mock_fxe_raw_run):
         sel.select('FXE_DET_LPD1M-1/DET/0CH0:xtdf', '*') \
            .keys_for_source('FXE_DET_LPD1M-1/DET/0CH0:xtdf')
     assert sel.keys_for_source('FXE_DET_LPD1M-1/DET/0CH0:xtdf') == \
-        sel.select({'FXE_DET_LPD1M-1/DET/0CH0:xtdf': {}}) \
+        sel.select({'FXE_DET_LPD1M-1/DET/0CH0:xtdf': set()}) \
```
takluyver, line 514: The previous implementation allowed an empty dict in place of an empty set, but now that's an error. I hope no-one was using that, but we could allow it if necessary.

extra_data/sourcedata.py
```diff
@@ -87,3 +95,109 @@ def keys(self, inc_timestamps=True):
...
+
+    def select_trains(self, trains) -> 'SourceData':
```
tmichela, line 178: Can you document this method?

extra_data/sourcedata.py
```diff
@@ -87,3 +95,109 @@ def keys(self, inc_timestamps=True):
...
+        res.train_ids = tids
+        res.files = [
+            f for f in self.files
+            if f.has_train_ids(tids, self.inc_suspect_trains)
+        ] or [self.files[0]]
+        # ^ Keep 1 file, even if 0 trains selected, to get keys, datatypes, etc.
+
+        return res
+
+    def union(self, *others) -> 'SourceData':
```
tmichela, line 192: Also adding docstring here?

tmichela (approved):
Thanks for the quick fix! Just some doc nitpicking else LGTM

extra_data/tests/test_reader_mockdata.py
takluyver: OK, done. I haven't added a warning, just let it carry on working.

PR #278, opened 2022-02-03
select_trains() and split_trains() methods for detector data classes

### LINKED ISSUE #255: split_trains method for xtdf detector classes
Something I noticed while working on code for FXE - it would be useful to be able to split the `LPD1M` object into chunks, like we can with `DataCollection` and `KeyData`. In particular, a `frames_per_part` option would be useful when you want to limit how much data a chunk contains without needing to know the number of frames per train. My idea would still round it to a whole number of trains in each chunk, though.

In a similar vein, `KeyData.split_trains` could have an option like `bytes_per_part`. :thinking:

takluyver:
`DataCollection` and `KeyData` both have methods to select a subset of data by trains, and to split the data along the trains dimension. Various examples would be simpler if the classes for multi-module detector data offered the same API. At present, you have to split your `DataCollection` and call e.g. `AGIPD1M(chunk)` on each piece. This API would let you call `AGIPD1M(run)` up front, and then split the result, which would make various examples more straightforward.

TBD: A `frames_per_part` option for splitting? This would let you split based on a maximum number of frames in a chunk, rather than a maximum number of trains, which would make it easier to write code that stays within a memory budget if the number of frames per train changes.

Closes #255

### DIFF
--- a/extra_data/tests/test_components.py
+++ b/extra_data/tests/test_components.py
@@ -276,6 +276,20 @@ def test_select_trains(mock_fxe_raw_run):
     assert arr.shape == (16, 2, 128, 256, 256)
 
 
+def test_split_trains(mock_fxe_raw_run):
+    run = RunDirectory(mock_fxe_raw_run)
+    det = LPD1M(run.select_trains(np.s_[:20]))
+    assert len(det.train_ids) == 20
+
+    parts = list(det.split_trains(parts=4))
+    assert len(parts) == 4
+    assert [len(p.train_ids) for p in parts] == [5, 5, 5, 5]
+    assert sum([p.train_ids for p in parts], []) == det.train_ids
+
+    arr = parts[-1].get_array('image.data', pulses=np.s_[:1])
+    assert arr.shape == (16, 5, 1, 256, 256)
+
+
 def test_iterate(mock_fxe_raw_run):
     run = RunDirectory(mock_fxe_raw_run)
     det = LPD1M(run.select_trains(by_index[:2]))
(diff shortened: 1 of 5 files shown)


tmichela:
Maybe worth adding the additional methods to the reference doc, else LGTM

takluyver:
I've added the `frames_per_part=` parameter, which I hope will be useful for memory budgeting.

The splitting with this is somewhat cruder, because we don't assume that there's a consistent number of frames per train, so the chunks won't necessarily be balanced. We could have a cleverer implementation when frames-per-train is constant, but having two ways to do the same thing is extra complexity.

extra_data/tests/test_components.py
```diff
@@ -265,6 +265,64 @@ def test_get_dask_array_jungfrau(mock_jungfrau_run):
...
+    assert [len(p.train_ids) for p in parts] == [1] * 20
+
+    # trains_per_part cuts off before frames_per_part
+    parts = list(det.split_trains(trains_per_part=3, frames_per_part=1024))
+    assert [len(p.train_ids) for p in parts] == ([3] * 6) + [2]
+
+    # trains_per_part cuts off before frames_per_part
+    parts = list(det.split_trains(trains_per_part=3, frames_per_part=1024))
+    assert [len(p.train_ids) for p in parts] == ([3] * 6) + [2]
+
```
tmichela, line 307:
```suggestion
    # trains_per_part cuts off before frames_per_part
    parts = list(det.split_trains(trains_per_part=3, frames_per_part=1024))
    assert [len(p.train_ids) for p in parts] == ([3] * 6) + [2]

```

tmichela (approved):
LGTM

PR #281, opened 2022-02-23
Allow open_run() to open runs with 'all' that have no proc data

JamesWrigley:
Currently `open_run(_, _, data="all")` will fail if the proc data does not exist. This PR tweaks the behaviour to allow for proc data not existing, and raising a warning if it doesn't (not completely sure if this is necessary).

Rationale:
- My (possibly biased and strange) intuition is that since proc data might not exist, `data="all"` shouldn't require it. I guess I think of the meaning of 'all' as 'everything that is available' rather than 'everything that could be available'.
- IIUC the future plan is to change the default to `data="all"`. Since runs might not have any proc data I think it makes sense for `open_run()` to be able to handle these by default without the user having to set `data="raw"`.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1481,22 +1481,27 @@ def open_run(
             proposal=proposal, run=run, include=include,
             file_filter=file_filter, inc_suspect_trains=inc_suspect_trains)
 
-        # Create separate data collections for raw and proc.
-        raw_dc = open_run(**common_args, data='raw')
-        proc_dc = open_run(**common_args, data='proc')
+        # Open the raw data
+        dc = open_run(**common_args, data='raw')
 
-        # Deselect to those raw sources not present in proc.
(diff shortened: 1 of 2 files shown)


philsmt:
Amazing how this completely slipped my mind. I'm sorry, and thanks for fixing it. LGTM!

Tangent: I'm not sure we should make `data='all'` the default, even though I'd like to, as it potentially _soft breaks_ existing analysis code that assumes to get raw data, but might suddenly get processed data. That being said, the CAL team would like to completely change source names of processed data in the mid-term, _maybe_ such a change could be slipped in at that point.

takluyver:
Thanks, this LGTM too.

I would like to change the default, but as Philipp says, it's liable to break existing code. I've been vaguely wondering about 'virtually' renaming corrected data, once we've chosen the new source names, presenting older data as though it had been written with the new names. But it sounds pretty fiddly to get that right. :shrug:

PR #282, opened 2022-03-01
Allow .select() to take a list of just source patterns

takluyver:
A tweak to reduce the amount of punctuation you need:

```python
# Currently:
sel = run.select([
    ('MID_EXP_FASTADC/ADC/DESTEST:channel_2.output', '*'),   # DES
    ('MID_XTD1_UND/DOOCS/ENERGY', '*'),  # Undulators
    ('MID_DET_AGIPD1M-1/DET/*CH0:xtdf', '*'),  # AGIPD
], require_all=True)

# With this PR:
sel = run.select([
    'MID_EXP_FASTADC/ADC/DESTEST:channel_2.output',   # DES
    'MID_XTD1_UND/DOOCS/ENERGY',  # Undulators
    'MID_DET_AGIPD1M-1/DET/*CH0:xtdf',  # AGIPD
], require_all=True)
```

When we added selection to control what the `.trains()` iterator loads and yields, you often wanted to select both sources and keys. The addition of the `require_all` option made it also useful for aligning data by trains, but for this purpose it's often sufficient to select only using source names, and keep all keys.

### DIFF
--- a/extra_data/tests/test_reader_mockdata.py
+++ b/extra_data/tests/test_reader_mockdata.py
@@ -496,6 +496,16 @@ def test_select(mock_fxe_raw_run):
     for source, source_data in data.items():
         assert set(source_data.keys()) == {'image.pulseId', 'metadata'}
 
+    sel_by_list = run.select([
+        ('*/DET/*', 'image.pulseId'),
+        'FXE_XAD_GEC/CAM/*',
+    ])
+    assert 'SPB_XTD9_XGM/DOOCS/MAIN' not in sel_by_list.control_sources
+    assert 'FXE_DET_LPD1M-1/DET/0CH0:xtdf' in sel_by_list.instrument_sources
+    assert sel_by_list['FXE_DET_LPD1M-1/DET/0CH0:xtdf'].keys() == {'image.pulseId'}
+    assert 'FXE_XAD_GEC/CAM/CAMERA_NODATA' in sel_by_list.control_sources
+    assert 'FXE_XAD_GEC/CAM/CAMERA_NODATA:daqOutput' in sel_by_list.instrument_sources
+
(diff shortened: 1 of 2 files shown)


extra_data/tests/test_reader_mockdata.py
```diff
@@ -496,6 +496,16 @@ def test_select(mock_fxe_raw_run):
...
+    sel_by_list = run.select([
+        ('*/DET/*', 'image.pulseId'),
+        'FXE_XAD_GEC/CAM/*',
+    ])
+    assert 'SPB_XTD9_XGM/DOOCS/MAIN' not in sel_by_list.control_sources
+    assert 'FXE_DET_LPD1M-1/DET/0CH0:xtdf' in sel_by_list.instrument_sources
+    assert sel_by_list['FXE_DET_LPD1M-1/DET/0CH0:xtdf'].keys() == {'image.pulseId'}
+    assert 'FXE_XAD_GEC/CAM/CAMERA_NODATA' in sel_by_list.control_sources
+    assert 'FXE_XAD_GEC/CAM/CAMERA_NODATA:daqOutput' in sel_by_list.instrument_sources
+
```
philsmt, line 508: One might add a check here that all keys for one of the `FXE_XAD_GEC/CAM/*` is present.

philsmt (approved):
Thanks! I've tripped over this in the field myself, and this will add a fair bit of convenience. That being said, naturally it is recommend to narrow down keys as much as possible in particular with the `.trains()` iterator to speed-up I/O.

Aside from small test nitpick, LGTM!

extra_data/tests/test_reader_mockdata.py
takluyver: Good point, added a check.

PR #288, opened 2022-03-15
introduce keep_dims flag to preserve the cell dimension for the selec…

kakhahmed:
MID requested correcting one cellId for AGIPD in the offline calibration.

As mentioned in this MR: https://git.xfel.eu/detectors/pycalibration/-/merge_requests/640

There was an error in correcting one cellId. This was a result of wrong expected dimensions.

When the CORR data is read for one train using `train_from_id` to plot AGIPD images, all arrays were returned without preserving the `1` for the cell dimensions.

Based on @philsmt 's advice. I added `keep_dims` flags for `trains()`, `train_from_id`, `train_from_index` for data collections and key data to have the option in preserving the cell dimension and slice the required data even though `count == 1`

I have added some tests as well for having `keep_dims = True`, I tried to only use already available tests and add test functions based on it.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -296,7 +296,7 @@ class DataCollection:
         return False
 
     def trains(self, devices=None, train_range=None, *, require_all=False,
-               flat_keys=False):
+               flat_keys=False, keep_dims=False):
         """Iterate over all trains in the data and gather all sources.
 
         ::
@@ -327,6 +327,10 @@ class DataCollection:
             iteration indexed by source and then key. True returns a
             flat dictionary indexed by (source, key) tuples.
 
+        keep_dims: bool
+            False (default) based on count index or slice the data.
+            True returns a sliced data to preserve the dimensions.
+
         Yields
         ------
 
@@ -341,9 +345,10 @@ class DataCollection:
         if train_range is not None:
             dc = dc.select_trains(train_range)
         return iter(TrainIterator(dc, require_all=require_all,
-                                  flat_keys=flat_keys))
+                                  flat_keys=flat_keys, keep_dims=keep_dims))
 
-    def train_from_id(self, train_id, devices=None, *, flat_keys=False):
+    def train_from_id(
+        self, train_id, devices=None, *, flat_keys=False, keep_dims=False):
         """Get train data for specified train ID.
 
         Parameters
@@ -357,6 +362,9 @@ class DataCollection:
         flat_keys: bool
             False (default) returns a nested dict indexed by source and then key.
             True returns a flat dictionary indexed by (source, key) tuples.
+        keep_dims: bool
+            False (default) based on count index or slice the data.
+            True returns a sliced data to preserve the dimensions.
 
         Returns
         -------
@@ -406,7 +414,7 @@ class DataCollection:
                     continue
 
                 path = '/INSTRUMENT/{}/{}'.format(source, key.replace('.', '/'))
-                if count == 1:
+                if count == 1 and not keep_dims:
                     source_data[key] = file.file[path][first]
                 else:
                     source_data[key] = file.file[path][first : first + count]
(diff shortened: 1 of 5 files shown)


philsmt:
Adding some context here: I believe the original branch for `if count == 1` was meant to distinguish between pulse-resolved and train-resolved fast data (say, AGIPD vs pnCCD), but unfortunately it can lead to the expected pulse axis be dropped when a pulse-resolved source only contains that one entry. While we could try to be smart and check for XTDF sources, this relies on the assumption only those may have several entries per pulse, something I already break with the REMI reconstruction notebook.

Hence, this flag allows at least to express the intent to preserve dimensions.

kakhahmed:
hmm, I dont understand the reason for the failed codecov/project

extra_data/reader.py
```diff
@@ -327,6 +327,10 @@ def trains(self, devices=None, train_range=None, *, require_all=False,
             iteration indexed by source and then key. True returns a
             flat dictionary indexed by (source, key) tuples.
 
+        keep_dims: bool
+            False (default) based on count index or slice the data.
+            True returns a sliced data to preserve the dimensions.
```
takluyver, line 332:
```suggestion
            False (default) drops the first dimension when there is
            a single entry. True preserves this dimension.
```

takluyver:
I wouldn't worry too much about codecov - my experience is that it often says coverage has decreased when it hasn't actually, so I'm used to ignoring it.

Numpy has a similar-ish parameter on e.g. [`np.mean()`](https://numpy.org/doc/stable/reference/generated/numpy.mean.html) called `keepdims`, i.e. without an underscore. I might be inclined to copy that naming, but it's not something I feel strongly about.

extra_data/tests/test_reader_mockdata.py
```diff
@@ -168,6 +188,17 @@ def test_iterate_spb_raw_run(mock_spb_raw_run):
     assert device in data
     assert data[device]['data.image.pixels'].shape == (1024, 768)
 
+
+def test_iterate_spb_raw_run_keep_dims(mock_spb_raw_run):
+    run = RunDirectory(mock_spb_raw_run)
+    trains_iter = run.trains(keep_dims=True)
```
takluyver, line 194: Is it worth doing a `.select()` in here? Even with no real data, creating all the arrays and making all the calls into HDF5 might be a bit slow.

extra_data/reader.py
kakhahmed: Thanks this is indeed more understandable.

philsmt:
> Numpy has a similar-ish parameter on e.g. [`np.mean()`](https://numpy.org/doc/stable/reference/generated/numpy.mean.html) called `keepdims`, i.e. without an underscore. I might be inclined to copy that naming, but it's not something I feel strongly about.

I could not easily find an instance in `extra_data` of such a naming style, so I'd argue it's less confusing to maintain the same style throughout instead of adopting `numpy`'s for only a single argument.

kakhahmed:
> I wouldn't worry too much about codecov - my experience is that it often says coverage has decreased when it hasn't actually, so I'm used to ignoring it.
> 
> Numpy has a similar-ish parameter on e.g. [`np.mean()`](https://numpy.org/doc/stable/reference/generated/numpy.mean.html) called `keepdims`, i.e. without an underscore. I might be inclined to copy that naming, but it's not something I feel strongly about.

So as you can see from the commits I was a bit uncertain what I prefer. But it looked weird having `flat_keys` and `require_all` and `keepdims` for the same function.

kakhahmed:
> > I wouldn't worry too much about codecov - my experience is that it often says coverage has decreased when it hasn't actually, so I'm used to ignoring it.
> > Numpy has a similar-ish parameter on e.g. [`np.mean()`](https://numpy.org/doc/stable/reference/generated/numpy.mean.html) called `keepdims`, i.e. without an underscore. I might be inclined to copy that naming, but it's not something I feel strongly about.
> 
> So as you can see from the commits I was a bit uncertain what I prefer. But it looked weird having `flat_keys` and `require_all` and `keepdims` for the same function.

And I see that @philsmt already said the same thing 3 hours ago.

PR #289, opened 2022-03-15
Avoid a race condition opening/closing HDF5 files

takluyver:
This should hopefully fix the race condition described in https://github.com/European-XFEL/pasha/pull/13 .

I think there's still a separate race condition with `open_files_limiter` - if the timing is right, the file may be open when `open_files_limiter` thinks it's closed. But the limiter aims for half the process limit on the number of files, so it's unlikely to be a big problem if files sometimes leak. There may well also be race conditions if FileAccess objects are created/unpickled in different threads. We may need to tackle these at some point, but they're harder to solve than this one. I'm reluctant to focus too much on threading support, because only a single thread can access HDF5 files at a time anyway.

### DIFF
--- a/extra_data/file_access.py
+++ b/extra_data/file_access.py
@@ -180,10 +180,12 @@ class FileAccess(metaclass=MetaFileAccess):
     @property
     def file(self):
         open_files_limiter.touch(self.filename)
-        if self._file is None:
-            self._file = h5py.File(self.filename, 'r')
+        # Local var to avoid a race condition when another thread calls .close()
+        file = self._file
+        if file is None:
+            file = self._file = h5py.File(self.filename, 'r')
 
-        return self._file
+        return file
 
     @property
     def valid_train_ids(self):


philsmt:
It's scary to see a race condition happening so easily, in particular because I'd argue this pattern of using properties is fairly common.

While this PR will prevent `None` from being returned, wouldn't the returned `h5py.File` object still be closed and no longer usable?

takluyver:
`FileAccess.close()` doesn't explicitly close the h5py `File` object - it just deletes the reference to it, and lets Python's reference counting deal with the rest. If another thread gets a reference to it while that is happening, then the File will stay open until that reference is dropped as well.

philsmt:
Of course, I'm sorry. I did not look close enough at `FileAccess.close()`. I was actually wondering whether some reference counting might be the right solution when posting the comment, but that's what it's already doing implicitly. So when thread B closes the file while thread A one just opened it, it'll stay open until thread A closes all references. Which quite likely is identical to the lifetime of its `DataCollection`?
Now when (in the case of `pasha`) thread B arrives in its kernel and again opens the file, will it be a new file object, with the old one also being around? In other words, do we open `N` files for `N` threads?

takluyver:
>  So when thread B closes the file while thread A one just opened it, it'll stay open until thread A closes all references. Which quite likely is identical to the lifetime of its DataCollection?

If thread A sets `fileaccess._file = <new file object>` last, then that reference will stick around as part of the DataCollection. It should also get used by other threads when they try to access the same file.

On the other hand, if thread B sets `fileaccess._file = None` last, then thread A's reference is just a local variable, and will probably be dropped again as soon as whatever statement using `fa.file` completes.

> In other words, do we open N files for N threads?

So long as we have one `FileAccess` object, there should only be one long-lived h5py `File` object, the one stored at `fileaccess._file`. There may briefly be duplicates but these should be dropped quickly.

We try to ensure there is only one `FileAccess` object per file, per process. There may be other race conditions meaning we could get duplicates of these (each of which may have a `File` object). This is harder to fix, at least without adding locks, but hopefully less of a problem. :thinking: 

(And yes, using properties this way is pretty common - though in a lot of cases once you set up the attribute once, you never delete it, so the only race condition is that you might compute it twice. The stdlib has a `cached_property` decorator which wraps this pattern up with a lock, though as we've just found on h5py, that can cause other issues.)

philsmt:
Thanks for this discussion, LGTM!

PR #290, opened 2022-03-15
Fix single run status when opening virtual overview files

### LINKED ISSUE #284: Run opened from virtual overview file is not considered a single run
Something we noticed today while debugging an issue in offline calibration.

The fix is very easy, but ideally we can also add a test that would have caught this (e.g. load a virtual overview file and ensure we can use `.get_run_value()`.

takluyver:
Closes #284

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1419,7 +1419,7 @@ def RunDirectory(
     if _use_voview and (sel_files == files):
         voview_file_acc = voview.find_file_valid(path)
         if voview_file_acc is not None:
-            return DataCollection([voview_file_acc])
+            return DataCollection([voview_file_acc], is_single_run=True)
 
     files_map = RunFilesMap(path)
     t0 = time.monotonic()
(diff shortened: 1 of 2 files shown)


philsmt:
Thanks! Based on how the function works right now, LGTM!

I was thinking of a more general problem though: Given that EXtra-data is supposed to support any location of EuXFEL files, a couple of files in a single directory do not necessarily constitute the same run. I could copy together files from several runs and logically use `RunDirectory` to obtain the `DataCollection` with `METADATA` then spanning multiple runs.

takluyver:
I think the naming of 'RunDirectory' implies that it represents a single run, so it's reasonable for EXtra-data to assume that's what it is. If you copied files from multiple runs into one directory, it's effectively a very small lie to open them as a 'RunDirectory'. You could instead list the filenames and use `DataCollection.from_paths()` (which is admittedly not documented).

What else could we do? We could check the filenames for the `-R0097-` bit to decide if they're a single run, but so far we've largely avoided relying on filenames. If you don't find a run number, do you assume they're the same or not? What about the same run number from different proposals?

We could look inside every file to see if their metadata matches. But we try quite hard to avoid opening all the files, because it gets slow for large runs.

philsmt:
I'm sorry, I realized after posting it may be misleading. In any case though, I would consider this a separate issue outside of this PR which fixes a specific issue within the semantics the function had all along.

takluyver:
OK, thanks. Feel free to open an issue to discuss that.

PR #293, opened 2022-03-20
Get run metadata and run values from a SourceData object

### LINKED ISSUE #263: Get RUN data from SourceData interface
This came up in a conversation with @philsmt : if you have a `SourceData` object, it should be possible to get RUN data associated with that source.

tmichela:
closes #263 

Slight difference with past behavior: now if you union multiple runs and a source is only present in one of the runs, this source can still get run metadata and run values.

### DIFF
--- a/extra_data/read_machinery.py
+++ b/extra_data/read_machinery.py
@@ -2,15 +2,17 @@
 
 The public API is in extra_data.reader; this is internal code.
 """
-from glob import iglob
 import logging
 import math
-import numpy as np
 import os.path as osp
 import re
 import time
+from collections import defaultdict
+from glob import iglob
+from typing import Union
 from warnings import warn
 
+import numpy as np
 
 log = logging.getLogger(__name__)
 
@@ -250,6 +252,24 @@ def find_proposal(propno):
         return d
 
     raise Exception("Couldn't find proposal dir for {!r}".format(propno))
+
+
+def same_run(*args) -> bool:
+    """return True if arguments objects contain data from the same RUN
+
+    arguments can be of type *DataCollection* or *SourceData*
+    """
+    # DataCollection union of format version = 0.5 (no run/proposal # in
+    # files) is not considered a single run.
+    proposal_nos = set()
+    run_nos = set()
+    for dc in args:
+        md = dc.run_metadata() if dc.is_single_run else {}
+        proposal_nos.add(md.get("proposalNumber", -1))
+        run_nos.add(md.get("runNumber", -1))
+
+    return (len(proposal_nos) == 1 and (-1 not in proposal_nos)
+            and len(run_nos) == 1 and (-1 not in run_nos))
 
 
 glob_wildcards_re = re.compile(r'([*?[])')
--- a/extra_data/read_machinery.py
+++ b/extra_data/read_machinery.py
@@ -7,9 +7,7 @@ import math
 import os.path as osp
 import re
 import time
-from collections import defaultdict
 from glob import iglob
-from typing import Union
 from warnings import warn
 
 import numpy as np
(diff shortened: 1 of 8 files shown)


takluyver:
Thanks! Is there a particular reason to expose `run_metadata()` this way? I'm a bit surprised because to my mind, that's outside any source, so it's odd to have it attached to a SourceData object.

The change in 'single run' behaviour also seems strange for `.run_metadata()` - you could union two runs together, and then `dc[src1].run_metadata()` might work where `dc[src2].run_metadata()` might fail. It makes more sense for run values - though even there I have some concern that it will be confusing for users if things sometimes work and sometimes don't.

extra_data/reader.py
```diff
@@ -550,25 +544,11 @@ def get_run_value(self, source, key):
...
-            # Allow for the .value suffix being omitted
-            ds = ds.get('value')
-        if not isinstance(ds, h5py.Dataset):
-            raise PropertyNameError(key, source)
-
-        val = ds[0]
-        if isinstance(val, bytes):  # bytes -> str
-            return val.decode('utf-8', 'surrogateescape')
-        return val
+        return source_data.run_value(key)
```
takluyver, line 551:
```suggestion
        return self._get_source_data(source).run_value(key)
```

This can save the try/except.

takluyver:
I might say that if you want information for a given source *and* some extra information from 'outside' that source, it's clearer to push people to pass two things - either the run object (DataCollection) and a source name, or SourceData plus the run metadata dict (or the single value you need from it). Maybe the convenience outweighs this, though. :shrug:

extra_data/read_machinery.py
```diff
@@ -2,15 +2,17 @@
...
-from glob import iglob
 import logging
 import math
-import numpy as np
 import os.path as osp
 import re
 import time
+from collections import defaultdict
+from glob import iglob
+from typing import Union
```
takluyver, line 12:
```suggestion
from glob import iglob
```

LGTM pointed out that there were a couple of unused imports.

tmichela:
> Thanks! Is there a particular reason to expose `run_metadata()` this way? I'm a bit surprised because to my mind, that's outside any source, so it's odd to have it attached to a SourceData object.

SourceData needs run metadata now for union-ing since this PR makes run values accessible. I just copied the method name to match the one from DataCollection (that made it convenient to use the same [function](https://github.com/European-XFEL/EXtra-data/pull/293/files#diff-3f63cd74e7a10ae666902ffca67496e10fc91974aa13a32364119e431569eb26R255) for that). I could make it private, but also thought that does not hurt 🙃 

> The change in 'single run' behaviour also seems strange for `.run_metadata()` - you could union two runs together, and then `dc[src1].run_metadata()` might work where `dc[src2].run_metadata()` might fail. It makes more sense for run values - though even there I have some concern that it will be confusing for users if things sometimes work and sometimes don't.

I believe in practical use cases, this would *never* be an issue (sources are present in all runs you'd usually want to merge) but just an artifact of the implementation. This happened in the existing tests, because entirely unrelated runs were merged.

extra_data/sourcedata.py
```diff
@@ -203,9 +206,73 @@ def _only_tids(self, tids) -> 'SourceData':
...
+        combine data from multiple runs, it will raise MultiRunError.
+
+        Returns the RUN parameter value corresponding to the *key* argument.
+        """
+        print('sourcedtata', self.is_single_run)
+        if not self.is_single_run:
+            raise MultiRunError()
+
+        if not self._is_control:
+            raise SourceNameError(self.source)
```
takluyver, line 238: This roughly made sense for `run.get_run_value(src, key)`, but I think it would be surprising for `src.run_value(key)` to raise SourceNameError. Maybe we should have a generic exception like ValueError with an explanation that instrument sources don't have RUN data.

tmichela:
Thanks for the review !

PR #296, opened 2022-03-22
Add mockdata for CONTROL sources for AGIPD1M and AGIPD500

kakhahmed:
While working on adding some tests for methods in pyCalibration to read parameter values from INSTRUMENT and RUN sources from AGIPD runs. I was wondering if I should use actual files or mock the data needed for tests.

After realizing that pyCalibration already uses the mocked examples in EXtra-data, I though that I might add to it in the purpose for using it for more pyCalibration unit tests.

In this MR, there is the new `mockdata/agipd.py` file with classes for writing CONTROL and RUN parameters needed for  the tests on https://git.xfel.eu/detectors/pycalibration/-/merge_requests/646

And three functions to make different agipd runs

1. `make_agipd1m_run`: AGIPD1M run with INSTRUMENT and subset of needed parameters at CONTROL and RUN sources.
2. `make_agipd1m_run_old`: AGIPD1M run with INSTRUMENT and subset of needed parameters at CONTROL and RUN sources. And with missing parameters compared to `make_agipd1m_run`
3. `make_agipd500k_run`: AGIPD500K run which differ in some of the CONTROL and RUN sources and keys.

### DIFF
--- a/extra_data/tests/mockdata/agipd.py
+++ b/extra_data/tests/mockdata/agipd.py
@@ -1,5 +1,3 @@
-import numpy as np
-
 from .base import DeviceBase
 
 
@@ -18,9 +16,12 @@ class AGIPDMDL(DeviceBase):
         # A sample of some of the available keys.
         self.control_keys = [
             ('acquisitionTime', 'u2', ()),
-            ('bunchStructure.nPulses', 'u2', ()),
-            ('bunchStructure.firstPulse', 'u2', ()),
-            ('bunchStructure.methodIndex', 'u2', ()),
+            ('bunchStructure/nPulses', 'u2', ()),
+            ('bunchStructure/firstPulse', 'u2', ()),
+            ('bunchStructure/methodIndex', 'u2', ()),
+            ('setupr', 'u2', ()),
+            ('patternTypeIndex', 'u2', ()),
+            ('gainModeIndex', 'u2', ()),
             ('egsgat', 'u2', ()),
             ('g1sgat', 'u2', ()),
             ('g2sgat', 'u2', ()),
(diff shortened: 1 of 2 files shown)


extra_data/tests/mockdata/agipd.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def write_control(self, f):
+        super().write_control(f)
+
+        ctrl_grp = f'CONTROL/{self.device_id}'
+        run_grp = f'RUN/{self.device_id}'
+        for grp in [ctrl_grp, run_grp]:
+            length = self.ntrains if grp == ctrl_grp else 1
+
+            f[f"{grp}/setupr/value"] = np.array([32]*length)
```
takluyver, line 41: For the ones that are always set, I'd recommend putting them in the `control_keys` list above, so that's as complete as possible, and then filling out the values here, something like this:

```suggestion
            g = f[grp]
            g["setupr/value"][()] = 32
```

extra_data/tests/mockdata/agipd.py
```diff
@@ -0,0 +1,130 @@
...
+            f[f"{grp}/patternTypeIndex/value"] = np.array([4]*length)
+            f[f"{grp}/gainModeIndex/value"] = np.array([0]*length)
+
+            if self.integration_time:
+                f[f"{grp}/integrationTime/value"] = np.array([15]*length)
```
takluyver, line 46: For the ones that are optional, don't forget to check what the dtype is in real files and set it appropriately here, e.g. `np.full(length, 15, np.uint64)`  (I haven't checked, so uint64 is probably wrong).

Also, don't forget to create the `/timestamp` field manually for any keys that aren't in `control_keys`.

extra_data/tests/mockdata/agipd.py
takluyver: (There are some similar examples for this in the modules `sidemic_camera` and `gec_camera`, although they're both for instrument data rather than control)

extra_data/tests/mockdata/agipd.py
takluyver: Thinking about it, it's probably easier to conditionally add them to `control_keys` inside `__init__`, then they'll get timestamps created automatically, and you can just fill them in here.

extra_data/tests/mockdata/agipd.py
kakhahmed: yes, I kept getting errors with what I wrote if the properties are already in control_keys. This is why I removed it.

Thanks I will try the suggested code.

takluyver:
Thanks @kakhahmed !

extra_data/tests/mockdata/agipd.py
kakhahmed: Suggested code is in place. Thank you Thomas!

kakhahmed:
😅 , it is already merged. I really need to keep better track for the github updates.

Thanks a lot @takluyver !

PR #298, opened 2022-03-22
Add support to open run data from any number of data locations

philsmt:
Somewhat coincidental to our internal discussion on differentiating data from `raw` and `proc`, I already planned to extend the `data` argument of `open_run` to allow specifying multiple and arbitrary locations relative to the proposal root. The main use case is the ONC offline corrections, which by default output to `<proposal>/scratch/cal`. With this PR, you could open this like:

```python
run = open_run(proposal=123, run=456, data=['raw', 'scratch/cal'])
```

The previous case `data='all'` is implemented as an alias for `['raw', 'proc']`, and it preserves the overwriting rules we established for it.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -21,7 +21,7 @@ import sys
 import tempfile
 import time
 from collections import defaultdict
-from collections.abc import Iterable, Mapping
+from collections.abc import Iterable, Mapping, Sequence
 from itertools import groupby
 from multiprocessing import Pool
 from operator import index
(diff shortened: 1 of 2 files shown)


philsmt:
Do you think we're still going for some layer implementation or we can continue here?

takluyver:
I've rebased this onto current master. Now that loading aliases no longer prints a message (#478), I've also simplified the logic a little bit to just load the aliases again for each origin.

@philsmt would you have time to write a test for this?

philsmt:
Thanks, I did! The diff looks weird, but essentially I moved the bits testing `data='all'` from `test_open_run` into `test_open_run_multiple` parametrizing `all` and `['raw', 'proc']`.

takluyver:
Thanks,  LGTM

philsmt:
Thanks for review, probably my MR with the longest review time. Could've waited the two weeks to make it 2 years 😁

PR #307, opened 2022-04-21
Allow KeyData.ndarray to read into a pre-allocated array

philsmt:
Noticed on the side while discussing AGIPD's memory problems during corrections, it' fairly easy to allow `KeyData.ndarray()` to accept a pre-allocated array.

I decided to only check for the correct shape rather than dtype, as `Dataset.read_direct`'s semantics allow for type coercion during reading.

### DIFF
--- a/extra_data/tests/test_keydata.py
+++ b/extra_data/tests/test_keydata.py
@@ -236,6 +236,7 @@ def test_drop_empty_trains(mock_sa3_control_data):
     assert frame_counts.shape == (250,)
     assert frame_counts.min() == 1
 
+
 def test_single_value(mock_sa3_control_data, monkeypatch):
     f = H5File(mock_sa3_control_data)
 
@@ -282,6 +283,18 @@ def test_single_value(mock_sa3_control_data, monkeypatch):
         intensity.as_single_value()
 
     np.testing.assert_equal(intensity.as_single_value(rtol=1), np.median(data))
+
+
+def test_ndarray_out(mock_spb_raw_run):
+    f = RunDirectory(mock_spb_raw_run)
+    cam = f['SPB_IRU_CAM/CAM/SIDEMIC:daqOutput', 'data.image.dims']
+
+    buf_new = cam.ndarray()  # New copy of data.
+    buf_in = np.zeros(cam.shape, dtype=cam.dtype)
+    buf_out = cam.ndarray(out=buf_in)  # In-place copy of data.
+
+    np.testing.assert_allclose(buf_new, buf_out)
+    assert buf_in is buf_out
 
 
 @pytest.fixture()
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -216,18 +216,22 @@ class KeyData:
 
     # Getting data as different kinds of array: -------------------------------
 
-    def ndarray(self, roi=()):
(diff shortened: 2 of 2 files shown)


extra_data/tests/test_keydata.py
```diff
@@ -284,6 +285,22 @@ def test_single_value(mock_sa3_control_data, monkeypatch):
...
 
+def test_ndarray_out(mock_sa3_control_data):
+    f = H5File(mock_sa3_control_data)
+    beamview = f['SA3_XTD10_IMGFEL/CAM/BEAMVIEW2:daqOutput', 'data.image.dims']
+
+    # Get data and get a new array.
+    buf_new = beamview.ndarray()
+
+    # Get data into a pre-allocated array.
+    buf_in = np.empty(  # Do not use .zeros here to get random data!
```
takluyver, line 296: If you want a test that doesn't need this caveat, the Basler cameras in the mock data actually have their `dims` array filled in.

extra_data/tests/test_keydata.py
```diff
@@ -284,6 +285,22 @@ def test_single_value(mock_sa3_control_data, monkeypatch):
...
+        beamview.shape[:1] + beamview.entry_shape, dtype=beamview.dtype)
+    buf_out = beamview.ndarray(out=buf_in)
+
+    np.testing.assert_allclose(buf_new, buf_out)
+    assert np.may_share_memory(buf_in, buf_out)
```
takluyver, line 301: TIL about this function. :slightly_smiling_face: 

I think we could just `assert buf_out is buf_in`, though?

philsmt: That is an interesting question, is that part of its API? I've always considered the `out=` argument to indicate memory to use, not necessarily exactly that object. But you're probably right.

philsmt: Interestingly the `numpy` documentation is not 100% explicit here. Taking [`numpy.add`](https://numpy.org/doc/stable/reference/generated/numpy.add.html) as an example:

> out: ndarray, None, or tuple of ndarray and None, optional
> A location into which the result is stored. [...] If not provided or None, a freshly-allocated array is returned.

> Returns: add: ndarray or scalar
> The sum of x1 and x2, element-wise.

The choice of words does not hint that the returned array at any point is the same one as passed to `out=`, only that there is no additional array allocation (is a new view an array allocation?). It leaves room for the returned value to possibly be of different type, or have additional annotations without breaking the API.

takluyver: Interesting - I've always taken it to mean that the same array object is returned. There's no obvious need to create a new view if you're not changing the shape or the dtype. :shrug:

philsmt: So would you like to make that part of the interface? I am fine either way, so far I was just under the assumption this is not guaranteed.

takluyver: I think I'm happy enough with that to make the test more specific - I can't imagine any reason we'd need to return a different view of the array that's passed in.

extra_data/tests/test_keydata.py
philsmt: Thanks, switched the source.

Issue #292, opened 2022-03-16
Combining raw & proc: specify which sources should be proc?


takluyver:
I got a request from FXE: the JUNGFRAU calibration pipeline had failed (because of issues with GPFS, outside our control), and so when they went to access data with `open_run(..., data='all')`, the JUNGFRAU data it offered them was from raw instead of from proc. They specifically wanted the corrected data, so it would have been clearer to say that no JF data was found.

The immediate issue is that EXtra-data doesn't know what sources are *meant* to be in proc, so anything that isn't in proc will be exposed from raw. One obvious way round this is to let users specify that sources are expected to be in proc - e.g. `proc_only='*/DET/JNGFR*'` . But this is a clumsy workaround - it mostly works without this, so people won't bother setting it before they hit the problem, and it's easy to exclude stuff you might want (e.g. a `.../DET/JNGFRCTRL` source which is not meant to go in proc).

In the longer term, I want correction to write data with a new source name (e.g. `.../CORR/JNGFR01`), so you can clearly refer to raw/corrected data as separate things. But this is going to be a big change in offline correction. We might want to offer something in EXtra-data before that.

- If we decide what the source names for corrected data will look like, we might try to 'rename' them in EXtra-data before the change in the files. This may get fiddly and confusing, though - so far, EXtra-data has always reflected what's in the files.
- We could add `run.raw` and `run.proc` (or `.corr`?) attributes, which would point to separate `DataCollection` objects for the raw and proc data, so you could use `run.proc['.../DET/JNGFR01']` to ensure you were getting the proc data. It would be pretty simple to do this in a crude way, but if you wanted e.g. `run.select()` to affect the separate raw & proc data collections, it would get much more involved.

We might also add a new high-level function like `open_run` that defaults to opening raw & proc together - it's hard to change `open_run` without breaking existing code.

PR #310, opened 2022-05-10
WIP: Layers

takluyver:
The idea of this is that you can 'overlay' raw & proc data and still be able to access either:

```
In [2]: run = open_run(2734, 182, data=('raw', 'proc'))

# Find the source in the first data section where it exists (here raw):
In [5]: run['SPB_DET_AGIPD1M-1/DET/10CH0:xtdf', 'image.data'].dtype
Out[5]: dtype('uint16')

# Get the source from a specific layer (syntax up for discussion):
In [7]: run.layer('raw')['SPB_DET_AGIPD1M-1/DET/10CH0:xtdf', 'image.data'].dtype
Out[7]: dtype('uint16')

In [8]: run.layer('proc')['SPB_DET_AGIPD1M-1/DET/10CH0:xtdf', 'image.data'].dtype
Out[8]: dtype('<f4')
```

I'm not sure about this yet - it makes bits of the code rather more complex, and it might not be used that often.

The aim is that you can have a combined view of the data - e.g. to do `.select(..., require_all=True)` - and specifically access e.g. AGIPD proc data, telling it that you want proc data so it fails if that's missing rather than quietly switching to raw (#292). We could even make `data=('raw', 'proc')` the default without changing things that already work (mostly).

@philsmt this would be an alternative to your #298. In this PR, the first named origin containing a source 'wins', as opposed to the last in your PR.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -571,6 +571,16 @@ class DataCollection:
             return self._sources_data.maps
         return [self._sources_data]
 
+    def layer(self, name):
+        ix = self._layer_names.index(name)
+        sources_dict = self._layers[ix]
+        files = set().union(*[sd.files for sd in sources_dict.values()])
+        return DataCollection(
+            files, sources_data=sources_dict, train_ids=self.train_ids,
+            inc_suspect_trains=self.inc_suspect_trains,
+            is_single_run=self.is_single_run, layer_names=(name,),
+        )
+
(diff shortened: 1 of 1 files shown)


takluyver:
It looks like we're going to move forwards with making new source names for corrected data, in which case this approach is unnecessary. We'll press ahead with #298 & #468 to handle that.

PR #317, opened 2022-05-31
FIX: open run with data='all' when proc folder exists but is empty

tmichela:
[open_run](https://github.com/European-XFEL/EXtra-data/blob/master/extra_data/reader.py#L1459) expects a `FileNotFoundError` when opening `data='all'` if the proc folder does not exist. Sometimes, a proc folder is created, but does not contain any data. In this case, `RunDirectory` throws a generic `Exception` when a `FileNotFoundException` is expected.


JamesWrigley (approved):
I just ran into this, so I'll give it a sneaky LGTM.

PR #324, opened 2022-06-14
Add a float value for measurementSenseVoltage only if bias_voltage fl…

kakhahmed:
This is a small fix for the agipd mock test. I forgot to remove the `measurementSenseVoltage` from `control_keys` when `bias_voltage` is False.

This is related to https://git.xfel.eu/detectors/pycalibration/-/merge_requests/646/diffs


takluyver:
Thanks!

PR #325, opened 2022-06-14
Add KeyData.source_file_paths

takluyver:
This is a property which finds the underlying source files even if we're using a virtual overview file for the run. cc @philsmt @kakhahmed

Open questions:

- What should it return? String paths (currently implemented), pathlib Paths, or extra_data FileAccess objects (like `.files`)?
- If no trains are selected, are we happy with it returning an empty set of source files?
- If a file is within the selection but the relevant data has length 0, should it still be included?
  - Probably not, because this would be hard to ensure when using a virtual overview file
  - If not, a key which was present but not recorded will have 0 source files


takluyver:
>> If no trains are selected, are we happy with it returning an empty set of source files?
>
> I am not sure if I understood this correctly. 

E.g. if you do:

```python
xgm_beam_x = run['SPB_XTD9_XGM/DOOCS/MAIN', 'beamPosition.ixPos.value']
xgm_beam_x[:0].source_files
```

Is `.source_files` empty? Or do we try to ensure that it includes at least one file, so you can always use `.source_files[0]`?

The `[:0]` is kind of silly, but you could get the same if you ask for only trains where there's data for X, Y and Z, and that leaves you with no trains, e.g. because one of them has no data at all.

kakhahmed:
Thank you. This LGTM. I can test it with one of the notebooks if you like

takluyver:
Thanks - if you've got time to check that it works in a real use case, that would be great, but it's not essential if you're busy.

philsmt:
I'm sorry for coming late to this review.

While I personally find `Path` objects very useful, I wonder if it makes sense to keep this property as close to the original `.files` property for backwards compatibility and return `FileAccess` objects. As we learned there is plenty of user code making use of the original `.files` semantics in some way, and kind-of-reintroducing it with a different type would require more changes to that code than just a simple rename.

We could move the current implementation to something like `_source_paths` and create `FileAccess` objects from them for `_source_files`.

takluyver:
I can definitely see the argument for keeping `.source_files` closer to `.files`. But if you've started from a virtual overview file, opening the individual source files and reading their trains & sources is an extra cost, which you might not need if you're handing out the files to worker processes or Slurm jobs straight away.

So maybe we should expose both? `.source_file_paths` for the paths (either as Path objects or str) and `.source_files` for FileAccess objects?

philsmt:
Yes, that was what I tried to suggest, I'm sorry if it wasn't clear 🙈 

Thanks, LGTM!

takluyver:
I was just about to hit merge, but one more question occurred to me: `source_files` - the shorter name - now returns a list of FileAccess objects, even though `FileAccess` is undocumented and meant primarily for internal use. On the other hand, it's already exposed via attributes like `.files`, so it's not exactly *private*.

We could:

1. Expose smaller DataCollections with `[H5File(p) for p in self.source_files_paths]` instead, pushing people towards documented APIs.
2. Tweak the names so that the property giving Paths is more convenient than the one giving FileAccesses
3. Leave it as is :shrug: 
4. Document `FileAccess` to make it an explicitly supported API. My hesitation is that we already have an [offputtingly large amount of deprecated API](https://extra-data.readthedocs.io/en/latest/reading_files.html), and I don't think there's much need for FileAccesss.

My preference at the moment is for option 2.

philsmt:
I see your point, and agree in principle. We set out to offer compatibility with an established pattern that we broke, independent of the fact strictly speaking it was using private APIs. Hence as premise I would use what have people doing with `.files` that we need to support with `.source_files`? 

I like your suggestion 1) assuming that we also continue to offer the `Path` objects. Then someone interested in the latter can use that property, and people actually looking to access data on a file-level have proper `DataCollection`. `FileAccess` is not useful in such a case anyway, and from the top of my head I don't even see what meaningful functionality it offers to users beyond the path.

I realize now asking for `FileAccess` was looking too much through the lens of restoring compatibility.

takluyver:
I think this has been hanging around too long. So I propose that we go with `source_file_paths` (removing one `s`) returning pathlib paths for now, so we can get this in.

It's easy to wrap these in `H5File` or `FileAccess` objects with a list comprehension in calling code if needed. And we can always add one or both of these options back later, or move `source_file_paths` to a shorter name, if we want to. It's easier to add things than to take them away.

philsmt:
That sounds like a good compromise, LGTM.

I will update https://github.com/European-XFEL/EXtra-data/pull/399 to be based on this.

kakhahmed:
LGTM

PR #332, opened 2022-08-01
Save all available voview metadata

JamesWrigley:
Previously only metadata for version 1.0 was written.

### DIFF
--- a/extra_data/tests/conftest.py
+++ b/extra_data/tests/conftest.py
@@ -8,7 +8,7 @@ from tempfile import TemporaryDirectory
 from . import make_examples
 
 
-@pytest.fixture(scope='session', params=['0.5', '1.0'])
+@pytest.fixture(scope='session', params=['0.5', '1.0', '1.2'])
 def format_version(request):
     return request.param
 
--- a/extra_data/writer.py
+++ b/extra_data/writer.py
@@ -117,7 +117,8 @@ class FileWriter:
             metadata = {}
 
         metadata_grp = self.file.create_group('METADATA')
-        if metadata.get('dataFormatVersion') == '1.0':
+        format_version = float(metadata.get('dataFormatVersion'))
+        if format_version >= 1.0:
             self.write_sources(metadata_grp.create_group('dataSources'))
 
             # File format 1.0 should also have INDEX/flag
--- a/extra_data/tests/test_reader_mockdata.py
+++ b/extra_data/tests/test_reader_mockdata.py
@@ -850,7 +850,7 @@ def test_open_run(mock_spb_raw_run, mock_spb_proc_run, tmpdir):
 def test_open_file(mock_sa3_control_data):
     f = H5File(mock_sa3_control_data)
     file_access = f.files[0]
-    assert file_access.format_version in ('0.5', '1.0')
+    assert file_access.format_version in ('0.5', '1.0', '1.2')
     assert 'SA3_XTD10_VAC/TSENS/S30180K' in f.control_sources
     if file_access.format_version == '0.5':
         assert 'METADATA/dataSourceId' in file_access.file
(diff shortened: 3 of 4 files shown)


takluyver:
I think we should mark the output as dataFormatVersion 1.0 when the input files are 1.0, 1.1 or 1.2, because we don't want to deal with the differences between those. 1.1 changed the meaning of `INDEX/flag` (this was thankfully short lived), 1.2 changed it back and also added the `INDEX/origin` field, which EXtra-data currently doesn't use for anything.

extra_data/tests/conftest.py
```diff
@@ -8,7 +8,7 @@
 from . import make_examples
 
 
-@pytest.fixture(scope='session', params=['0.5', '1.0'])
+@pytest.fixture(scope='session', params=['0.5', '1.0', '1.2'])
```
takluyver, line 11: I'm not super keen on running a whole load of tests again with only a trivial difference, but we can live with that.

If we are going to mock format version 1.2, though, we should do it properly, which means adding the `INDEX/origin` dataset to the mocked data. I think it's fine to fill it all with `-1` for now.

https://github.com/European-XFEL/EXtra-data/blob/dc66ddab58d6d7d0a05a318c5fdb0c76a07e5114/extra_data/tests/mockdata/base.py#L189-L198

extra_data/writer.py
```diff
@@ -117,7 +117,8 @@ def write_metadata(self):
             metadata = {}
 
         metadata_grp = self.file.create_group('METADATA')
-        if metadata.get('dataFormatVersion') == '1.0':
+        format_version = float(metadata.get('dataFormatVersion'))
+        if format_version >= 1.0:
```
takluyver, line 121:
```suggestion
        if metadata.get('dataFormatVersion') != '0.5':
```

The float comparison would work here, but we should avoid converting version numbers to floats just as a general thing, because version 1.10 is different from 1.1.

philsmt: The objects returned by `pkg_resources.parse_version` can be compared using inequality operators as expected for version numbers. I think we can assume `pkg_resources` to be available anyway with `setuptools`/`distutils` and it should not add a dependency.

takluyver: `pkg_resources` can be very slow to import, though - it scans all the other installed packages when it's imported. So I strongly prefer to avoid using it if at all possible.

I always used to use a LooseVersion class from distutils for this, but now distutils is deprecated - though we could still use it for now - I think it will be gone in Python 3.12 or something. Or we could do a very simple bit of parsing ourselves, like `[int(m) for m in re.findall('\d+', version)] .

So I stick to `==` and `!=` whenever possible :wink:

philsmt: TIL! Quick googling lead me to this: https://stackoverflow.com/a/11887885
Could that be an option? It seems to be installed in my `extra_data` dev environment without having done so manually, and I didn't notice any perceivable slowness when importing (unlike `pkg_resources` tested directly afterwards).

takluyver: Yup, good point, `packaging` is an option too. Technically it's a new dependency, but it's one that's pretty widely available.

extra_data/tests/conftest.py
JamesWrigley: Alrighty, added in 3cb8ded.

extra_data/writer.py
JamesWrigley: Yeouch good catch, I did not realize `float()` had that behaviour :fearful: I decided to go with `packaging` because it's also useful in the tests for 1.2 (3cb8ded). Fixed in c4f1cd7.

JamesWrigley:
> I think we should mark the output as dataFormatVersion 1.0 when the input files are 1.0, 1.1 or 1.2, because we don't want to deal with the differences between those. 1.1 changed the meaning of `INDEX/flag` (this was thankfully short lived), 1.2 changed it back and also added the `INDEX/origin` field, which EXtra-data currently doesn't use for anything.

Roger that, added in c4f1cd7.

(and apologies for taking so long to get back to this, last couple of weeks have been hectic...)

tmichela:
LGTM

takluyver:
LGTM too, and sorry for taking way, way longer to get back to this. :speak_no_evil:

PR #337, opened 2022-08-18
Multi-module KeyData interface

takluyver:
This brings an API similar to KeyData to multi-module detector data, so instead of `agipd.get_array('image.data')` you can do something like `agipd['image.data'].xarray()`. I hope this will make using EXtra-data more consistent, at least once people are used to using these newer APIs.

This also adds the ability to get an unlabelled numpy array for multi-module detector data; previously you had to get an xarray, even if you didn't want the labels. I still think dimension labels are a good idea, but xarray is certainly an extra level of complexity, and many people are more familiar with numpy arrays.

Finally, this changes how Dask arrays are created for multi-module data - rather than making chunks directly based on the files, so each frame is split across multiple chunks, it now uses the `.split_trains()` method to break data into chunks to load, so each Dask chunk spans all modules. This simplifies the task graph for the common case where you want to process data frame-wise with all modules together. I had hoped that this would solve all our issues with Dask and make #333 obsolete, but unfortunately that doesn't seem to be the case - we might want some variant of that PR as well.

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -985,7 +985,6 @@ class XtdfImageMultimodKeyData(MultimodKeyData):
             # Ensure ROI applies to pixel dimensions, not the extra
             # dim in raw data (except AGIPD, where it is data/gain)
             roi = np.index_exp[:] + roi
-        print(f"{out_shape=}, {out.shape=}, {reading_view.shape=}, {self._extraneous_dim=}")
 
         for mod_ix, (modno, kd) in enumerate(sorted(self.modno_to_keydata.items())):
             for chunk in kd._data_chunks:
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -1017,7 +1017,7 @@ class XtdfImageMultimodKeyData(MultimodKeyData):
 
         return out
 
-    def dask_array(self, *, labelled=True, subtrain_index='pulseId',
+    def dask_array(self, *, labelled=False, subtrain_index='pulseId',
                    fill_value=None, astype=None, frames_per_chunk=None):
         from dask.delayed import delayed
         from dask.array import concatenate, from_delayed
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -656,9 +656,9 @@ class XtdfDetectorBase(MultimodDetectorBase):
             # Preserve the quirks of this method before refactoring
             if self[key]._extraneous_dim:
                 arr = arr.expand_dims('tmp_name', axis=2)
-            arr.coords['train_pulse'] = arr.indexes['train_pulse'].rename(
-                ['trainId', subtrain_index]
-            )
+            arr.coords['train_pulse'] = arr.indexes['train_pulse'].rename({
+                'train': 'trainId', subtrain_index[:-2]: subtrain_index
+            })
             return arr.rename({name: f'dim_{i}' for i, name in enumerate(arr.dims[2:])})
         else:
             return super().get_dask_array(key, fill_value=fill_value, astype=astype)
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -1018,15 +1018,17 @@ class XtdfImageMultimodKeyData(MultimodKeyData):
         return out
 
     def dask_array(self, *, labelled=True, subtrain_index='pulseId',
-                   fill_value=None, astype=None):
+                   fill_value=None, astype=None, frames_per_chunk=None):
         from dask.delayed import delayed
         from dask.array import concatenate, from_delayed
 
         entry_size = (self.dtype.itemsize *
             len(self.modno_to_keydata) * np.product(self._eg_keydata.entry_shape)
         )
-        # Aim for 1GB chunks, with an arbitrary maximum of 1024 frames
-        split = self.split_trains(frames_per_part=min(1024 ** 3 / entry_size, 1024))
+        if frames_per_chunk is None:
+            # Aim for 2GB chunks, with an arbitrary maximum of 1024 frames
+            frames_per_chunk = min(2 * 1024 ** 3 / entry_size, 1024)
+        split = self.split_trains(frames_per_part=frames_per_chunk)
 
         arr = concatenate([from_delayed(
             delayed(c.ndarray)(fill_value=fill_value, astype=astype),
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -1011,7 +1011,7 @@ class XtdfImageMultimodKeyData(MultimodKeyData):
         if unstack_pulses:
             # Separate train & pulse dimensions, and arrange dimensions
             # so that the data is contiguous in memory.
-            dim_order = ['module'] + out.coords['train_pulse'].names + self.dimensions[2:]
+            dim_order = ['module'] + out.indexes['train_pulse'].names + self.dimensions[2:]
             return out.unstack('train_pulse').transpose(*dim_order)
 
         return out
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -656,9 +656,10 @@ class XtdfDetectorBase(MultimodDetectorBase):
             # Preserve the quirks of this method before refactoring
             if self[key]._extraneous_dim:
                 arr = arr.expand_dims('tmp_name', axis=2)
-            renames = {'train': 'trainId', subtrain_index[:-2]: subtrain_index}
-            renames.update({name: f'dim_{i}' for i, name in enumerate(arr.dims[2:])})
-            return arr.rename(renames)
+            arr.coords['train_pulse'] = arr.indexes['train_pulse'].rename(
+                ['trainId', subtrain_index]
+            )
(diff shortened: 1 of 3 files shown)


extra_data/tests/test_components.py
```diff
@@ -249,6 +249,7 @@ def test_get_dask_array_lpd_parallelgain(mock_lpd_parallelgain_run):
     assert det.detector_name == 'FXE_DET_LPD1M-1'
 
     arr = det.get_dask_array('image.data')
+    print(arr)
```
philsmt, line 252: Leftover?

extra_data/components.py
```diff
@@ -926,6 +738,311 @@ def zip_trains_pulses(trains, pulses):
...
+        return MultimodKeyData(det_selected, self.key)
+
+    def select_trains(self, trains):
+        return self._with_selected_det(self.det.select_trains(trains))
+
+    def split_trains(self, parts=None, trains_per_part=None, frames_per_part=None):
+        for det_split in self.det.split_trains(parts, trains_per_part, frames_per_part):
+            yield self._with_selected_det(det_split)
+
+    def ndarray(self, *, fill_value=None, out=None, roi=(), astype=None, module_gaps=False):
```
philsmt, line 793: I remember I originally also wanted to add an `out=` parameter to `components` in https://github.com/European-XFEL/EXtra-data/pull/307, but exposing the expected shape turned out not as trivial.

Looking at this now, it seems fairly straightforward and should only require `module_gaps` and `roi`. This could it significantly easier to pre-allocate arrays.

EDIT: It turns out `XtdfImageMultimodKeyData` actually has that method `_shape()` albeit private.

extra_data/components.py
```diff
@@ -926,6 +738,311 @@ def zip_trains_pulses(trains, pulses):
...
+        return DataArray(arr, dims=self.dimensions, coords=coords)
+
+    def dask_array(self, *, labelled=False, fill_value=None, astype=None):
+        from dask.delayed import delayed
+        from dask.array import concatenate, from_delayed
+
+        entry_size = (self.dtype.itemsize *
+            len(self.modno_to_keydata) * np.product(self._eg_keydata.entry_shape)
+        )
+        # Aim for 1GB chunks, with an arbitrary maximum of 256 trains
```
philsmt, line 832: Why 1 GiB? Any sense in making it configurable?

extra_data/components.py
```diff
@@ -926,6 +738,311 @@ def zip_trains_pulses(trains, pulses):
...
+
+        if labelled:
+            from xarray import DataArray
+            coords = {'module': self.modules, 'trainId': self.train_id_coordinates()}
+            return DataArray(arr, dims=self.dimensions, coords=coords)
+
+        return arr
+
+
+class XtdfImageMultimodKeyData(MultimodKeyData):
```
philsmt, line 848: Not a comment but out of curiosity, is only `XtdfImageMultimodKeyData` fundamentally able to be pulse-resolved in the sense of having `INDEX/<source>/count > 1`?

extra_data/components.py
```diff
@@ -926,6 +738,311 @@ def zip_trains_pulses(trains, pulses):
...
+    _sel_frames_cached = None
+
+    def __init__(self, det: XtdfDetectorBase, key, pulse_sel=by_index[0:MAX_PULSES:1]):
+        super().__init__(det, key)
+        self.det = det  # Makes PyCharm happy that det is XtdfDetectorBase
+        self._pulse_sel = pulse_sel
+        entry_shape = self._eg_keydata.entry_shape
+        self._extraneous_dim = (len(entry_shape) >= 1) and (entry_shape[0] == 1)
```
philsmt, line 856: This is meant to catch this extra `,1,` axis injected by the DAQ?

extra_data/components.py
```diff
@@ -926,6 +738,311 @@ def zip_trains_pulses(trains, pulses):
...
+            # image.data, image.gain, image.mask in calibrated data
+            entry_dims = ['slow_scan', 'fast_scan']
+        else:
+            # Everything else seems to be 1D, but just in case
+            entry_dims = [f'dim_{i}' for i in range(ndim_inner)]
+        return ['module', 'train_pulse'] + entry_dims
+
+    # Used for .select_trains() and .split_trains()
+    def _with_selected_det(self, det_selected):
+        return XtdfImageMultimodKeyData(det_selected, self.key, self._pulse_sel)
```
philsmt, line 907: Do you think it might be beneficial to use `self.__class__` instead to facilitate further inheritance?
(same for method below)

extra_data/components.py
```diff
@@ -90,6 +89,32 @@ def _check_pulse_selection(pulses):
...
+def _select_pulse_ids(pulses, data_pulse_ids):
+    """Select pulses by ID across a chunk of trains
+
+    Returns a boolean array of which entries in data_pulse_ids match.
+    """
+    if isinstance(pulses.value, slice):
+        s = pulses.value
+        desired = np.arange(s.start, s.stop, step=s.step, dtype=np.uint64)
+    else:
+        desired = pulses.value
```
philsmt, line 101: I'm 99% sure the `else` branch can only be a `by_index`/`by_id` object, but I wonder if an additional check might prevent cryptic exceptions.

extra_data/tests/test_components.py
takluyver: Yup, thanks

extra_data/components.py
takluyver: They have `.shape` as a property, which is equivalent to `module_gaps=False` and no ROI. This matches KeyData objects, and various array-like things. `._shape()` is a private method just so I didn't have to come up with a good name for it - but if we agree on one, I'm happy to make it public.

extra_data/components.py
takluyver: It's entirely guesswork. I think Dask might prefer smaller chunks, but then you're also more likely to end up opening the same file on several nodes, because it's been split up into many tiny chunks.

I've got a separate PR #322 to let Dask pick the chunk size for `KeyData.dask_array()`, which would also make it configurable through a Dask config file. But its default is 128 MiB, and I guess that spreading that over 16 modules and loading 8 MiB at a time might be smaller than ideal.

extra_data/components.py
takluyver: Yup, it's only expected for the Xtdf classes. The base `MultimodKeyData` can still handle JUNGFRAU-style data, where pulses are stored along a second axis.

extra_data/components.py
takluyver: Theoretically, maybe, but I'd strongly discourage anyone subclassing this outside of EXtra-data - it would be very brittle - and if we subclass it internally, we can fix that when we come to it.

extra_data/components.py
takluyver: Yes, that makes sense.

takluyver: Hmmm, in fact the else branch is something like a list or an array. Presumably, anything which numpy can convert into an array. So it gets a bit more complicated, e.g. do we allow something that converts to an array of floats? :thinking:

philsmt: How so? I tried `isinstance(by_index, slice)` and it's not `True`, and if I traced the code paths correctly whatever is passed *should* be converted to `by_index`/`by_id` if it's not a `slice`.

extra_data/components.py
philsmt: I see, thanks. I asking because of `REMI` (to be renamed `DLD`), where I already pulse-resolved data as well. At the time I didn't fully realize that is indeed limited to XTDF currently, any non-XTDF source injects additional axes. On the other hand, it's especially useful here as the number of pulses is not guaranteed to stay constant.
In the field however, it does tend to require a few additional hoops to jump through for analysis and I'd like to simplify that with some `extra_data.components`.

extra_data/components.py
philsmt: Thanks.

It's not entirely scientific, but backed by anecdotal evidence and expert's suspicions: dCache does not like parallel access to the same file, but nicely scales with different nodes accessing different files. So it might be interesting to pick the default based on the storage location.

extra_data/components.py
philsmt: Oh, it's _only_ the naming bikeshed.

Maybe using the word buffer would make the intent clear?
`.buffer_shape(roi=(), module_gaps=False)`?
`.shape_for_buffer(...)?`

extra_data/components.py
takluyver: The isinstance check here is for `pulses.value`, i.e. the thing inside a `by_id` object. We could indeed do an extra check where this is called, though.

extra_data/components.py
takluyver: I've called this `buffer_shape` at your suggestion - I don't have a strong feeling about the name.

philsmt:
Thank you, LGTM!

takluyver:
Thanks for the review :+1:

PR #338, opened 2022-08-22
Propagate the parallelize option in open_run()

JamesWrigley:
This comes up when using `open_run(data="all")`, which will make another call
internally to load proc data with `open_run(data="proc")`. Since the
parallelize option wasn't passed, the call to open proc data would use the
default of `parallelize=True`.

Note: I refactored `test_open_run()` to re-use the code for creating a mock root directory with raw and proc directories etc, but I'm not sure I like how messy it is to return so many paths from a fixture :shrug: I could also copy and paste the setup lines into `test_open_run_daemonized()`.

### DIFF
--- a/extra_data/tests/conftest.py
+++ b/extra_data/tests/conftest.py
@@ -1,3 +1,4 @@
+import os
 import os.path as osp
 
 import h5py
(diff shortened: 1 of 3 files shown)


extra_data/tests/conftest.py
```diff
@@ -108,6 +109,24 @@ def mock_spb_raw_run(format_version):
...
+        raw_run_dir = osp.join(prop_dir, 'raw', 'r0238')
+        os.makedirs(raw_run_dir)
+        make_examples.make_spb_run(raw_run_dir)
+
+        # Set up proc
+        proc_run_dir = osp.join(prop_dir, 'proc', 'r0238')
+        os.makedirs(proc_run_dir)
+        make_examples.make_spb_run(proc_run_dir, raw=False)
+
+        yield td, prop_dir, raw_run_dir, proc_run_dir
```
takluyver, line 127: I agree that returning all these paths isn't the most elegant. I think you could skip returning `prop_dir` - the only place that it's used, it can easily be replaced with `proc_run_dir`. Then I think it's good enough.

takluyver:
Other than that, LGTM

extra_data/tests/conftest.py
JamesWrigley: Good catch, fixed in f9fef4e.

PR #340, opened 2022-09-08
Options to read & decompress data in parallel

### LINKED ISSUE #49: Load data from multiple files in parallel?
Since #30, EXtra-data can open multiple files in parallel when initially opening a run. But if we then need to pull data from many files - sequence files or detector modules - we still handle these one after the other. I'm noticing this in particular with detector data access - just constructing an object like `DSSC1M` needs to open all the files to get the number of entries recorded for each train, so it can identify gaps in the data. Maybe we could parallelise loading this.

Creating the 'virtual overview file' we discussed before would improve this case (because the indexes would be copied into the overview file). But it wouldn't help (much) with loading real data (e.g. pulse IDs): HDF5 would do more of the work, but it would still need to open & read from the separate files.

takluyver:
This adds `read_procs` and `decomp_threads` parameters to `KeyData.ndarray()` and `.xarray()`. They control the number of processes used to read data from HDF5 files, and threads used to decompress data in the specific pattern we use for gain/mask datasets in 2D data. They both default to 1, i.e. the status quo, and we avoid launching separate processes/threads when they're 1.

Testing with ~55 GB of JUNGFRAU data, I got a better-than-2x speedup reading uncompressed data with 10 processes (~1 minute -> ~24 seconds), and something like a 10x speedup reading compressed data with `decomp_threads=-1`, i.e. 1 thread per core, on a 72-core node (~1 min 40 s -> 10 s). The timings are pretty variable - AFAICT, filesystem access always is.

The `read_procs` option is kind of incompatible with passing in an `out` array, because the array needs to be in shared memory. I'm not sure how to deal with that in the API - we could reject using `out` and `read_procs` together, but you could also pass in an array in shared memory, and I don't know of any way to check for that.

Future work:

- Extend this to multi-module detector data (waiting on #337)
- More efficient decompression, with fewer temporary memory allocations (@JamesWrigley has been experimenting with this).

Closes #49.

### DIFF
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -365,7 +365,7 @@ class KeyData:
         ]
         return np.concatenate(chunks_trainids)
 
-    def xarray(self, extra_dims=None, roi=(), name=None):
+    def xarray(self, extra_dims=None, roi=(), name=None, read_procs=1, decomp_threads=1):
         """Load this data as a labelled xarray array or dataset.
 
         The first dimension is labelled with train IDs. Other dimensions
@@ -395,10 +395,20 @@ class KeyData:
         name: str
             Name the array itself. The default is the source and key joined
             by a dot. Ignored for structured data when a dataset is returned.
+        read_procs: int
+            Use this many processes to read data. Using up to 8-10 processes
+            seems to accelerate data access.
+        decomp_threads: int
+            Use this many threads to decompress certain data, or -1 for one
+            thread per CPU core. This only applies to compressed gain/mask
+            data from corrected 2D detectors, and will be ignored for other
+            keys.
         """
         import xarray
 
-        ndarr = self.ndarray(roi=roi)
+        ndarr = self.ndarray(
+            roi=roi, read_procs=read_procs, decomp_threads=decomp_threads
+        )
 
         # Dimension labels
         if extra_dims is None:
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -7,6 +7,7 @@ from .file_access import FileAccess
 from .read_machinery import (
     contiguous_regions, DataChunk, select_train_ids, split_trains, roi_shape
 )
+from .utils import available_cpu_cores
 
 class KeyData:
     """Data for one key in one source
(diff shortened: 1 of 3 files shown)


extra_data/read_machinery.py
```diff
@@ -271,3 +271,22 @@ def same_run(*args) -> bool:
 
 
 glob_wildcards_re = re.compile(r'([*?[])')
+
+
+# Based on _alloc function in pasha
+def zeros_shared(shape, dtype):
```
JamesWrigley, line 277: Could we expose this? Or add something like `KeyData.allocate_out(shared=True)`?

extra_data/keydata.py
```diff
@@ -216,33 +218,128 @@ def as_single_value(self, rtol=1e-5, atol=0.0, reduce_by='median'):
...
+                np.frombuffer(shuffled, dtype=np.uint8).reshape((out.itemsize, -1)).transpose()
+            ).view(out.dtype).reshape(1, 512, 1024)
+            out[target_ix] = data[roi]
+
+        with ThreadPool(decomp_threads) as tpool:
+            tpool.starmap(decompress_frame, enumerate(all_compressed_frames))
+
+        return out
+
+    def ndarray(self, roi=(), out=None, read_procs=1, decomp_threads=1):
```
JamesWrigley, line 288: `decomp_threads` failed for me when loading an AGIPD module with shape `(200000, 512, 128)`:
image

JamesWrigley (changes requested):
Some benchmarks :racehorse: I selected AGIPD runs with ~200 cells and loaded the first 1000 trains of a single module into memory.

This is loading a compressed `uint16` AGIPD module from p3025 with 200 cells:
image
(same but with a linear scale)
image


And loading an uncompressed `float32` AGIPD module from p3046 with 202 cells:
image
(same but with a linear scale)
image


Amusingly, it's faster to load the uncompressed data despite it being ~80.5GB on disk compared to ~2.34GB of compressed data :upside_down_face: (both `proc/` directories still being on GPFS).

But still a huge improvement :tada:

PR #348, opened 2022-10-13
Allow opening runs in parallel to work with spawned subprocesses

### LINKED ISSUE #347: Crash opening datasets with multiprocessing
In a somewhat unusual case, the library I use tries to parallelize the loading of the data. Due to the way the library I use works (ie: due to the usage of GPUs) and for technical reasons (ie: https://github.com/pytorch/pytorch/issues/40403), this requires that multiple processes be created (ie: not just threads). When this is attempted, the following error happens and the reason is that `multiprocessing` tries to `pickle` the entire `RunDirectory` object, but it fails to do so due to the existence of a local function, namely this one: https://github.com/European-XFEL/EXtra-data/blob/master/extra_data/reader.py#L156

Given that this function is only used to prevent `KeyboardInterrupt`s and it is quite general, it seems it would be easy to solve this problem by transforming it into a module-wide function with no harm to the functionality of the module.

```
    data = extra_data.RunDirectory(self.path)
  File "/gpfs/exfel/data/user/danilo/sherlock_env/lib/python3.7/site-packages/extra_data/reader.py", line 1450, in RunDirectory
    is_single_run=True, parallelize=parallelize
  File "/gpfs/exfel/data/user/danilo/sherlock_env/lib/python3.7/site-packages/extra_data/reader.py", line 169, in from_paths
    with Pool(processes=nproc, initializer=initializer) as pool:
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1.2/lib/python3.7/multiprocessing/context.py", line 119, in Pool
    context=self.get_context())
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1.2/lib/python3.7/multiprocessing/pool.py", line 176, in __init__
    self._repopulate_pool()
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1.2/lib/python3.7/multiprocessing/pool.py", line 241, in _repopulate_pool
    w.start()
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1.2/lib/python3.7/multiprocessing/process.py", line 112, in start
    self._popen = self._Popen(self)
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1.2/lib/python3.7/multiprocessing/context.py", line 284, in _Popen
    return Popen(process_obj)
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1.2/lib/python3.7/multiprocessing/popen_spawn_posix.py", line 32, in __init__
    super().__init__(process_obj)
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1.2/lib/python3.7/multiprocessing/popen_fork.py", line 20, in __init__
    self._launch(process_obj)
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1.2/lib/python3.7/multiprocessing/popen_spawn_posix.py", line 47, in _launch
    reduction.dump(process_obj, fp)
  File "/gpfs/exfel/sw/software/xfel_anaconda3/1.1.2/lib/python3.7/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
AttributeError: Can't pickle local object 'DataCollection.from_paths.<locals>.initializer'
```

takluyver:
Closes #347

This is @daniloefl's change

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -61,8 +61,8 @@ RUN_DATA = 'RUN'
 INDEX_DATA = 'INDEX'
 METADATA = 'METADATA'
 
-def initializer():
-    # prevent child processes from receiving KeyboardInterrupt
+def ignore_sigint():
+    # Used in child processes to prevent them from receiving KeyboardInterrupt
     signal.signal(signal.SIGINT, signal.SIG_IGN)
 
 
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -61,6 +61,10 @@ RUN_DATA = 'RUN'
 INDEX_DATA = 'INDEX'
 METADATA = 'METADATA'
 
+def initializer():
+    # prevent child processes from receiving KeyboardInterrupt
+    signal.signal(signal.SIGINT, signal.SIG_IGN)
+
 
 class DataCollection:
     """An assemblage of data generated at European XFEL
@@ -153,10 +157,6 @@ class DataCollection:
                 uncached.append(path)
 
         if uncached:
-            def initializer():
-                # prevent child processes from receiving KeyboardInterrupt
-                signal.signal(signal.SIGINT, signal.SIG_IGN)
-
             # Open the files either in parallel or serially
             if parallelize:
                 nproc = min(available_cpu_cores(), len(uncached))
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -160,7 +160,7 @@ class DataCollection:
             # Open the files either in parallel or serially
             if parallelize:
                 nproc = min(available_cpu_cores(), len(uncached))
-                with Pool(processes=nproc, initializer=initializer) as pool:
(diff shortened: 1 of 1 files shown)


extra_data/reader.py
```diff
@@ -61,6 +61,10 @@
 INDEX_DATA = 'INDEX'
 METADATA = 'METADATA'
 
+def initializer():
+    # prevent child processes from receiving KeyboardInterrupt
```
takluyver, line 65: I think a bit more clarity is useful now that this is separated from where it's used.

```suggestion
def ignore_sigint():
    # Used in child processes to prevent them from receiving KeyboardInterrupt
```

daniloefl:
Indeed, updating the name made sense, so I just agreed with your previous suggestion.

daniloefl:
Small fix now: the new name needs to be actually used in the code.

takluyver:
Thanks @daniloefl !

PR #359, opened 2022-12-01
FIX: handle empty CONTROL datasets

tmichela:
It happens to have empty CONTROL datasets (e.g. `/gpfs/exfel/exp/SPB/202202/p003111/raw/r0111/RAW-R0111-DA02-S00000.h5`)
The `INDEX/SOURCE/[first|count]` report correctly 0 on both datasets, but we don't check that case in `DataCollection.train_from_id` and `trains()`.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1298,11 +1298,12 @@ class TrainIterator:
                 if ds is None:
                     continue
 
-                count = file.get_index(source, '')[1][pos]
+                firsts, counts = file.get_index(source, group)
+                first, count = firsts[pos], counts[pos]
                 if not count:
                     continue
 
-                self._set_result(res, source, key, ds[pos])
+                self._set_result(res, source, key, ds[first])
 
         for source in self.data.instrument_sources:
             self._set_result(res, source, 'metadata',
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -388,7 +388,7 @@ class DataCollection:
             if file is None:
                 continue
 
-            firsts, counts = file.get_index(source, group)
+            firsts, counts = file.get_index(source, '')
             first, count = firsts[pos], counts[pos]
             if not count:
                 continue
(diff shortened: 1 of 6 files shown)


extra_data/reader.py
```diff
@@ -388,6 +388,10 @@ def train_from_id(
             if file is None:
                 continue
 
+            count = file.get_index(source, '')[1][pos]
+            if not count:
+                continue
+
             for key in self.keys_for_source(source):
                 path = '/CONTROL/{}/{}'.format(source, key.replace('.', '/'))
                 source_data[key] = file.file[path][pos]
```
takluyver, line 397: If CONTROL data isn't always there, let's also make sure we do the right thing if it's only partly missing. At present we assume that the position of the train in the index is also the position of the data.

```suggestion
            firsts, counts = file.get_index(source, group)
            first, count = firsts[pos], counts[pos]
            if not count:
                continue

            for key in self.keys_for_source(source):
                path = '/CONTROL/{}/{}'.format(source, key.replace('.', '/'))
                source_data[key] = file.file[path][first]
```

extra_data/reader.py
```diff
@@ -1288,11 +1292,15 @@ def _assemble_data(self, tid):
...
-                _, pos, ds = self._find_data(source, key, tid)
+                file, pos, ds = self._find_data(source, key, tid)
                 if ds is None:
                     continue
+
+                count = file.get_index(source, '')[1][pos]
+                if not count:
+                    continue
+
                 self._set_result(res, source, key, ds[pos])
```
takluyver, line 1304:
```suggestion
                firsts, counts = file.get_index(source, group)
                first, count = firsts[pos], counts[pos]
                if not count:
                    continue

                self._set_result(res, source, key, ds[first])
```

extra_data/reader.py
tmichela: My assumption was that the data is either present or entirely missing, but you're right, better be safe!

takluyver:
Thanks, LGTM

PR #361, opened 2022-12-05
Raise more explicit exceptions when HDF5 files are missing essential datasets

philsmt:
Scanning all proposals and their runs for statistics on the file format version also yielded data on the kind of _corrupt_ files out there in the wild.

For those files which are HDF but are missing enough critical structure to not be readable, I found the two possible options of missing the `INDEX` group (actually missing everything except `INSTRUMENT`) and missing the `METADATA` section. While EXtra-data technically handles that already, it fails rather cryptically with HDF's original exception of `Unable to open object (component not found)`. This PR adds checks at the first instances of accessing these datasets and throws a custom exception.

Really unsure about the exception text, any opinions? It's featured in particular when using `RunDirectory.from_paths` and `extra-data-validate`.

### DIFF
--- a/extra_data/exceptions.py
+++ b/extra_data/exceptions.py
@@ -1,6 +1,10 @@
 """Exception classes specific to extra_data."""
 
 
+class FileStructureError(Exception):
+    pass
+
+
 class SourceNameError(KeyError):
     def __init__(self, source):
         self.source = source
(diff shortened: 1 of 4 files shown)


takluyver:
Thanks, this makes sense. On the wording, I'd be more specific and say that a dataset is missing, something like:

> INDEX/trainId dataset not found (required in EuXFEL HDF5 files)

We can always generalise this or add other options later if we want to use the exception for other possible unexpected conditions.

takluyver:
Thanks, LGTM

PR #367, opened 2023-01-04
Add support for source and key aliases

philsmt:
It's been requested frequently to somehow allow easier access to sources and keys without the need to deal with Karabo device names and their properties. This PR adds support for this via an alias mechanism. As discussed towards the end of last year, it is mostly separate from the existing code using literal sources and keys via an index property `run.alias[...]` inspired by the likes of `pandas` and `xarray`.

Aliases may refer to a **source** or a **source-key** or equivalently to a  `SourceData` or a `KeyData` object. There are **no** pure key aliases that may be used with different sources. They can be attached to a `DataCollection` object via two methods:

* `DataCollection.with_aliases(*alias_defs)`
* `DataCollection.only_aliases(*alias_defs, strict=False, require_all=False)`

They may be passed any number of alias definitions, which may be

* A `dict` mapping the alias to a `str` (for sources) or 2-`tuple` of `str` (for source-keys).
* A `str` pointing to a JSON, YAML or TOML file with alias definitions. These should define a `dict` in the same way, with the additional option of using nested dictionaries for multiple keys per source. (See docstring)

With the returned `DataCollection`, aliases may **only** be used through the `DataCollection.alias` property.

```
run = open_run(proposal=1, run=2).with_aliases('aliases.yaml')
run.alias['my-source']  # Returns SourceData
run.alias['my-source', 'a-key']  # Returns KeyData
run.alias['my-source-key']  # Returns KeyData
run.alias['my-alias-to-a-missing-source']  # Raises SourceNameError
run.alias['my-bogus-alias']  # Raises AliasError
run.alias['my-source-key', 'another-key']  # Raises ValueError
run.alias.select('my-source')  # returns DataCollection
run.alias.deselect('my-source')  # returns DataCollection
```

Once attached, aliases are handed down to any `DataCollection` created from this, e.g. by selection. The aliases are merged as well when using `DataCollection.union()` as long as there is no conflict - in that case a `ValueError` is raised.

Aliases are shown in `DataCollection.info()` in tags `<` `>` for both sources and keys, even if a source is not shown in detail.

Documentation follows after the initial discussion 😃

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1547,29 +1547,113 @@ class AliasIndexer:
     def __init__(self, data):
         self.data = data
 
+    def _resolve_any_alias(self, alias):
+        try:
+            literal = self.data._aliases[alias]
+        except KeyError:
+            raise AliasError(alias) from None
+
+        return literal
+
+    def _resolve_source_alias(self, alias):
+        source = self._resolve_any_alias(alias)
+
+        if isinstance(source, tuple):
+            raise ValueError(f'{alias} not aliasing a source for this data')
+
+        return source
+
     def __getitem__(self, aliased_item):
         if isinstance(aliased_item, tuple) and len(aliased_item) == 2:
-            try:
-                source = self.data._aliases[aliased_item[0]]
-            except KeyError:
-                raise AliasError(aliased_item[0]) from None
-
-            if isinstance(source, tuple):
-                raise ValueError(f'{aliased_item[0]} not only aliasing a '
-                                    f'source for this data')
-
             # Source alias with key literal.
-            return self.data[source, aliased_item[1]]
-        else:
-            try:
-                literal_item = self.data._aliases[aliased_item]
-            except KeyError:
-                raise AliasError(aliased_item) from None
-
+            return self.data[self._resolve_source_alias(aliased_item[0]),
+                                aliased_item[1]]
+        elif isinstance(aliased_item, str):
             # Source or key alias.
-            return self.data[literal_item]
+            return self.data[self._resolve_any_alias(aliased_item)]
 
-        raise TypeError('expected source or sourcekey identifier')
+        raise TypeError('expected alias or (source alias, key) tuple')
+
+    def _resolve_aliased_selection(self, selection):
+        if isinstance(selection, dict):
+            res = {self._resolve_source_alias(alias): keys
+                    for alias, keys in selection.items()}
+
+        elif isinstance(selection, Iterable):
+            res = []
+
+            for item in selection:
+                if isinstance(item, tuple) and len(item) == 2:
+                    # Source alias and literal key.
+                    item = (self._resolve_source_alias(item[0]), item[1])
+                elif isinstance(item, str):
+                    item = self._resolve_any_alias(item)
+
+                    if isinstance(item, str):
+                        # Source alias.
+                        item = (item, '*')
+
+                res.append(item)
+
+        return res
+
+    def select(self, seln_or_alias, key_glob='*', require_all=False):
+        """Select a subset of sources and keys from this using aliases.
+
+        In contrast to :method:`DataCollection.select`, only a subset of
+        ways to select data via aliases is supported:
+
+        1. With a source alias and literal key glob pattern::
+
+            # Select all pulse energy keys for an aliased XGM fast data.
+            sel = run.alias.select('sa1-xgm', 'data.intensity*')
+
+        2. With an iterable of aliases and/or (source alias, key pattern) tuples::
+
+            # Select specific keys for an aliased XGM fast data.
+            sel = run.alias.select([('sa1-xgm', 'data.intensitySa1TD'),
+                                    ('sa1-xgm', 'data.intensitySa3TD')]
+
+            # Select several aliases, may be both source and key aliases.
+            sel = run.alias.select(['sa1-xgm', 'mono-hv'])
+
+            Data is included if it matches any of the aliases. Note that
+            this method does not support glob patterns for the source alias.
+
+        3. With a dict of source aliases mapped to sets of key names
+            (or empty sets to get all keys)::
+
+                # Select image.data from an aliased AGIPD and all data
+                # from an aliased XGM.
+                sel = run.select({'agipd': {'image.data'}, 'sa1-xgm': set()})
+
+        The optional `require_all` argument restricts the trains to those for
+        which all selected sources and keys have at least one data entry. By
+        default, all trains remain selected.
+
+        Returns a new :class:`DataCollection` object for the selected data.
+        """
+
+        if isinstance(seln_or_alias, str):
+            seln_or_alias = [(seln_or_alias, key_glob)]
+
+        return self.data.select(self._resolve_aliased_selection(
+            seln_or_alias))
+
+    def deselect(self, seln_or_alias, key_glob='*'):
+        """Select everything except the specified sources and keys using aliases.
+
+        This takes the same arguments as :meth:`select`, but the sources
+        and keys you specify are dropped from the selection.
+
+        Returns a new :class:`DataCollection` object for the remaining data.
+        """
+
+        if isinstance(seln_or_alias, str):
+            seln_or_alias = [(seln_or_alias, key_glob)]
+
+        return self.data.deselect(self._resolve_aliased_selection(
+            seln_or_alias))
 
 
 def H5File(path, *, inc_suspect_trains=True):
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -572,10 +572,12 @@ class DataCollection:
 
         This can be used to join multiple sources for the same trains,
         or to extend the same sources with data for further trains.
-        The order of the datasets doesn't matter.
+        The order of the datasets doesn't matter. Any aliases defined on
+        the collections are combined as well unless their values conflict.
 
         Returns a new :class:`DataCollection` object.
         """
+
         sources_data_multi = defaultdict(list)
         for dc in (self,) + others:
             for source, srcdata in dc._sources_data.items():
@@ -584,12 +586,29 @@ class DataCollection:
         sources_data = {src: src_datas[0].union(*src_datas[1:])
                         for src, src_datas in sources_data_multi.items()}
 
+        # Check for alias conflicts across all data collections.
+        duplicate_aliases = set(self._aliases.keys()).intersection(
+            *[dc._aliases for dc in others])
+
+        for alias in duplicate_aliases:
+            # Find all values occuring for duplicate aliases.
+            duplicate_alias_values = set([self._aliases[alias]]).union(
+                [dc._aliases[alias] for dc in others])
+
+            if len(duplicate_alias_values) > 1:
+                raise ValueError('conflicting aliases across data collections '
+                                 'to be merged')
+
+        aliases = self._aliases.copy()
+        for dc in others:
+            aliases.update(dc._aliases)
+
         train_ids = sorted(set().union(*[sd.train_ids for sd in sources_data.values()]))
         files = set().union(*[sd.files for sd in sources_data.values()])
 
         return DataCollection(
             files, sources_data=sources_data, train_ids=train_ids,
-            inc_suspect_trains=self.inc_suspect_trains,
+            aliases=aliases, inc_suspect_trains=self.inc_suspect_trains,
             is_single_run=same_run(self, *others),
         )
 
@@ -901,7 +920,7 @@ class DataCollection:
         files = set().union(*[sd.files for sd in sources_data.values()])
 
         return DataCollection(
-            files, sources_data, train_ids=train_ids,
+            files, sources_data, train_ids=train_ids, aliases=self._aliases,
             inc_suspect_trains=self.inc_suspect_trains,
             is_single_run=self.is_single_run
         )
@@ -939,7 +958,7 @@ class DataCollection:
 
         return DataCollection(
             files, sources_data=sources_data, train_ids=self.train_ids,
-            inc_suspect_trains=self.inc_suspect_trains,
+            aliases=self._aliases, inc_suspect_trains=self.inc_suspect_trains,
             is_single_run=self.is_single_run,
         )
 
@@ -974,7 +993,7 @@ class DataCollection:
 
         return DataCollection(
             files, sources_data=sources_data, train_ids=new_train_ids,
-            inc_suspect_trains=self.inc_suspect_trains,
+            aliases=self._aliases, inc_suspect_trains=self.inc_suspect_trains,
             is_single_run=self.is_single_run,
         )
 
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -25,6 +25,7 @@ from collections.abc import Iterable, Mapping
 from itertools import groupby
 from multiprocessing import Pool
 from operator import index
+from pathlib import Path
 from typing import Tuple
 from warnings import warn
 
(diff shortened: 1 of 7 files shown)


extra_data/reader.py
```diff
@@ -1004,52 +1355,84 @@ def keys_detail(s, keys, prefix=''):
...
-            if any(p.match(s) for p in details_sources_re):
-                # Detail for control sources: list keys
-                ctrl_keys = self[s].keys(inc_timestamps=False)
+            is_detailed = any(p.match(s) for p in details_sources_re)
+            has_aliases = s in srckey_aliases
+
+            print('  -', s, src_alias_list(s))
+
+            # Detail for control sources: list keys
+            ctrl_keys = self[s].keys(inc_timestamps=False)
```
takluyver, line 1393: Part of why we don't already show keys by default is that it can be significantly slower to get them - we cache source names but not key names. Virtual overview files help with that, but we're still some way from being able to assume voview files exist for all runs.

So I'd like to continue ensuring that `run.info()` and `lsxfel` don't load keys for all sources. I appreciate this code is getting pretty complex - maybe it should be split out in its own module.

extra_data/reader.py
```diff
@@ -581,17 +586,340 @@ def union(self, *others):
...
+        new_data._aliases.update(aliases)
 
+        return new_data
+
+    @property
+    def alias(self):
+        """Enables item access via source and key aliases."""
+        return DataCollection.AliasIndexer(self)
+
+    class AliasIndexer:
```
takluyver, line 807: I'm generally against putting classes inside other classes - I'd rather have it defined at the module level. I don't think it makes any difference to how it works, it's just the more 'normal' option IMO.

extra_data/reader.py
```diff
@@ -581,17 +586,340 @@ def union(self, *others):
...
+
+            with open(aliases_path, 'r') as f:
+                data = yaml.safe_load(f)
+
+        elif aliases_path.suffix == '.toml':
+            from sys import version_info as py_version
+
+            # Use the standard library's tomllib if possible and try
+            # tomli if necessary.
+            if py_version.major >= 3 and py_version.minor >= 11:
```
takluyver, line 657: I'd normally just try importing `tomllib` and fall back on ImportError, though I don't think it matters.

extra_data/tests/test_reader_mockdata.py
```diff
@@ -990,3 +991,273 @@ def test_run_metadata_no_trains(mock_scs_run):
...
+    with pytest.raises(ValueError):
+        run.alias['hv', 'pressure']
+
+    # Test using an existing source alias for a non-existing source.
+    with pytest.raises(SourceNameError):
+        run.alias['bogus-source']
+
+    # Test using an existing key alias for a non-existing key.
+    with pytest.raises(PropertyNameError):
+        run.alias['bogus-key']
```
takluyver, line 1034: I was initially somewhat surprised that you could create invalid aliases and get an error only when accessing them.

I think that probably makes sense - it allows someone to define one big collection of aliases where runs will only use a subset of the relevant sources. But it would be nice to know about typos in the alias definitions before someone tries to use them. Maybe we could include a method like `run.alias.check()` to tell you which aliases point to missing sources or keys?

This doesn't have to be part of this PR, though - maybe we want to think more about how best to expose that information.

extra_data/tests/test_reader_mockdata.py
```diff
@@ -990,3 +991,273 @@ def test_run_metadata_no_trains(mock_scs_run):
...
+
+
+def test_json_alias_file(mock_sa3_control_data, mock_sa3_control_aliases):
+    with TemporaryDirectory() as td:
+        aliases_path = Path(td) / 'aliases.json'
+        aliases_path.write_text('''
```
takluyver, line 1040: There's a fixture for that.

```suggestion
def test_json_alias_file(mock_sa3_control_data, mock_sa3_control_aliases, tmp_path):
    aliases_path = tmp_path / 'aliases.json'
    aliases_path.write_text('''
```

extra_data/reader.py
```diff
@@ -581,17 +586,340 @@ def union(self, *others):
...
+             SA3_XTD10_MONO/MDL/PHOTON_ENERGY:
+               mono-central-energy: actualEnergy
+
+        Returns a new :class:`DataCollection` object with the aliases
+        for sources and keys.
+        """
+
+        return DataCollection(
+            self.files, sources_data=self._sources_data,
+            train_ids=self.train_ids, aliases=self._parse_aliases(alias_defs),
```
takluyver, line 726: Should using `.with_aliases()` a second time discard your existing aliases or add to them? I was imagining it would add to them, but I'm not set on that.

E.g. I could imagine that you might have some shared aliases from the instrument or the proposal, plus some local ones specific to a particular analysis. You could apply both in one call to `.with_aliases()`, of course, but maybe it would feel more natural to do them in separate steps. :shrug:

takluyver:
I've commented on a few specific parts, but overall I think this is in good shape.

extra_data/reader.py
philsmt: Amusingly I didn't even intend either behaviour when writing this, even though it may seem like that with `.with_aliases()` accepting multiple definitions at the same time. The only reason for removing aliases I can think of is to avoid conflicts during a merge.

So the options are keeping `.with_aliases()` accepting multiple inputs and replacing, or having it extend aliases probably taking ony a single definition, and add some `.strip_aliases()` method to remove them.

I don't have a particular good argument for either, except maybe the other `.select()`/`.deselect()` methods by nature cannot be chained in a "merging" sense, so one may keep this idiom here.

extra_data/tests/test_reader_mockdata.py
philsmt: This would be possible right now by `run.only_aliases(strict=True)`, albeit you wouldn't get a summary, but the included `.select()` call would fail at the first missing source/key.

JamesWrigley:
I wonder if we should have support for loading aliases automatically? The reason is because device names and properties can change so people will need to load different aliases for different proposals; for example MID has gone through a couple iterations of Karabo devices that aggregate properties from multiple devices (usually motors) in a single device, and each time the naming has changed slightly.

The first way I thought of handling this was through instrument libraries (like SCS's toolbox), which could wrap `open_run()`:
```python
from extra_data import open_run as ed_open_run

def open_run(proposal, run_no, **kwargs):
    run = ed_open_run(proposal, run_no, **kwargs)
    
    # Check if an alias file exists at a certain location (usr/extra-data-aliases.yaml?),
    # or load a default list of aliases (e.g. for the XGM etc), or use an alias based on the date of the run.
    return run.with_aliases(foo)
```

So the user could run e.g.:
```python
import mid

run = mid.open_run(1234, 56)
```

And then they would have their instrument specific aliases without having to call `with_aliases()`. But what if we supported loading aliases from a specific file by default? I think standardizing it would be quite convenient since it means you wouldn't have to use instrument-specific libraries to load the right aliases for a proposal.

takluyver:
When I discussed this with Philipp, we agreed to get explicit alias support in and then discuss this question of how to make it more convenient.

I'm against automatically loading aliases based on something like the working directory of the notebook, or the current user running it, or anything like that. I don't want to make it easy for people to write notebooks which break (or worse, appear to work but do something different) when they're moved, or when someone else runs them, or whatever. But I could see using the proposal number we pass to `open_run` to get aliases defined for the proposal.

philsmt:
Ah, you beat me to replying to James just barely.

Indeed the file support was initially motivated by the idea of loading such files automatically, e.g. based on location of the running code. I ended up leaving it out of this MR for the sake of easier review, but also because I didn't end up liking the idea so much anymore after trying to implement it. There were some cheap ways in terms of runtime cost to do it, but you quickly run into corner cases where it's outright confusing. Next to the point Thomas raised, some people write modules rather than notebook, should it apply here as well?

I like the idea of attaching it to the proposal rather than the file as well, though it implies there's a canonical way to do aliases across a single user group. But one could define a fixed filename in some proposal location (say `<proposal>/usr/.extra-data-aliases.yaml`) that is searched for when `open_run` is used.

JamesWrigley:
Sorry I guess wasn't clear, I never had anything in mind other than alias files belonging to proposals :)

> But one could define a fixed filename in some proposal location (say `<proposal>/usr/.extra-data-aliases.yaml`) that is searched for when `open_run` is used.

Yes, that's exactly what I put in my code snippet (though I used the non-hidden `usr/extra-data-aliases.yaml`, I think that's better than a hidden file). Would that be too much for this PR? (I'm happy to make a follow-up PR if you like)

extra_data/reader.py
takluyver: If we do add automatic loading of aliases from a proposal number, I think that pushes me a bit towards saying that `.with_aliases()` could be additive, so that `open_run(3124, 11).with_aliases(...)` can combine the shared proposal aliases with your additional ones. I'm still not 100% set on that if there's a good argument to the contrary, but leaning that way.

extra_data/reader.py
philsmt: Pondering this, I don't see any other way than including the aliases only when a source is detailed anyway.

extra_data/reader.py
philsmt: I see the point, it'd be confusing to use the automatic aliases and suddenly lose them when adding more custom aliases.

The functionality of an additive `.with_aliases()` is present in this version though, albeit clunky:

```
dc = dc.merge(dc.with_aliases('additional-aliases.toml'))
```
Moving this functionality to `.with_aliases` begs the question how it should handle collisions? At the moment those in between the individual alias definitions are ignored, while `.merge()` checks for conflicts.

If we handle conflicts on some though, someone likely would like to remove aliases to add entirely different ones, so we need a give-me-this-DataCollection-without-aliases function as well.

takluyver: > Moving this functionality to `.with_aliases` begs the question how it should handle collisions?

Good question. My programmer instinct is to say 'last source wins', like `dict.update()`, but it would clearly get confusing if some code expects 'motor1' to be one source and you're redefining it to another source. So maybe checking for conflicts is the right way to go. Redefining the alias pointing to the same name should be OK, though (as I think it is in the current `.merge()` implementation).

 > If we handle conflicts on some though, someone likely would like to remove aliases to add entirely different ones, so we need a give-me-this-DataCollection-without-aliases function as well.

Makes sense. `.drop_aliases()`? `.without_aliases()`?

extra_data/reader.py
takluyver: That solution would be fine, IMO. If you think it's a substantially better UX to show aliased keys even without requesting the detailed info, we could build a dict of `key_aliases_by_source` and display info for those keys. Hopefully this would still be reasonably quick, but getting the full list of keys for every source definitely isn't always.

extra_data/reader.py
philsmt: > My programmer instinct is to say 'last source wins'

I wonder, how does this case differ from the discussion in https://github.com/European-XFEL/EXtra-data/pull/298 regarding order of overwriting? As a matter of fact, `.with_aliases()` worked like that up to now and simply overwrote whatever aliases are defined, in order.

I wonder, how about we make `.with_aliases()` simply run the line merge line above when it's called on a `DataCollection` that already has aliases defined?

takluyver: I suspect there are more efficient ways to do it than `.union()`, because that does a bunch of set operations on files and train IDs that are unnecessary for this. But it's probably not going to be a massive difference, so if that's the easiest way to get it working, it's probably fine.

> I wonder, how does this case differ from the discussion in https://github.com/European-XFEL/EXtra-data/pull/298 regarding order of overwriting?

I wondered if you'd think of that. :wink: I don't have a satisfying answer, but my mental model seemingly goes differently if I'm envisaging a single pool of data (last update wins) or some kind of layers (first=highest wins). In this case, I think that uncertainty is a good argument for error-ing on conflicts and letting the user clear it up.

extra_data/reader.py
philsmt: Okay, I reverted any key access if the source is not included in the details pattern.

extra_data/reader.py
philsmt: I gave this another thought and before implementation would like to re-check our plan:

* `DataCollection.with_aliases` merges any new alias definitions into the existing ones (erroring out on conflicts) on the `DataCollection`
* `DataCollection.only_aliases` merges any new alias definitions into the existing ones (erroring out on conflicts) and selects down to **all** aliases then defined on the `DataCollection`
* `DataCollection.merge` works as before and identical to `DataCollection.with_aliases` with respect to aliases
* Add a new call `DataCollection.drop_aliases`

philsmt: Yes, given the alias API is somewhat intended to "hide" the true source, I would avoid using any form of overwriting scheme and visibly error out if there's a conflict.

I changed the behaviour of all calls now to always add aliases instead of replacing them. When implementing the conflict detection, I noticed the previous implementation of checking for those was actually not complete - if you merged multiple `DataCollections` using `.merge()` and the conflict was only in between the added ones rather than the one you call it from, it would not have been detected. The new code now active for all alias operations properly checks for any combination.

extra_data/reader.py
```diff
@@ -564,15 +567,30 @@ def get_run_values(self, source) -> dict:
...
         """Join the data in this collection with one or more others.
 
         This can be used to join multiple sources for the same trains,
         or to extend the same sources with data for further trains.
-        The order of the datasets doesn't matter.
+        The order of the datasets doesn't matter. Any aliases defined on
+        the collections are combined as well unless their values conflict.
 
-        Returns a new :class:`DataCollection` object.
+        Returns a new :class:`DataCollection` object.s
```
takluyver, line 591:
```suggestion
        Returns a new :class:`DataCollection` object.
```

extra_data/reader.py
```diff
@@ -1331,6 +1584,126 @@ def __iter__(self):
...
+    def __getitem__(self, aliased_item):
+        if isinstance(aliased_item, tuple) and len(aliased_item) == 2:
+            # Source alias with key literal.
+            return self.data[self._resolve_source_alias(aliased_item[0]),
+                             aliased_item[1]]
+        elif isinstance(aliased_item, str):
+            # Source or key alias.
+            return self.data[self._resolve_any_alias(aliased_item)]
+
+        raise TypeError('expected source alias or (source alias, key) tuple')
```
takluyver, line 1623:
```suggestion
        raise TypeError('expected alias or (source alias, key) tuple')
```

extra_data/reader.py
```diff
@@ -1331,6 +1584,126 @@ def __iter__(self):
...
+
+                    if isinstance(item, str):
+                        # Source alias.
+                        item = (item, '*')
+
+                res.append(item)
+
+        return res
+
+    def select(self, seln_or_source_glob, key_glob='*', require_all=False):
```
takluyver, line 1648:
```suggestion
    def select(self, seln_or_alias, key_glob='*', require_all=False):
```

This parameter isn't a glob pattern, right? (Same for .deselect())

extra_data/tests/test_reader_mockdata.py
```diff
@@ -1,6 +1,7 @@
 from datetime import datetime, timedelta, timezone
 from itertools import islice
 from multiprocessing import Process
+from pathlib import Path
```
takluyver, line 4:
```suggestion
```

This didn't end up getting used.

setup.py
```diff
@@ -76,6 +76,8 @@ def find_version(*parts):
               'pytest',
               'pytest-cov',
               'testpath',
+              'pyyaml',
+              'tomli'
```
takluyver, line 80:
```suggestion
              'tomli; python_version < "3.11"',
```

takluyver (commented):
A few nitpicks, but otherwise this LGTM

extra_data/reader.py
philsmt: Thanks, good catch on all of these.

philsmt:
Thanks, I've cleaned up all the fixup commits then, ready to go from my side.

takluyver:
Thanks, LGTM

philsmt:
Would you like to have documentation as part of this MR or add it afterwards? I plan to get it done next week.

takluyver:
I'm happy enough for it to come a bit later - even after a release if needed - so long as it doesn't get forgotten.

philsmt:
I rebased on top of `master` to fix the merge conflict, feel free to merge.

extra_data/tests/test_aliases.py
```diff
@@ -0,0 +1,294 @@
...
+    "mcp-mpod": "SA3_XTD10_MCP/MCPS/MPOD",
+    "mcp-voltage": ["SA3_XTD10_MCP/MCPS/MPOD", "channels.U3.voltage"],
+    "mcp-trace": ["SA3_XTD10_MCP/ADC/1:channel_5.output", "data.rawData"],
+
+    "bogus-source": "SA4_XTD20_XGM/DOOCS/MAIN",
+    "bogus-key": ["SA1_XTD2_XGM/DOOCS/MAIN", "foo"]
+}
+    ''')
+
+    run = H5File(mock_sa3_control_data).with_aliases(mock_sa3_control_aliases)
```
takluyver, line 97:
```suggestion
    run = H5File(mock_sa3_control_data).with_aliases(aliases_path)
```

I just noticed that codecov shows the file-reading code is not being exercised. Same for the other two tests.

extra_data/tests/test_aliases.py
```diff
@@ -0,0 +1,294 @@
...
+bogus-source = "SA4_XTD20_XGM/DOOCS/MAIN"
+bogus-key = ["SA1_XTD2_XGM/DOOCS/MAIN", "foo"]
+
+["SA3_XTD10_XGM/XGM/DOOCS"]
+hv = "pulseEnergy.wavelengthUsed"
+beam-x = "beamPosition.ixPos"
+beam-y = "beamPosition.iyPos"
+    ''')
+
+    run = H5File(mock_sa3_control_data).with_aliases(mock_sa3_control_aliases)
```
takluyver, line 153:
```suggestion
    run = H5File(mock_sa3_control_data).with_aliases(aliases_path)
```

PR #370, opened 2023-01-17
Add support for LPD Mini

cydanil:
To add support for the LPD Mini, the LPD1M class was modified to a `LPDBase` containing all the functionalities for parallel gain processing, and two subclass were added for the LPD 1M and Mini.  

The added Mini class can be tested as follows:

```python
from pathlib import Path
import extra_data as xd
import extra_geom as xg
import numpy as np
from extra_data.components import LPDMINI

dc = xd.DataCollection.from_paths([Path("/gpfs/exfel/exp/FXE/202231/p900316/raw/r0010/RAW-R0010-LPDMINI00-S00000.h5")]).select_trains(np.s_[0])

det = LPDMINI(dc, detector_name="FXE_DET_LPD_MINI")
data = det.get_array('image.data')

geom = xg.LPD_MiniGeometry.from_origin()

geom.plot_data(data[:, 0, 0])
```

This is still a draft as tests are being made, and documentation has yet to be updated.  

Thanks,  
Cyril


extra_data/components.py
```diff
@@ -1394,6 +1374,54 @@
         )
 
 
+@multimod_detectors
+class LPD1M(LPDBase, XtdfDetectorBase):
```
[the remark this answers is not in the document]
cydanil: I resolved this by explicitly setting `module_shape` in theses classes.

cydanil:
Very good point, indeed updating thr regex could be enough.  
However, it would break usability in runs containing both detectors (of which there will be lots of). 
With this approach, it will still be possible to do `LPD1M(run)` and get the correct data for the correct detector.

extra_data/components.py
```diff
@@ -1394,6 +1373,56 @@ def _make_image_index(self, tids, inner_ids, inner_name='pulse'):
...
```
takluyver, line 1377: Does it work for the inheritance to go in a chain (XtdfDetectorBase -> LPDBase -> (LPD1M, LPDMINI)) rather than LPDBase being a mixin class? If it works, I generally find code easier to think about without multiple inheritance involved.

takluyver:
That is a good question. From my point of view, it's a trade-off between simplicity for us - maintaining fewer class definitions - and convenience for typical users - being able to access e.g. LPD1M without having to remember precisely how its name is spelled (`FXE_DET_LPD1M-1`).

For JUNGFRAU, it was clear there would be quite a few different configurations, so it seemed more practical to make one class to handle them all. We might still one day decide to add specific shortcuts for e.g. JUNGFRAU 1M.

Plus there's the backwards compatibility, as Cyril mentioned. We can't remove the LPD1M class, and having it find LPDMini as well would be confusing (it's not part of the 1M detector), so if we added a flexible class, it would still be in addition to the LPD1M class.

extra_data/components.py
```diff
@@ -1394,6 +1373,56 @@ def _make_image_index(self, tids, inner_ids, inner_name='pulse'):
...
+      and low gain data are stored sequentially within each train. This will
+      repeat the pulse & cell IDs from the first 1/3 of each train, and add gain
+      stage labels from 0 (high-gain) to 2 (low-gain).
+    """
+    _source_re = re.compile(r'(?P<detname>.+_LPD1M.*)/DET/(?P<modno>\d+)CH')
+    module_shape = (256, 256)
+
+
+@multimod_detectors
+class LPDMINI(LPDBase, XtdfDetectorBase):
```
takluyver, line 1403: Is it common to write LPD MINI in all capitals? If not, let's call the class `LPDMini`.

JUNGFRAU is in block capitals because [that's how it's written](https://www.psi.ch/en/lxn/jungfrau) - it's an incredibly tortured acronym.

extra_data/components.py
cydanil: Not really: if `LPDBase` inherits from `XtdfDetectorBase`, then it becomes findeable by `identify_multimod_detectors` as a subclass, and breaks the functions that attempt to identify the detector by name and comes up first instead of LPD1M or LPDMini.  

An alternative would be to define the `LPDBase(XtdfDetectorBase)` below the LPD1M and Mini classes :/

extra_data/components.py
cydanil: I see, I went with all capitals as other detectors we all capitals.  
I'm happy to make it camelCased.

extra_data/components.py
turkot: With the current implementation `identify_multimod_detectors` searches only among classes decorated with `@multimod_detectors`, so if you don't decorate `LPDBase` it will not be findable.

cydanil: Yep, you're right, I misspoke.  
It's [`_iter_trains`](https://github.com/European-XFEL/EXtra-data/blob/master/extra_data/export.py#L72) that iterates through the subclasses and breaks, where `LPDBase` needn't be in the list altogether.

An example is [`test_streamer.test_merge_detector`](https://github.com/European-XFEL/EXtra-data/blob/master/extra_data/tests/test_streamer.py#L13):



```python
mock_fxe_raw_run = '/tmp/tmpqpt42uiw', mock_fxe_control_data = '/tmp/tmpcykco1sb/RAW-R0450-DA01-S00001.h5', mock_spb_proc_run = '/tmp/tmpykz76qep'

    def test_merge_detector(mock_fxe_raw_run, mock_fxe_control_data, mock_spb_proc_run):
        with RunDirectory(mock_fxe_raw_run, include="*LPD[!MINI]*") as run:
>           for tid, data in _iter_trains(run, merge_detector=True):

test_streamer.py:15: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../export.py:72: in _iter_trains
    det = detector(data)
../components.py:1296: in __init__
    super().__init__(data, detector_name, modules, min_modules=min_modules)
../components.py:501: in __init__
    super().__init__(data, detector_name, modules, min_modules=min_modules)
../components.py:113: in __init__
    source_to_modno = self._identify_sources(data, detector_name, modules)
../components.py:176: in _identify_sources
    source_to_modno = dict(self._source_matches(data, detector_name))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <[AttributeError("'LPDBase' object has no attribute 'detector_name'") raised in repr()] LPDBase object at 0x2b1366583520>, data = <extra_data.DataCollection for 16 sources and 480 trains>, detector_name = 'FXE_DET_LPD1M-1'

    def _source_matches(self, data, detector_name):
        for source in data.instrument_sources:
            m = self._source_re.match(source)
            if m and m.group('detname') == detector_name:
>               yield source, int(m.group('modno'))
E               IndexError: no such group

../components.py:173: IndexError
```

takluyver: I see, thanks. I think probably the right thing to do there is to use `identify_multimod_detectors` instead of `XtdfDetectorBase.__subclasses__`. It's precisely the sort of scenario that function is there for, but we've never got round to looking at it and figuring out if JUNGFRAU needs something slightly different.

However, that bit of refactoring doesn't need to be in this PR either.

takluyver:
LGTM :+1:

turkot:
LGTM as well :+1:

turkot (approved):

cydanil:
Sweet, thank you both for your time!

PR #378, opened 2023-02-27
Update Sphinx & related packages on RTD

takluyver:
To see if this fixes search on the docs.


takluyver:
That seems to have done the trick.

PR #379, opened 2023-03-07
Add first_modno input argument for components.JUNGFRAU

kakhahmed:
This MR fixes issues with:
 - Plotting xarray for FXE_XAD_JF500K which has one module `JNGFR03`.
 - Overwrite and correcting `modno_to_source` and `source_to_modno` with the correct `modno` corresponding to the module array index.

This fix will used by https://git.xfel.eu/calibration/pycalibration/-/merge_requests/775 [Not yet tested with all reference runs.] It was tested with SPB_IRDA_JF4M and FXE_XAD_JF500K

And it is an alternative for https://github.com/European-XFEL/EXtra-geom/pull/206

This issue was discovered after creating `components.JUNGFRAU` object for `FXE_XAD_JF500K` to form a DataArray for `data.adc`, then plotting it using `plot_data_fast`. The problem is that the module coordinates for this DataArray are of a value `3`. As the data channel has JNGFR03.

Screenshot-from-2023-02-14-12-18-18

Both `modno_to_source` and `source_to_modno` are modified when st_modno is set to > 0 value.
This helps in fixing the cases mentioned below in the comments with JF detectors with one module  and receiver with integer > 1

Tested were added for `st_modno` and `mock_fxe_jungfrau_run` was added for /gpfs/exfel/exp/FXE/202101/p002478/raw/r0052

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -1446,9 +1446,9 @@ class JUNGFRAU(MultimodDetectorBase):
     n_modules: int
       Number of detector modules in the experiment setup. Default is
       None, in which case it will be estimated from the available data.
-    st_modno: int
+    first_modno: int
       The module number that the detector starts with.
-      e.g. FXE_XAD_JF500K/DET/JNGFR03:daqOutput should has st_modno = 3
+      e.g. FXE_XAD_JF500K/DET/JNGFR03:daqOutput should has first_modno = 3
     """
     # We appear to have a few different formats for source names:
     # SPB_IRDA_JNGFR/DET/MODULE_1:daqOutput  (e.g. in p 2566, r 61)
@@ -1464,16 +1464,16 @@ class JUNGFRAU(MultimodDetectorBase):
     module_shape = (512, 1024)
 
     def __init__(self, data: DataCollection, detector_name=None, modules=None,
-                 *, min_modules=1, n_modules=None, st_modno=1):
+                 *, min_modules=1, n_modules=None, first_modno=1):
         super().__init__(data, detector_name, modules, min_modules=min_modules)
 
-        if st_modno:
+        if first_modno:
             self.modno_to_source = {}
             # Overwrite modno based on given starting module number and update
             # source_to_modno and modno_to_source.
             for source in self.source_to_modno.keys():
                 # JUNGFRAU modono is expected (e.g. extra_geom) to start with 1.
-                modno = int(self._source_re.search(source)['modno']) - st_modno + 1
+                modno = int(self._source_re.search(source)['modno']) - first_modno + 1
                 self.source_to_modno[source] = modno
                 self.modno_to_source[modno] = source
 
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -1480,7 +1480,7 @@ class JUNGFRAU(MultimodDetectorBase):
             self.n_modules = int(n_modules)
         else:
             # For JUNGFRAU modules are indexed from 1
-            self.n_modules = max(modno for (_, modno) in self._source_matches(
+            self.n_modules = max(modno - first_modno + 1 for (_, modno) in self._source_matches(
                 data, self.detector_name
             ))
 
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -1471,7 +1471,7 @@ class JUNGFRAU(MultimodDetectorBase):
         # Overwrite modno based on given starting module number and update
         # source_to_modno and modno_to_source.
         for source in self.source_to_modno.keys():
-            # JUNGFRAU modono is expected (e.g. extra_geom) to start with 1.
+            # JUNGFRAU modno is expected (e.g. extra_geom) to start with 1.
             modno = int(self._source_re.search(source)['modno']) - first_modno + 1
             self.source_to_modno[source] = modno
             self.modno_to_source[modno] = source
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -1477,7 +1477,7 @@ class JUNGFRAU(MultimodDetectorBase):
         self._frames_per_entry = self.data[src, self._main_data_key].entry_shape[0]
 
     @staticmethod
-    def _label_dims(arr):
+    def _label_dims_update_module(arr, detector_name=None):
         # Label dimensions to match the AGIPD/DSSC/LPD data access
         ndim_pertrain = arr.ndim
         if 'trainId' in arr.dims:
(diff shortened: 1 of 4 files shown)


kakhahmed:
I am not sure what the reason the codecov/project failed check

kakhahmed:
I am not sure anymore if this is the right approach.

There are more examples of this with other detector names.


Screenshot from 2023-03-09 11-51-08

Screenshot from 2023-03-09 11-50-53

Screenshot from 2023-03-09 11-50-39

I know that this is the karabo_da shown here in CALCAT but the detector's receiver sources are named based on it as far as I know.

Screenshot from 2023-03-09 11-57-24

Screenshot from 2023-03-09 11-57-13

takluyver:
Ugh :roll_eyes: . Do you know if this is only an issue with JF500K detectors? We could go back to your idea of applying it to all single module detectors. Or we could let the user pass in something like `modnum_minus=2`.

Or if the user specifies `n_modules=4` or similar and we have 4 modules, we could automatically renumber them 1, 2, 3, 4. The only reason we can't do that without `n_modules` is that if we see only `JNGFR02`, we assume that it's a 2-module detector and the first module is missing from the data.

kakhahmed:
It is with JF500K detectors for now. But there is no guarantee that this will be the case always. 

Also, adding more info to this. The below screenshot shows that in offline calibration we are dependent on that DA/Receiver relation.

Screenshot from 2023-03-09 17-55-57


> Or if the user specifies n_modules=4 or similar and we have 4 modules, we could automatically renumber them 1, 2, 3, 4. The only reason we can't do that without n_modules is that if we see only JNGFR02, we assume that it's a 2-module detector and the first module is missing from the data.

I always think about this solution. The only thing I couldn't solve is:


For example, we have JF4M -> JNGFR02 & JNGFR03 & JNGFR04 & JNGFR05.

The user passes n_modules = 4 and we see only JNGFR02 how would we know that this is the 1st detector?
Maybe we add a new argument for the first source name? e.g. `first_receiver_name = "JNGFR02"`

Passing `n_modules` and `first_receiver_name` might solve this case.

takluyver:
> The user passes n_modules = 4 and we see only JNGFR02 how would we know that this is the 1st detector? Maybe we add a new argument for the first receiver name? e.g. first_receiver_name = "JNGFR02"

Yeah, if there are missing modules, you'd have to specify where the numbers start from. I'd probably do it with passing an integer rather than a string, because this 'JNGFR02' part isn't something that EXtra-data otherwise asks you to look at.

kakhahmed:
> I'd probably do it with passing an integer rather than a string, because this 'JNGFR02' part isn't something that EXtra-data otherwise asks you to look at.

Yes, that makes sense 👍🏽

kakhahmed:
I have replaced the changes in the MR by adding `st_modno` input argument to be able to point to the components.JUNGFRAU on the first `modno` for a JF detector.

extra_data/components.py
```diff
@@ -1446,6 +1446,9 @@ class JUNGFRAU(MultimodDetectorBase):
     n_modules: int
       Number of detector modules in the experiment setup. Default is
       None, in which case it will be estimated from the available data.
+    st_modno: int
```
takluyver, line 1449: I think `st` is a bit too cryptic (and it makes me think of the [results from stat()](https://docs.python.org/3/library/os.html#os.stat_result)). Let's make it a bit longer, like `start_modno` or `first_modno`.

extra_data/components.py
```diff
@@ -1446,6 +1446,9 @@ class JUNGFRAU(MultimodDetectorBase):
...
+      The module number that the detector starts with.
```
takluyver, line 1450:
```suggestion
      The module number in the source name  for the first detector module.
```

extra_data/components.py
```diff
@@ -1461,9 +1464,19 @@ class JUNGFRAU(MultimodDetectorBase):
     module_shape = (512, 1024)
 
     def __init__(self, data: DataCollection, detector_name=None, modules=None,
-                 *, min_modules=1, n_modules=None):
+                 *, min_modules=1, n_modules=None, st_modno=1):
         super().__init__(data, detector_name, modules, min_modules=min_modules)
 
+        if st_modno:
```
takluyver, line 1470: This branch will always run, right? Unless someone makes a `JNGFR00`, but in that case it should run, because we want JUNGFRAU numbering to start from 1 for EXtra-geom.

So I think we can lose the `if`.

extra_data/components.py
```diff
@@ -1461,9 +1464,19 @@ class JUNGFRAU(MultimodDetectorBase):
...
+            # source_to_modno and modno_to_source.
+            for source in self.source_to_modno.keys():
+                # JUNGFRAU modono is expected (e.g. extra_geom) to start with 1.
+                modno = int(self._source_re.search(source)['modno']) - st_modno + 1
+                self.source_to_modno[source] = modno
+                self.modno_to_source[modno] = source
+
         if n_modules is not None:
             self.n_modules = int(n_modules)
         else:
```
takluyver, line 1482: I think this branch (estimating `n_modules`) needs to be adjusted too.

takluyver (commented):
I think this looks like a good direction, my comments are all on the details.

Sorry it's taken so long to look at it again.

takluyver:
Oh, and I meant to say, don't worry too much about codecov - it often seems to report that PRs cause a small drop in coverage. I imagine there's some fiddly issue with coverage collection, but so far I haven't bothered to figure it out.

extra_data/components.py
kakhahmed: I like `first_modno` [4ace38c02a92a8aa7db6ac363f9865cbf4386565](https://github.com/European-XFEL/EXtra-data/pull/379/commits/4ace38c02a92a8aa7db6ac363f9865cbf4386565)

extra_data/components.py
kakhahmed: Thank you. I updated the description. [9897575bd59989213d70931f09dde5c67d390841](https://github.com/European-XFEL/EXtra-data/pull/379/commits/9897575bd59989213d70931f09dde5c67d390841)

extra_data/components.py
kakhahmed: Good point, this condition is removed. [37cfa0874adac26d2df26f8a4c429ab745111aaa](https://github.com/European-XFEL/EXtra-data/pull/379/commits/37cfa0874adac26d2df26f8a4c429ab745111aaa)

extra_data/components.py
kakhahmed: Nice catch. I fixed this branch when `n_modules is None` [77b88ec9ed1944e746bcd979a917e0f10d45abf6](https://github.com/European-XFEL/EXtra-data/pull/379/commits/77b88ec9ed1944e746bcd979a917e0f10d45abf6)

takluyver (approved):
Thanks @kakhahmed !

PR #390, opened 2023-04-04
Expose srcdata.device_class property

takluyver:
Make it easier to get the Karabo class name for CONTROL sources. This will allow things like:

```python
xgms = [src for src in run.control_sources if run[src].device_class == 'DoocsXGM']
```

I've added an `allow_multi_run` parameter to `.run_value()` for things that we don't expect to change even between runs, like this. It feels a bit clumsy, though. Another option would be to drop the is_single_run check and trust that people understand that run values on the union of two runs may not be reliable. :thinking:

### DIFF
--- a/extra_data/sourcedata.py
+++ b/extra_data/sourcedata.py
@@ -11,10 +11,12 @@ from .read_machinery import glob_wildcards_re, same_run, select_train_ids
 
 
 class SourceData:
-    """Data for one key in one source
+    """Data for one source
 
     Don't create this directly; get it from ``run[source]``.
     """
+    _device_class = ...
+
     def __init__(
             self, source, *, sel_keys, train_ids, files, section,
             is_single_run, inc_suspect_trains=True
@@ -222,7 +224,7 @@ class SourceData:
 
         return self.files[0].metadata()
 
-    def run_value(self, key):
+    def run_value(self, key, *, allow_multi_run=False):
         """Get a single value from the RUN section of data files.
 
         This method is intended for use with data from a single run. If you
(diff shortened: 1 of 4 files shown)


extra_data/sourcedata.py
```diff
@@ -11,10 +11,12 @@
 
 
 class SourceData:
-    """Data for one key in one source
+    """Data for one source
 
     Don't create this directly; get it from ``run[source]``.
     """
+    _device_class = ...
```
philsmt, line 18: Why the choice of the ellipsis as a default / unknown value?

philsmt:
I'm not convinced "leave it up to the user" is the right approach here, as we should keep in mind this library in particular may often be used by someone not fully grasping the entire system.
How about we actually "do the right thing" and compare all `RUN` values across the entire `DataCollection`, and just shortcut this expensive operation when we are certain it's from a single run? (in a different MR of course)

extra_data/sourcedata.py
takluyver: I wanted to use None to represent 'not available', so I needed something else for 'haven't looked yet'. `...` is kind of a funny option, but I'm happy to use `False` or something if that's clearer.

takluyver:
That could work. I think we didn't do this before because we didn't want to track the runs as groups within a DataCollection, but we could look at every file related to a source.

philsmt:
LGTM!

Concerning the `allow_multi_run`, we had a brief discussion in person that we want to move to a proper `is_single_run` property that evaluates the "singleness" of a run and/or evaluate the uniqueness of specific run values, but still keep this particular flag in as it may prove useful in any case.

PR #391, opened 2023-04-05
FIX: write metadata in voview files

tmichela:
current version is 1.2

### DIFF
--- a/extra_data/writer.py
+++ b/extra_data/writer.py
@@ -117,7 +117,7 @@ class FileWriter:
             metadata = {}
 
         metadata_grp = self.file.create_group('METADATA')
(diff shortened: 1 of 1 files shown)


philsmt:
I'm not sure it's safe to assume the version string will remain float-compatible, we do not have immediate control over it and someone may do `x.y.z`, raising a `ValueError`. We discussed this in https://github.com/European-XFEL/EXtra-data/pull/332#discussion_r951661060 and came up with `packaging.version`, which is cheaper to use than `pkg_resources.parse_version`.

tmichela:
Ah I missed the other PR, will close that one

philsmt:
:see_no_evil: 

I knew that of course when I wrote the comment!

PR #393, opened 2023-04-06
Fix search in RTD docs

takluyver:
With a workaround suggested in https://github.com/readthedocs/sphinx_rtd_theme/issues/1452


takluyver:
Yup, that seems to have done it: https://extra-data--393.org.readthedocs.build/en/393/search.html?q=union&check_keywords=yes&area=default

PR #394, opened 2023-04-17
Move components to their own package

philsmt:
The current components module is already quite heavy at more than 1700 lines and we plan to add more components in the near future covering more than multi-modular frame-based detectors. This PR turns `extra_data.components` from a module into a package and moves the existing components into a module `extra_data.components.multimod_det`.

To preserve the existing public API, the package initializer star imports this module (and I would suggest to add any future components modules in the same way). It is technically possible for code to break that imported some not in `extra_data.components.multimod_det.__all__`. We could work around that with a `__getitem__` handler if you feel strongly about it.

### DIFF
--- a/extra_data/components/multimod_det.py
+++ b/extra_data/components/multimod_det.py
@@ -1,4 +1,4 @@
-"""Interfaces to data from specific instruments
+"""High-level interfaces for multi-module frame-based detectors
 """
 import logging
 import re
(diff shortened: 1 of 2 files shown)


extra_data/components/multimod_det.py
```diff
@@ -8,13 +8,14 @@
...
-from .writer import FileWriter
-from .write_cxi import XtdfCXIWriter, JUNGFRAUCXIWriter
+from ..exceptions import SourceNameError
+from ..reader import DataCollection, by_id, by_index
+from ..read_machinery import DataChunk, roi_shape, split_trains
+from ..writer import FileWriter
+from ..write_cxi import XtdfCXIWriter, JUNGFRAUCXIWriter
 
 __all__ = [
+    'XtdfDetectorBase',
```
philsmt, line 18: I added this class here, as it was the only symbol used outside of the `components` and in particular also referenced on docstrings there.

JamesWrigley:
LGTM, because I need this too :stuck_out_tongue: I doubt it's necessary to define a custom `__getitem__()`, if someone's using internal things then they shouldn't expect it to be stable.

philsmt:
Based on the discussion to move `components` entirely to EXtra, in particular the new ones, I'm closing this.

PR #395, opened 2023-04-17
An interface for timing of FEL and PPL pulses

philsmt:
This is a component providing high-level access to timing data of pulses. In the spirit of the existing components for multi-module detectors, it tries to be clever about picking the sources automatically, mask compatibility issues and offer prepare data in a more useful format than the one it may be saved in. The functionality that is so far implemented:

* Find timeserver and BAM devices in data, optionally with user help
* Figure out which SASE the data belongs to
* Pick the correct indices from the bunch pattern table, either as a constant pattern or search for contiguous portion of identical patterrs
* Use the pattern above to slice out the correct arrival times from BAM data
* Return data in the form of an iterator, ndarray or labeled xarray

For now it doesn't actually look at the PPL pulse indices, as depending on the pattern this can be somewhat tricky.

I'm opening this PR early to get some initial feedback, as there are quite some firsts when it comes to this interface with it wrapping more than a single source (or group of sources). It also targets the `feat/components-package` branch to keep the diff smaller.

### DIFF
--- a/extra_data/components/__init__.py
+++ b/extra_data/components/__init__.py
@@ -1,2 +1,3 @@
 
+from .auxiliary import *
 from .multimod_det import *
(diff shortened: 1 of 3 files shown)


philsmt:
To be moved to EXtra, closing this here.

PR #399, opened 2023-04-25
Expose filename-based traits in FileAccess and SourceData

philsmt:
Possibly a niche feature, but I've been wanting something like this at times. If you don't mind, I think EXtra-data is a nice place to encode such a detail.

Do you think sources changing between aggregators in a single run / collection is important enough to make a new fixture and test?

### DIFF
--- a/extra_data/sourcedata.py
+++ b/extra_data/sourcedata.py
@@ -5,7 +5,7 @@ from typing import Dict, List, Optional
 import numpy as np
 import h5py
 
-from .exceptions import MultiRunError, PropertyNameError
+from .exceptions import MultiRunError, PropertyNameError, NoDataError
 from .file_access import FileAccess
 from .keydata import KeyData
 from .read_machinery import glob_wildcards_re, same_run, select_train_ids, split_trains
@@ -17,6 +17,7 @@ class SourceData:
     Don't create this directly; get it from ``run[source]``.
     """
     _device_class = ...
+    _first_source_file = ...
 
     def __init__(
             self, source, *, sel_keys, train_ids, files, section,
@@ -78,6 +79,38 @@ class SourceData:
             eshape=ds0.shape[1:],
             inc_suspect_trains=self.inc_suspect_trains,
         )
+
+    def _get_first_source_file(self):
+        first_kd = self[next(iter(self.keys()))]
+
+        try:
+            # This property is an empty list if no trains are selected.
+            sample_path = first_kd.source_file_paths[0]
+        except IndexError:
+            raise NoDataError(self.source) from None
+
+        return FileAccess(sample_path)
+
+    @property
+    def storage_class(self):
+        if self._first_source_file is ...:
+            self._first_source_file = self._get_first_source_file()
+
+        return self._first_source_file.storage_class
+
+    @property
+    def data_category(self):
+        if self._first_source_file is ...:
+            self._first_source_file = self._get_first_source_file()
+
+        return self._first_source_file.data_category
+
+    @property
+    def aggregator(self):
+        if self._first_source_file is ...:
+            self._first_source_file = self._get_first_source_file()
+
+        return self._first_source_file.aggregator
 
     def keys(self, inc_timestamps=True):
         """Get a set of key names for this source
--- a/extra_data/tests/test_sourcedata.py
+++ b/extra_data/tests/test_sourcedata.py
@@ -2,7 +2,7 @@ import numpy as np
 import pytest
 
 from extra_data import RunDirectory, by_id, by_index
-from extra_data.exceptions import PropertyNameError, SourceNameError
+from extra_data.exceptions import PropertyNameError, SourceNameError, NoDataError
 
 
 def test_get_sourcedata(mock_spb_raw_run):
@@ -138,6 +138,21 @@ def test_device_class(mock_spb_raw_run):
 
     xgm_inst = run['SPB_XTD9_XGM/DOOCS/MAIN:output']
     assert xgm_inst.device_class is None
+
+
+def test_path_traits(mock_spb_raw_run):
+    run = RunDirectory(mock_spb_raw_run)
+    xgm = run['SPB_XTD9_XGM/DOOCS/MAIN']
+
+    assert xgm.storage_class is None  # Not an EuXFEL path.
+    assert xgm.data_category == 'RAW'
+    assert xgm.aggregator == 'DA01'
+
+    run = RunDirectory(mock_spb_raw_run).select_trains(np.s_[:0])
+    xgm = run['SPB_XTD9_XGM/DOOCS/MAIN']
+
+    with pytest.raises(NoDataError):
+        xgm.storage_class
 
 
 @pytest.mark.parametrize('source', [


takluyver:
I think that seems reasonable, but I don't like that it will break if you open data with a virtual overview file. Maybe we need to finish #325 first?

philsmt:
Oh yes, I meant to add a comment changing this when needed. In general any order you wish.

philsmt:
I've rebased the PR on top of https://github.com/European-XFEL/EXtra-data/pull/325 and generalized the support to various data called traits gleaned from the (parent) path and the filename. The separation serves to allows the properties to continue working when files are moved to a different non-standard location, but retain their original filename.

Rather than taking sets across all files, only the first file for the _first_ key of a `SourceData` object is used. Since we currently cannot really load sources from multiple overlapping files, it is probably not worth the effort to check.

extra_data/file_access.py
```diff
@@ -122,8 +123,13 @@ class FileAccess(metaclass=MetaFileAccess):
     """
     _file = None
     _format_version = None
+    _path_traits = None
+    _filename_traits = None
     metadata_fstat = None
 
+    exdf_path_pattern = re.compile(r'\/(gpfs\/exfel\/d|pnfs\/xfel.eu\/exfel\/archive\/XFEL)\/([a-z]+)\/(\w+)\/(\d{6})\/p\d{6}\/r\d{4}')
```
takluyver, line 130: This won't work on ONC, where the paths really are under `/gpfs/exfel/exp`, with the components in a different order from this.

I'm OK with this - clearly the path based info can't always work, because we allow opening files & folders from anywhere. So I just want to ensure we're making a conscious choice, not expecting something to work which won't.

extra_data/sourcedata.py
```diff
@@ -15,6 +15,8 @@ class SourceData:
 
     Don't create this directly; get it from ``run[source]``.
     """
+    _first_source_file = ...
```
takluyver, line 18: Welcome to team `...` :grinning: 

I tend to only do this if `None` is a possible value in the cache. But I've no objection to using `...` even if None would work.

extra_data/sourcedata.py
```diff
@@ -70,6 +72,31 @@ def __getitem__(self, key):
             inc_suspect_trains=self.inc_suspect_trains,
         )
 
+    def _get_first_source_file(self):
+        sample_path = self[next(iter(self.keys()))].source_file_paths[0]
```
takluyver, line 76: If there are no trains selected, `.source_file_paths` will be an empty list, in which case this will raise IndexError. This is unlike `sourcedata.files`, which we ensure always keeps at least one file even if you deselect all trains.

Are you happy for these properties to just not work with no data selected? Fixing it feels like a lot of effort for a corner case that might well not be important.

extra_data/sourcedata.py
```diff
@@ -70,6 +72,31 @@ def __getitem__(self, key):
...
+            self._first_source_file = self._get_first_source_file()
+
+        return self._first_source_file.storage_class
+
+    @property
+    def file_prefix(self):
+        if self._first_source_file is ...:
+            self._first_source_file = self._get_first_source_file()
+
+        return self._first_source_file.file_prefix
```
takluyver, line 91: Do we have a use case for this one? 'File prefix' seems vague enough that I'm inclined to leave it out.

takluyver (commented):
Thanks! I've got a few comments on specific details.

I wouldn't refer to these as 'traits' in public docs or names, because [traits](https://docs.enthought.com/traits/) is a package for doing fancy things with Python attributes (also [traitlets](https://traitlets.readthedocs.io/en/stable/), a variant on the theme which is used in IPython & Jupyter). Using it internally isn't an issue, though.

extra_data/sourcedata.py
philsmt: In general I agree, but I did this in preparation and to be symmetric with `_device_class` 😉

extra_data/sourcedata.py
philsmt: I would say yes, but thanks for the hint, I will add a custom exception text that makes it explicit why it happened.

extra_data/sourcedata.py
philsmt: Inspired by the introduction of a data category for files in the SDP draft which can move through storage classes, I would like to use this field for this purpose.

It addresses the problem that using it for the actual storage class would mean renaming the files as they are copied (say from `raw` to `red`) as well as lose a distinction about their origin. At the same time, it can immediately be used to indicate the result of a reduction, say by prefixing the output of a `RAW-` file with `RRAW-`.

We can drop it if you consider it too vague for now, but I have need of it in the data reduction toolchain.

philsmt:
> I wouldn't refer to these as 'traits' in public docs or names [...]

I do not have any serious feelings about that particular word. I wanted to convey the meaning of _property_ (as in properties of the path) without using that term, and trait is what I got from a thesaurus 🤷 
Indeed I don't foresee any need for public documentation, but I'm happy to use another.

extra_data/sourcedata.py
philsmt: Added a custom `ValueError` indicating there are no trains to pull source files from.

extra_data/tests/test_sourcedata.py
```diff
@@ -140,6 +140,21 @@ def test_device_class(mock_spb_raw_run):
...
+    xgm = run['SPB_XTD9_XGM/DOOCS/MAIN']
+
+    assert xgm.storage_class is None  # Not an EuXFEL path.
+    assert xgm.data_category == 'RAW'
+    assert xgm.aggregator == 'DA01'
+
+    run = RunDirectory(mock_spb_raw_run).select_trains(np.s_[:0])
+    xgm = run['SPB_XTD9_XGM/DOOCS/MAIN']
+
+    with pytest.raises(NoDataError):
```
takluyver, line 154: Looks like the code is still raising ValueError.

philsmt: Yes, I'm sorry for the commit spam, I don't seem to be fully concentrated. I happened to be in `tests/` and didn't commit the changes to `../sourcedata.py`.

extra_data/tests/test_file_access.py
```diff
@@ -60,6 +60,51 @@ def test_pickle(pickle_mod, mock_sa3_control_data):
...
+        '/gpfs/exfel/exp/SA3/202301/p001234/raw/r0100/foo.h5',
+        '/gpfs/exfel/d/raw/SA3/202301/p001234/r0100/foo.h5',
+        '/pnfs/xfel.eu/exfel/archive/XFEL/raw/SA3/202301/p001234/r0100/foo.h5'
+    ]:
+        fa = FileAccess(mock_sa3_control_data)
+        fa.filename = filename
+
+        # FileAccess instances are always unique by filename, so the
+        # cache property needs to be cleared explicitly.
+        fa._path_traits = None
```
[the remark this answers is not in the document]
philsmt: I agree this is way too brittle even for a test.
Apart from your suggested refactoring, I managed to get it working without cheating the metaclass via the `_cache_info` keyword argument to `FileAccess`. While it is a private API, it is clearly meant for the specific purpose of avoiding an actual read when constructing the object. Using this I can construct `FileAccess` objects with any non-existing path and run through the actual code I want to test.

PR #400, opened 2023-05-04
Add require_any keyword argument to source selection

philsmt:
Complementing the [earlier support](https://github.com/European-XFEL/EXtra-data/pull/113)  to match data during a selection, this PR adds a `require_any` keyword argument as a weaker version of `require_all`. Rather than requiring all source to have data for each train, it only requires one source to have data for each train. The use cases for this are definitely smaller, but it seemed trivial to add to the existing `require_all` code by tweaking the logic only slightly.

Also used the opportunity to fix an oversight from https://github.com/European-XFEL/EXtra-data/pull/367 with `require_all` not actually being forwarded from `run.alias.select()` to `run.select()`.

An open question is whether this should be ported to `run.trains()` as well.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -909,7 +909,8 @@ class DataCollection:
                              .format((source_glob, key_glob)))
         return matched
 
-    def select(self, seln_or_source_glob, key_glob='*', require_all=False):
+    def select(self, seln_or_source_glob, key_glob='*', require_all=False,
+               require_any=False):
         """Select a subset of sources and keys from this data.
 
         There are four possible ways to select data:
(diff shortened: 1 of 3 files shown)


extra_data/reader.py
```diff
@@ -969,17 +970,21 @@ def select(self, seln_or_source_glob, key_glob='*', require_all=False):
...
+        if require_all or require_any:
+            # Select only those trains for which all (require_all) or at
+            # least one (require_any) selected sources and keys have
+            # data, i.e. have a count > 0 in their respective INDEX
+            # section.
 
-            train_ids = self.train_ids
+            if require_all:
+                train_ids = self.train_ids
+            else:
```
philsmt:
```suggestion
            else:  # require_any
```

extra_data/tests/test_reader_mockdata.py
```diff
@@ -607,6 +608,23 @@ def test_select_require_all(mock_sa3_control_data, select_str):
     assert all([isinstance(x, np.uint64) for x in subrun.train_ids])
 
 
+def test_select_require_any(mock_sa3_control_data):
+    # See above, de-select all sources without data.
+    run = H5File(mock_sa3_control_data) \
+        .deselect([('SA3_XTD10_MCP/ADC/1:*', '*'),
+                   ('SA3_XTD10_IMGFEL/CAM/BEAMVIEW:*', '*')])
```
philsmt: These deselections aren't needed here.

takluyver:
Reviewed in person, LGTM

PR #403, opened 2023-05-16
Move SourceData.is_control to a public API

philsmt:
In recent tooling, I frequently needed to distinguish between `CONTROL` and `INSTRUMENT` sources. The `SourceData` API already has a corresponding property, so it seems harmless to move this to be a public API. Incidentally I also added an accompanying `SourceData.is_instrument` property and covered them by tests.


tmichela:
I would not mention "fast/slow data" as this is Karabo jargon (and they actually don't want to advertise these terms as they are confusing to non experts, and prefer "pipeline" / "control" -or broker- data).
My preference is to just leave it as "control" and "instrument" data here as this is what's in the files.

philsmt:
I don't disagree, though I put that monikers there explicitly to potentially make the connection easier for staff. I suspect users coming from a less Karabo-centric perspective will find the `CONTROL`/`INSTRUMENT` terms more familiar. I don't have strong feelings about it though and am happy to remove it.

takluyver:
The API seems fine to me. I don't have a strong preference on the docstrings, but if controls don't want to talk about fast/slow data publicly, let's get rid of those terms.

philsmt:
I happened to be able to ask Steffen, and he did not have a strong opinion on it. For ease of discussion, I removed it from the docstring.

takluyver:
LGTM

philsmt:
(fixed merge conflict due to whitespace fixes clashing with https://github.com/European-XFEL/EXtra-data/pull/404)

PR #404, opened 2023-05-16
Add SourceData.split_trains()

takluyver:
`DataCollection` and `KeyData` both have `.split_trains()` methods, so it's natural for SourceData to have this too.


takluyver:
Thanks for the review :smiley:

PR #406, opened 2023-05-17
Update Github actions versions

takluyver:
Github actions is giving deprecation notices saying:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/setup-python@v2, actions/cache@v1, codecov/codecov-action@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

I think all we need to do is use the current version of these actions.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,14 +15,14 @@ jobs:
 
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


takluyver:
Before: https://github.com/European-XFEL/EXtra-data/actions/runs/5000677051
After: https://github.com/European-XFEL/EXtra-data/actions/runs/5002596958

PR #407, opened 2023-05-19
Add SourceData.drop_empty_trains()

philsmt:
A natural and small addition building on top of https://github.com/European-XFEL/EXtra-data/pull/405, and the code is even almost identical to `KeyData.drop_empty_trains`.


takluyver:
LGTM

PR #409, opened 2023-05-23
Clearer name for former path traits

philsmt:
Sorry for having the idea after merging https://github.com/European-XFEL/EXtra-data/pull/399, but I did find a clearer (internal!) name for this feature to avoid using the debatable word _trait_. Incidentally I'd also like to change the regexp name to indicate it's about files saved at EuXFEL, not strictly in the EXDF format.


takluyver:
Thanks, LGTM

philsmt:
Thank you!

PR #410, opened 2023-05-26
Improve AliasIndexer.__repr__() when there are no aliases loaded

JamesWrigley:
Another silly bug, previously it would just print out 'Loaded aliases:'.


takluyver:
Thanks, LGTM

PR #411, opened 2023-05-26
Add documentation for aliases

JamesWrigley:
This required exporting `AliasIndexer` for the links to work properly.


JamesWrigley:
I forgot to document some methods :flushed: Added in 63c39f0.

philsmt:
Thank you very much James :pray: 

This looks great! I tried to somehow hide any mentioning of `AliasIndexer` but didn't manage.  It does also work using `.. property:: alias` and referencing `.. automethod:: AliasIndexer.select` directly, but just to have it show up there instead. Not sure this is an actual improvement :shrug:

philsmt:
What do you think of going this route though and adding an additional string in the top of `AliasIndexer.select()`'s docstring with something like "This method is accessible through `DataCollection.alias`" or something?

JamesWrigley:
Sure, that looks good to me too :+1: Fixed in dd1efbb.

philsmt:
This is an even better solution, thanks! LGTM

PR #414, opened 2023-06-06
Add optional dependency groups

tmichela:


tmichela:
I went with adding a "extra_data[complete]" including all known optional deps.

setup.py
```diff
@@ -63,24 +62,27 @@ def find_version(*parts):
               'karabo-bridge >=0.6',
               'psutil',
           ],
+          'complete': [
+              'cloudpickle',
```
takluyver, line 66: I don't think cloudpickle is used for anything apart from some tests - though it's a dependency of dask, so it doesn't make much difference in practice.

takluyver:
Thanks, LGTM

PR #415, opened 2023-06-06
Expose origin indices in FileAccess

philsmt:
EXDF-1.2 introduced the `INDEX/origin` dataset, which expands upon `INDEX/flag` by telling from which source a train "originates" through indexing into `METADATA/dataSources`. This PR exposes this information through the `FileAccess` API including a fallback for older files that lack this information.

Since it's not used internally at the moment, I kept it a simple property without caching.


philsmt:
I've changed my mind, without actual internal use or a public API it doesn't seem necessary. Also, I don't really feel comfortable making up an origin source for suspicious trains for the sake of offering this value.

PR #416, opened 2023-06-08
Ensure that train IDs are propagated in union()

JamesWrigley:
`SourceData` has an internal list of train IDs that are used in e.g. `__repr__()` and `data_counts()`. This ensures that the internal list is kept up-to-date with the global list of train IDs when taking a union of two `DataCollection`s (e.g. by `open_run(data="all")`).

### DIFF
--- a/extra_data/tests/test_reader_mockdata.py
+++ b/extra_data/tests/test_reader_mockdata.py
@@ -742,21 +742,36 @@ def test_train_timestamps_nat(mock_fxe_control_data):
 def test_union(mock_fxe_raw_run):
     run = RunDirectory(mock_fxe_raw_run)
 
-    sel1 = run.select('SPB_XTD9_XGM/DOOCS/MAIN', 'beamPosition.ixPos')
-    sel2 = run.select('SPB_XTD9_XGM/DOOCS/MAIN', 'beamPosition.iyPos')
+    xgm = "SPB_XTD9_XGM/DOOCS/MAIN"
+    camera = "FXE_XAD_GEC/CAM/CAMERA"
+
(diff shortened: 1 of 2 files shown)


extra_data/tests/test_reader_mockdata.py
```diff
@@ -742,22 +742,37 @@ def test_train_timestamps_nat(mock_fxe_control_data):
...
 
+    # Test union of different sources in different train selections
+    sel1 = run.select(xgm).select_trains(by_index[:5])
+    sel2 = run.select(camera).select_trains(by_index[-5:])
+    joined = sel1.union(sel2)
+    expected_tids = run.train_ids[:5] + run.train_ids[-5:]
+
+    assert joined.train_ids == expected_tids
+    assert joined[xgm].train_ids == expected_tids
+    assert joined[camera].train_ids == expected_tids
```
takluyver, line 774: Hmmm, I'll have to think about this one. Intuitively, it feels more correct for union-ing them to keep the separate train IDs which are selected for each, and not re-select them. But it also feels desirable that `run.train_ids == run[src].train_ids`, and this would be more consistent with going from source to key (KeyData objects are created on demand, rather than always existing).

JamesWrigley: Added some docs in b65658e.

JamesWrigley: (bump :) )

philsmt:
With the explicit mentioning in docs, LGTM from my side.

takluyver:
Thanks James!

PR #418, opened 2023-06-14
Remove LPD Mini component class (pending redesign)

takluyver:
The current component is wrong, based on when we thought LPD Mini was a single 256 x 256 pixel detector, and we're still working out some issues with the redesign. We don't want to keep delaying the release for this, so it was suggested that we remove the LPD Mini wrapper for now so we can release other improvements and spend more time on it.

See #381 for the work to add it back.

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -19,7 +19,6 @@ __all__ = [
     'AGIPD500K',
     'DSSC1M',
     'LPD1M',
-    'LPDMini',
     'JUNGFRAU',
     'identify_multimod_detectors',
 ]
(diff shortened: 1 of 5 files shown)


philsmt:
Looks like there's some documentation at least in [docs/lpd_data.ipynb](https://github.com/European-XFEL/EXtra-data/blob/rm-lpd-mini/docs/lpd_data.ipynb) and [docs/aipgd_lpd_data.rst](https://github.com/European-XFEL/EXtra-data/blob/rm-lpd-mini/docs/agipd_lpd_data.rst).

LGTM otherwise.

takluyver:
Thanks, good catch.

PR #419, opened 2023-06-15
Add publishing to PyPI from Github actions on tag

takluyver:
Trying the new ['Trusted publisher'](https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/) support to automatically upload packages when we push a tag to Github.


takluyver:
This will only be tested when I try to make a release, so I'm going to merge it. :crossed_fingers:

PR #426, opened 2023-07-04
Remove Pillow from constraints file for CI

takluyver:
This is not being installed anyway.


takluyver:
nvm, I was mistaken, it does get installed.

PR #430, opened 2023-07-16
Implement .nbytes, .size_mb, and .size_gb for KeyData

JamesWrigley:
Added because it can be useful to see how large a dataset will be in memory without actually loading it.

`.size_mb` and `.size_gb` are there because I wanted a way to quickly check what the size of a dataset is in human-readable units. I could also implement a `.size()` function that takes in a unit, but I'd like to keep `.size_mb` and `.size_gb` anyway for ease-of-use.

### DIFF
--- a/docs/reading_files.rst
+++ b/docs/reading_files.rst
@@ -177,6 +177,12 @@ below, e.g.::
 
    .. automethod:: data_counts
 
+   .. autoattribute:: nbytes
+
+   .. autoattribute:: size_mb
+
+   .. autoattribute:: size_gb
+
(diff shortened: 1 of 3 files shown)


philsmt:
Sounds useful and strightforward, LGTM. I doubt there's a need for any decimal prefix other than M and G, so taking in arbitrary prefixes seems over the top.

takluyver:
prnote: New attributes `.nbytes`, `.size_mb`` and `.size_gb` to conveniently see how much data is present for a given source & key.

PR #431, opened 2023-07-17
Add SourceData.train_id_coordinates()

philsmt:
Another method I found missing when building EXtra components.

In principle a straightforward implementation on top of the index group handling, but unlike for `SourceData.data_counts()` I decided to error out when no index group is passed **and** multiple are present **and** they differ in their data counts. The rationale here is that this method is advertised (in its `KeyData` version) to work alongside `KeyData.ndarray()`, and I do not want to break that assumption. Hence, in almost all cases something like `run['<xtdf-source'].train_id_coordinates()` will not work.

You could argue that people *mostly* want to go for `image`, but it didn't seem worth the risk to me. CMV 🤷 

@takluyver @JamesWrigley

### DIFF
--- a/extra_data/sourcedata.py
+++ b/extra_data/sourcedata.py
@@ -91,6 +91,17 @@ class SourceData:
 
         return FileAccess(sample_path)
 
+    def _get_index_group_sample(self, index_group):
+        if self.is_control and not index_group:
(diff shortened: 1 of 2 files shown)


philsmt:
Thanks for review!

PR #440, opened 2023-08-24
Include XTDF sources in lsxfel when details are enabled

philsmt:
The condensed summary section for multi-module detectors at the top of `lsxfel` is useful, but filtering out their sources among the actual instrument sources prevents using `--detail` on them.

This PR treats XTDF sources as regular instrument sources whenever `--detail` is passed / `details_for_sources` is not empty. The rationale is that when details are enabled, a more verbose and detailed output is anyway expected.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1197,14 +1197,6 @@ class DataCollection:
             span_txt = '{}.{}'.format(datetime.timedelta(seconds=seconds),
                                       int(deciseconds))
 
-        detector_modules = {}
-        for source in self.detector_sources:
-            name, modno = DETECTOR_SOURCE_RE.match(source).groups((1, 2))
-            detector_modules[(name, modno)] = source
-
-        # A run should only have one detector, but if that changes, don't hide it
-        detector_name = ','.join(sorted(set(k[0] for k in detector_modules)))
-
         # disp
         print('# of trains:   ', train_count)
         print('Duration:      ', span_txt)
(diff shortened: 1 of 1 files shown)


extra_data/reader.py
```diff
@@ -1197,37 +1197,41 @@ def info(self, details_for_sources=()):
...
+
+            detector_modules = {}
+            for source in self.detector_sources:
+                name, modno = DETECTOR_SOURCE_RE.match(source).groups((1, 2))
+                detector_modules[(name, modno)] = source
+
+            # A run should only have one detector, but if that changes, don't hide it
+            detector_name = ','.join(sorted(set(k[0] for k in detector_modules)))
+
+            print("{} detector modules ({})".format(
```
takluyver, line 1219:
```suggestion
            print("{} XTDF detector modules ({})".format(
```

While we're looking at this code, let's take a moment to make the messages a bit more accurate.

extra_data/reader.py
```diff
@@ -1274,9 +1278,16 @@ def keys_detail(s, keys, prefix=''):
...
-        print(len(non_detector_inst_srcs), 'instrument sources (excluding detectors):')
-        for s in sorted(non_detector_inst_srcs):
+        if details_for_sources:
+            # All instrument sources with details enabled.
+            displayed_inst_srcs = self.instrument_sources
+            print(len(displayed_inst_srcs), 'instrument sources:')
+        else:
+            # Only non-XTDF instrument sources without details enabled.
+            displayed_inst_srcs = self.instrument_sources - self.detector_sources
+            print(len(displayed_inst_srcs), 'instrument sources (excluding detectors):')
```
takluyver, line 1288:
```suggestion
            print(len(displayed_inst_srcs), 'instrument sources (excluding XTDF detectors):')
```

philsmt:
Thanks, added the clarifications.

PR #449, opened 2023-10-05
Expose units symbol & name on KeyData objects

takluyver:
As two new properties, `.units` and `.units_name`. These combine the unit & metric prefix attributes written by Karabo to give symbols like 'μJ' or 'fs'. They will be `None` if these attributes aren't available. [Pint](https://pypi.org/project/Pint/) recognises either the symbol or the name (for the couple of cases I tried).

There's no caching at present. I assume people won't retrieve units many times for the same key, and the OS disk cache could help a bit anyway.

### DIFF
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -170,7 +170,7 @@ class KeyData:
             with h5py.File(filename, 'r') as f:
                 attrs = dict(f[self.hdf5_data_path].attrs)
 
-        return dict(dset.attrs)
+        return attrs
 
     def select_trains(self, trains):
         """Select a subset of trains in this data as a new :class:`KeyData` object.
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -109,6 +109,7 @@ class KeyData:
 
     @property
     def units(self):
+        """The units symbol for this data, e.g. 'μJ', or None if not found"""
         attrs = self.attributes()
         base_unit = attrs.get('unitSymbol', None)
         if base_unit is None:
@@ -121,6 +122,7 @@ class KeyData:
 
     @property
     def units_name(self):
+        """The units name for this data, e.g. 'microjoule', or None if not found"""
         attrs = self.attributes()
         base_unit = attrs.get('unitName', None)
         if base_unit is None:
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -107,6 +107,28 @@ class KeyData:
         return self.nbytes / 1e9
 
     @property
+    def units(self):
+        dset = self.files[0].file[self.hdf5_data_path]
+        base_unit = dset.attrs.get('unitSymbol', None)
+        if base_unit is None:
+            return None
+
+        prefix = dset.attrs.get('metricPrefixSymbol', '')
+        if prefix == 'u':
+            prefix = 'μ'  # We are not afraid of unicode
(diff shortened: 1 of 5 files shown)


extra_data/keydata.py
```diff
@@ -106,6 +106,28 @@ def size_gb(self):
         """The size of the data in memory in gigabytes."""
         return self.nbytes / 1e9
 
+    @property
+    def units(self):
+        dset = self.files[0].file[self.hdf5_data_path]
```
JamesWrigley, line 111: Will this work with a voview file?

extra_data/keydata.py
```diff
@@ -106,6 +106,28 @@ def size_gb(self):
...
+        base_unit = dset.attrs.get('unitSymbol', None)
+        if base_unit is None:
+            return None
+
+        prefix = dset.attrs.get('metricPrefixSymbol', '')
+        if prefix == 'u':
+            prefix = 'μ'  # We are not afraid of unicode
```
JamesWrigley, line 118: :muscle: :muscle: :muscle:

JamesWrigley (changes requested):

extra_data/keydata.py
takluyver: Excellent point; no, at present it won't. I can add that to the code for creating virtual overview files, of course, but then it will only available on newly created files. :thinking:

JamesWrigley: Reminds me of https://github.com/European-XFEL/DAMNIT/issues/93, except here we can't even assume that there's a `raw/` directory :sweat_smile: I think either having a method to follow the links in a voview file or adding the attributes to voview files is fine (I can regenerate all the existing files like I did before with the run metadata bug), but I lean towards having a method to follow the links. That seems a bit more future proof and would also be useful for that DAMNIT bug.

takluyver: OK, I've added an `.attributes()` method returning a dict of everything we get from the HDF5 dataset, and if that is empty and its a virtual dataset, it looks up one of the source files.

extra_data/keydata.py
```diff
@@ -151,6 +152,18 @@ def source_file_paths(self):
...
         return [Path(p) for p in paths]
 
+    def attributes(self):
+        dset = self.files[0].file[self.hdf5_data_path]
+        if len(dset.attrs) == 0 and dset.is_virtual:
+            # Virtual datasets were initially created without these attributes.
+            # Find a source file. Not using source_file_paths as it can give [].
+            _, filename, _, _ = dset.virtual_sources()[0]
+            # Not using FileAccess: no need for train or source lists.
+            f = h5py.File(filename, 'r')
```
JamesWrigley, line 162: `with` statement to be on the safe side?

JamesWrigley (changes requested):

JamesWrigley:
(oh, and this should probably be documented too)

JamesWrigley (approved):
LGTM!

takluyver:
Thanks!

prnote: New properties `units` and `units_name` on KeyData objects to retrieve units metadata written by Karabo.

philsmt:
LGTM!

Some time ago I scanned over several runs and found only `unitSymbol` and `metricPrefixSymbol` to be present in any reliable manner. If necessary, we can add a note to the documentation.

PR #458, opened 2023-10-24
karabo-bridge-serve-run command

### LINKED ISSUE #455: CLI to stream combined raw+proc data
An instrument scientist spotted that there's no option at present to stream a combination of raw+proc data, even though EXtra-data generally makes it easy to work with both together.

I'd also like to improve selecting sources & keys to stream, and make it harder to accidentally stream all keys from all sources, which tends to be slow. So maybe we can have a new command like `karabo-bridge-serve-run`.

takluyver:
This is like `karabo-bridge-serve-files` but with the following improvements:

- You specify proposal & run number rather than a run directory
- It combines raw & proc data for streaming (#455)
- You have to specify sources with `--include`, rather than sending everything by default, which is slow. If you really want everything, `--include '*'` should do it.
- You can select >1 pattern by passing `--include` multiple times.
- Trains with missing data are skipped by default, use `--allow-partial` to include them (find better name?)
- Some useful info is shown while it runs (also added to -serve-files)

```shell
#                   Proposal run
karabo-bridge-serve-run 4237 219 --port 41234 \
    --include 'FXE_XAD_JF1M/DET/JNGFR*:daqOutput[data.adc]' \
    --include 'FXE_AUXT_LIC/DOOCS/PPODL[*Position]'
```

You can select a subset of keys using `[]` syntax. For now, I've done it this way for both instrument and control data, whereas Metro only supports that syntax for instrument data. I'm still not sure what I prefer.

Closes #455

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -44,6 +44,75 @@ Check the structure of an EuXFEL run or HDF5 file:
 If it finds problems with the data, the program will produce a list of them and
 exit with status 1. See :doc:`validation` for details of what it checks.
 
+.. _cmd-serve-run:
+
+``karabo-bridge-serve-run``
+---------------------------
+
+Stream data from a run, by proposal & run number, in the `Karabo bridge
+<https://rtd.xfel.eu/docs/data-analysis-user-documentation/en/latest/online.html#streaming-from-karabo-bridge>`_
+format. See :doc:`streaming` for more information.
+
+.. code-block:: shell
+
+   #                     Proposal run
+   karabo-bridge-serve-run 700000 40 --port 4321 \
+        --include 'SPB_IRDA_JF4M/DET/JNGFR*:daqOutput' \
+        --include '*/MOTOR/*[*Position]'
+
+.. program:: karabo-bridge-serve-run
+
+.. option:: --port <port>
+
+   Either a numeric TCP port, e.g. ``4321``, or a ZMQ endpoint address, e.g.
+   ``tcp://0.0.0.0:4321``. You will need to give the receiving code this port
+   number or address as well.
+
+   If no port is specified, the program will pick an unused port, and display
+   the endpoint address as it starts.
+
+.. option:: --include <pattern>
+
+   Sources matching the pattern will be included in the streamed data.
+   You can also match keys by putting a key pattern in ``[]`` square brackets
+   at the end of the source pattern.
+
+   You must specify at least one ``--include`` pattern, and you can use the
+   option several times to expand the selection.
+
+   If data is flowing slower than you expect, see if you can use more specific
+   patterns to avoid sending unnecessary sources.
+
+.. option:: --allow-partial
+
+   By default, trains are only sent if they contain all the data selected by
+   ``--include``. This option also sends trains where some of that data is
+   missing.
+
+.. option:: --append-detector-modules
+
+   If the file data contains multiple detector modules as separate sources,
+   i. e. for big area detectors (AGIPD, LPD and DSSC), append these into one
+   single source.
+
+.. option:: --dummy-timestamps
+
+   Add mock timestamps if missing in the original meta-data.
+
+These two options above - appended module sources and dummy timestamps - are
+required if streamed data shall be provided to OnDA.
+
+.. option:: -z <type>, --socket-type <type>
+
+   The ZMQ socket type to use, one of ``PUB``, ``PUSH`` or ``REP``.
+   Default: ``REP``.
+
+.. option:: --use-infiniband
+
+   Use the infiniband network interface (``ib0``) if it's present.
+   This is ignored if ``--port`` is used with a full ZMQ endpoint address.
+
+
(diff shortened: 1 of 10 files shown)


takluyver:
streaming

extra_data/cli/serve_run.py
```diff
@@ -0,0 +1,81 @@
...
+        "-z", "--socket-type", help="ZeroMQ socket type",
+        choices=['PUB', 'PUSH', 'REP'], default='REP'
+    )
+    args = ap.parse_args(argv)
+
+    try:
+        from ..export import serve_data
+    except ImportError as e:
+        sys.exit(IMPORT_FAILED_MSG.format(e))
+
```
dallanto, line 52: What is the reason to do this import here within the function and not at the top of the file? (except of course the fail message constant needs to be defined first)

extra_data/export.py
```diff
@@ -189,7 +189,7 @@ def serve_data(data, port, append_detector_modules=False,
         count += 1
         new_time = time.monotonic()
         if count % 5 == 0:
-            rate = len(deque) / (new_time - sent_times[0])
+            rate = len(sent_times) / (new_time - sent_times[0])
```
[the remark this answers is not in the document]
takluyver: Yup, that was just a silly mistake. :wink:

extra_data/cli/serve_run.py
takluyver: I'd usually try to avoid side-effects (like `sys.exit()`) when loading a module, although it doesn't matter so much for a module defining a CLI like this. It's also handy that `--help` still works even without the extra dependencies.

We could still have the import at the top like this:

```python
# Top of file
try:
    from ..export import serve_data
except ImportError:
    serve_data = None

# In the function
if serve_data is None:
    sys.exit(msg)
```

But that looks less neat to me.

dallanto:
Hi Thomas, irrespective of my questions, LGTM.

takluyver:
Thanks Fabio!

I just tweaked the CLI design slightly. I thought the 3 numbers together (proposal, run, port) with no markers was somewhat confusing, so I've made `--port` an option:

```shell
karabo-bridge-serve-run 4237 219 --port 41234 ...
```

I think this makes the meaning clearer. It also means port can be unspecified, which defaults to picking a random unused port. The 'streamer started on' message gives you the actual port in use, so you can tell the consumer where to connect.

dallanto:
The optional port argument is a very useful change/addition. (LEBTM 😉 )

takluyver:
Testing this was something of a challenge; I've introduced an environment variable (`EXTRA_DATA_DATA_ROOT`) to override the `/gpfs/exfel/exp` root location in a child process. This is undocumented because it's only for testing, at least for now.

The child processes also still aren't counted for coverage, and I don't know why not. pytest-cov is meant to count subprocesses, and it does when I run the tests locally. :shrug:

tmichela:
seems it's bit tricker to measure coverage for spakned subprocesses: https://coverage.readthedocs.io/en/latest/subprocess.html#subprocess

takluyver:
[pytest-cov claims](https://pytest-cov.readthedocs.io/en/latest/readme.html) that it handles subprocesses automatically, and it does work on my local machine. I thought it was missing because we use SIGKILL to stop the child process, so I made the tests send SIGINT first and give it some time to clean itself up, but it still doesn't seem to count the coverage on CI. :shrug:

docs/cli.rst
```diff
@@ -53,6 +122,9 @@ Stream data from files in the `Karabo bridge
 <https://rtd.xfel.eu/docs/data-analysis-user-documentation/en/latest/online.html#streaming-from-karabo-bridge>`_
 format. See :doc:`streaming` for more information.
 
+For streaming data from a run directory, we recommend the newer
+:ref:`cmd-serve-run` command in place of this.
+
```
dallanto, line 127: In fact, I have used the old command (only) for streaming from a run directory, using the full path as argument. Apart from the fact that ``...-serve-run`` is indeed more convenient to achieve this, what would be the main use case for using the old command now?

takluyver: The `-serve-files` command makes it easy to stream from a non-standard run directory location, e.g. if we do an experimental correction of a run, we might put it in proposal `scratch`. Or it gives you a way to stream from `red` before we've integrated support for that. Or if users transfer run data back to their home institution and want to use EXtra-data there. I think the new `-serve-run` command will be better for ~95% of use cases.

The biggest reason to retain the old command is compatibility & familiarity, though - don't break what's working for people. :slightly_smiling_face:

takluyver:
I'm going to merge this on the grounds that I'd already got an LGTM on the interface & implementation, and Fabio has looked through the tests (we also discussed some more points about this on Zulip) without identifying any problems.

Thanks for the review!

takluyver:
prnote: New command `karabo-bridge-serve-run` to more conveniently stream data from a saved run in Karabo Bridge format.

PR #462, opened 2023-11-06
Try to improve coverage measurement

takluyver:
By telling coverage that installed paths are equivalent to source paths ([docs](https://coverage.readthedocs.io/en/7.3.2/config.html#paths)).


takluyver:
That seems to have caught some additional code in `validation.py`, where we're using multiprocessing, but not the CLI pieces I was expecting to pick up.

Issue #463, opened 2023-11-14
`components.JUNGFRAU`'s `_source_re` doesn't include all possible jungfrau strings


kakhahmed:
In pycalibration `components.JUNGFRAU` is used to read data to plot in the offline correction report.

With the new Jungfrau detectors `DET_LAB_JF` and `SCS_HRIXS_JUNGF` this module fails to find any valid data to read because it doesn't expect to parse such detector names.

![Screenshot from 2023-11-14 09-28-49](https://github.com/European-XFEL/EXtra-data/assets/36923212/be498be2-6e5d-4521-8dab-e397e1295a2f)

At the moment (since deploying [3.11.4](https://git.xfel.eu/calibration/pycalibration/-/tags/3.11.4)), I have hotfixed the used extra-data library to include `JF` and `JUNGF`. We can add these to extra-data as a fix as well. But this isn't enough. I think we should have `_source_re` check for a new input argument (e.g. source_re) that can have the expected detector name to look for. This would avoid similar issues in the future and save us from the essentiality of hardcoding these strings.

https://github.com/European-XFEL/EXtra-data/blob/96a2b77d0cd59fb9f23a0abfe9fcdc8beb897277/extra_data/components.py#L1573-L1575

takluyver:
> I think we should have _source_re check for a new input argument (e.g. source_re) that can have the expected detector name to look for.

We already have a `detector_name` argument. Maybe if we split the source regex into two parts, we can use the detector name pattern only to find the name if one isn't already specified. E.g. something like this:

```python
_detector_name_re = re.compile(r'(?P<detname>.+_(JNGFR|JF[14]M|JF500K\d?))')
_source_suffix_re = re.compile(r'/DET/(MODULE_|RECEIVER-|JNGFR)(?P<modno>\d+)')
```

It does still admittedly rely on a regular pattern for the module numbers. Ugh, I'm starting to hate module numbers.

kakhahmed:
> > I think we should have _source_re check for a new input argument (e.g. source_re) that can have the expected detector name to look for.
> 
> We already have a `detector_name` argument. Maybe if we split the source regex into two parts, we can use the detector name pattern only to find the name if one isn't already specified. E.g. something like this:
> 
> ```python
> _detector_name_re = re.compile(r'(?P<detname>.+_(JNGFR|JF[14]M|JF500K\d?))')
[...]

PR #464, opened 2023-11-15
Expand JUNGFRAU components to accept SCS_HRIXS_JUNGF and DET_LAB_JF

kakhahmed:
This addresses the mentioned bug in this issue: https://github.com/European-XFEL/EXtra-data/issues/463

This is the first and fast step to fix the issue faced by pycalibration when correcting data for SCS_HRIXS_JUNGF and DET_LAB_JF.

The error at Jungfrau correction notebook
```
SourceNameError: This data has no source named 'DET_LAB_JF/DET/...'.
See data.all_sources for available sources.
```

I will try to propose a solution to avoid such issues later with new detector names.


takluyver:
Thanks, LGTM

PR #467, opened 2023-11-24
Clone euxfel data file structure

tmichela:
Add a little tool to generate EuXFEL data files by copying existing run data, but without any of the original data in the resulting files.

I made this originaly to build test for the [exdf-tools](https://git.xfel.eu/dataAnalysis/exdf-tools) as this is critical to test on real-like data from various cycles, but I thought this would be useful more generally, hence it ends up here.


extra_data/tests/utils/gendata.py
```diff
@@ -0,0 +1,127 @@
...
+    print(pbar, file=sys.stderr)
+
+
+def _clone_file_structure(
+    h5file: Path, output: Path, *, run_data=False, control_data=False
+) -> None:
+    clone = h5py.File(output / h5file.name, "w")
+
+    def visitor(name, obj):
+        if isinstance(obj, h5py.Group):
```
takluyver, line 30: We're planning to start using HDF5 links soon to (temporarily) have two different names for the corrected data sources, while we transition to new names: https://git.xfel.eu/calibration/planning/-/issues/170

It would be useful if this tool could preserve that (the current code won't). Checking for soft links is fairly easy:

```python
link = original.get(name, getlink=True)
if isinstance(link, h5py.SoftLink):
    ...
```

Checking for hard links is trickier - normally everything is a hard link, so you're really checking for two links pointing to the same object. But I think we're more likely to go with soft links, so we can probably overlook that for now.

takluyver (commented):
One more comment: naming!

I'd move this out of tests, since it's useful for things beyond EXtra-data's own tests.

`gendata` doesn't really describe what it does - it's not generating anything, and the output doesn't have most of the data. :wink:  What about something like these options:

- `extra_data.mkskeleton`
- `extra_data.mkemptyfiles`
- `extra_data.hollowfiles`
- `extra_data.stripfiles`

extra_data/tests/utils/gendata.py
```diff
@@ -0,0 +1,127 @@
...
+            clone.create_group(name)
```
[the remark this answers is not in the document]
tmichela: good point, but now that we want to copy attributes as well, we'll need this branch.

extra_data/tests/utils/gendata.py
tmichela: ah, thanks for the info. I've added support for soft links.

tmichela:
> I'd move this out of tests

Fine with that, I though it is niche enough to hide it there, but 🤷 

> One more comment: naming!

yea, I gave it purposely a terrible name to get ideas back 😁  
I don't have any particular preference, so if you have a favourite?

Maybe also to consider:
`clone_empty`
`blanc_copy`
`empty_replica`

takluyver:
Or `copy_structure`, to give yet another option. I think that might be my favourite so far, but I don't have a strong preference.

extra_data/tests/utils/gendata.py
tmichela: While testing copying the softlinks, I realised `visititems` do not visit softlinks (already reported to h5py [issues](https://github.com/h5py/h5py/issues/671)) so I'd have to make my own visitor to handle that case... or ignore soft links for now.

takluyver: Aha, good point. I think we can get around that by using the somewhat hidden [`group.id.links.visit()`](http://api.h5py.org/h5l.html#h5py.h5l.LinkProxy.visit) method, which visits the links below a group, rather than the objects which the links point to. But that's extra complexity, so let's leave that until we actually start making files that use it.

takluyver:
`extra_data.copy_structure` is still the name I like most so far, but I'd be OK with any of our suggestions if you're developing a preference.

Other than that, LGTM

tmichela:
So I added support for soft/hard links anyway, if you want to have an other look at it.
(+ some bonus unit tests)

extra_data/copy.py
```diff
@@ -0,0 +1,182 @@
...
+        if obj.name != "/":
+            link = obj.file.get(obj.name, getlink=True)
+            if isinstance(link, h5py.SoftLink):
+                output[obj.name] = h5py.SoftLink(link.path)
+                return
+            elif isinstance(link, h5py.ExternalLink):
+                # TODO do we want to support external links?
+                # this *might* work, but external softlinks may point to non reacheable data
+                # with h5py.File(link.filename) as ext:
+                #     Cloner(ext[link.path], output[obj.name], run_data=self.run_data, control_data=self.control_data)
```
takluyver, line 50: Since we've got the code here ready, why not just copy the link itself like we're doing for soft links? I think just doing `output[obj.name] = link` should do it.

extra_data/copy.py
```diff
@@ -0,0 +1,182 @@
...
+        self.visit(input, output)
+
+    @staticmethod
+    def _copy_attrs(input, output):
+        for key, value in input.attrs.items():
+            output.attrs.create(key, value)
+
+    def visit(self, obj, output):
+        if obj.name != "/":
+            link = obj.file.get(obj.name, getlink=True)
```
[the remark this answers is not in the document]
takluyver: (It seems it does work for soft links - i.e. if you get an object via a soft link, the `.name` is the name of the link, not the name it points to. I was somewhat surprised by that. :slightly_smiling_face: )

extra_data/copy.py
tmichela: good point, although, the idea is that these generated files might be used somewhere with no access the external files it links to (particularly test envs). But sure I'll add support to it.

takluyver:
Thanks, LGTM :+1:

extra_data/copy.py
takluyver: If there's a voview file inside the run directory, it will contain links to other files in the same directory. Those should be done as relative paths, so the copy of that file should point to the other files in the same directory as itself. :crossed_fingers:

tmichela:
thanks for the review! :)

takluyver:
prnote: Add a development tool to copy the structure of EuXFEL data files without the data.

PR #469, opened 2023-12-01
Fix some deprecation warnings from tests

takluyver:
- `np.product()` is deprecated in favour of `np.prod()`
- `datetime.utcnow()` is deprecated in Python 3.12, because it gives a UTC timestamp without timezone information
- There was a warning from numpy because the JUNGFRAU mock data had uint16 masks when they should be uint32.


tmichela:
LGTM

takluyver:
prnote: Fixed a selection of deprecation warnings.

PR #472, opened 2023-12-13
Validation: clearer info on filesystem errors

takluyver:
If opening an HDF5 file fails, see if we can read from it at all. If not, present the OS error in place of the HDF5 error, as HDF5's error messages obscure what's really going wrong.

```
$ python -m extra_data.validation /gpfs/exfel/exp/SPB/202302/p004511/raw/r0011/RAW-R0011-AGIPD02-S00010.h5
Checking file: /gpfs/exfel/exp/SPB/202302/p004511/raw/r0011/RAW-R0011-AGIPD02-S00010.h5

Could not access file
  error: [Errno 5] Input/output error
  file: /gpfs/exfel/exp/SPB/202302/p004511/raw/r0011/RAW-R0011-AGIPD02-S00010.h5
```

### DIFF
--- a/extra_data/tests/test_validation.py
+++ b/extra_data/tests/test_validation.py
@@ -6,7 +6,7 @@ import numpy as np
 from pytest import fixture, raises
 from tempfile import TemporaryDirectory
 
-from extra_data.validation import FileAccess, FileValidator, RunValidator, ValidationError
+from extra_data.validation import FileAccess, FileValidator, RunValidator, ValidationError, main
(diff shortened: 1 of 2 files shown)


philsmt:
Thanks, LGTM

Nested `try-except` are ugly, but I cannot think of a better way.

takluyver:
Thanks for the review!

takluyver:
prnote: `extra-data-validate` gives clearer messages for filesystem errors

PR #473, opened 2023-12-15
Draft: Introduce REDUCTION and ERRATA file sections

steffenhauf:
@philsmt - a draft that introduces the additional sections in the H5 structure we had discussed yesterday. We can shape and rename this as needed. For now this brach is to get the CI on the DAQ going again with files that contain additional sections.

### DIFF
--- a/extra_data/file_access.py
+++ b/extra_data/file_access.py
@@ -146,6 +146,8 @@ class FileAccess(metaclass=MetaFileAccess):
             self.train_ids = _cache_info['train_ids']
             self.control_sources = _cache_info['control_sources']
             self.instrument_sources = _cache_info['instrument_sources']
+            self.reduction_data = _cache_info['reduction_data']
+            self.errata = _cache_info['errata']
             self.validity_flag = _cache_info.get('flag', None)
         else:
             try:
@@ -155,7 +157,8 @@ class FileAccess(metaclass=MetaFileAccess):
 
             self.train_ids = tid_data[tid_data != 0]
 
-            self.control_sources, self.instrument_sources = self._read_data_sources()
+            (self.control_sources, self.instrument_sources,
+                self.reduction_data, self.errata) = self._read_data_sources()
 
             self.validity_flag = None
 
@@ -295,6 +298,7 @@ class FileAccess(metaclass=MetaFileAccess):
 
     def _read_data_sources(self):
         control_sources, instrument_sources = set(), set()
+        reduction_info, errata = set(), set()
 
         # The list of data sources moved in file format 1.0
         if self.format_version == '0.5':
@@ -320,6 +324,10 @@ class FileAccess(metaclass=MetaFileAccess):
                 # TODO: Do something with groups?
             elif category == 'CONTROL':
                 control_sources.add(h5_source)
+            elif category == 'REDUCTION':
+                reduction_info.add(h5_source)
+            elif category == 'ERRATA':
+                errata.add(h5_source)
             elif category == 'Karabo_TimerServer':
                 # Ignore virtual data source used only in file format
                 # version 1.1 / pclayer-1.10.3-2.10.5.
@@ -327,7 +335,8 @@ class FileAccess(metaclass=MetaFileAccess):
             else:
                 raise ValueError("Unknown data category %r" % category)
 
-        return frozenset(control_sources), frozenset(instrument_sources)
+        return (frozenset(control_sources), frozenset(instrument_sources),
+                frozenset(reduction_info), frozenset(errata))
 
     def _guess_valid_trains(self):
         # File format version 1.0 includes a flag which is 0 if a train ID
@@ -418,6 +427,10 @@ class FileAccess(metaclass=MetaFileAccess):
             return {''}
         elif source in self.instrument_sources:
             return set(self.file[f'/INDEX/{source}'].keys())
+        elif source in self.reduction_data:
+            return set(self.file[f'/INDEX/{source}'].keys())
+        elif source in self.errata:
+            return set(self.file[f'/INDEX/{source}'].keys())
         else:
             raise SourceNameError(source)
 
@@ -454,6 +467,10 @@ class FileAccess(metaclass=MetaFileAccess):
             group = '/CONTROL/' + source
         elif source in self.instrument_sources:
             group = '/INSTRUMENT/' + source
+        elif source in self.reduction_data:
+            group = '/REDUCTION/' + source
+        elif source in self.errata:
+            group = '/ERRATA/' + source
         else:
             raise SourceNameError(source)
 
@@ -478,6 +495,10 @@ class FileAccess(metaclass=MetaFileAccess):
             group = '/CONTROL/' + source
         elif source in self.instrument_sources:
             group = '/INSTRUMENT/' + source
+        elif source in self.reduction_data:
+            group = '/REDUCTION/' + source
+        elif source in self.errata:
+            group = '/ERRATA/' + source
         else:
             raise SourceNameError(source)
 
@@ -527,6 +548,10 @@ class FileAccess(metaclass=MetaFileAccess):
             path = '/CONTROL/{}/{}'.format(source, key.replace('.', '/'))
         elif source in self.instrument_sources:
             path = '/INSTRUMENT/{}/{}'.format(source, key.replace('.', '/'))
+        elif source in self.reduction_data:
+            path = '/REDUCTION/{}/{}'.format(source, key.replace('.', '/'))
+        elif source in self.errata:
+            path = '/ERRATA/{}/{}'.format(source, key.replace('.', '/'))
         else:
             raise SourceNameError(source)
 
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -101,6 +101,10 @@ class DataCollection:
                     files_by_sources[source, 'CONTROL'].append(f)
                 for source in f.instrument_sources:
                     files_by_sources[source, 'INSTRUMENT'].append(f)
+                for source in f.reduction_data:
+                    files_by_sources[source, 'REDUCTION'].append(f)
+                for source in f.errata:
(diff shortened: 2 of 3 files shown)


philsmt:
Some context from recent floor discussions:
- As with `exdf-tools`, we were aiming to also have the DAQ introduce metadata about any (online) data reduction performed. The former tool has so far did this with custom datasets in `METADATA/reduction` and `INDEX`, if applicable.
- Looking into this from a DAQ perspective, Steffen suggested it would be easiest for their side to have a new root category and treat any such entries as fast data under a (semi-)virtual source name.
- It looks to me we should also be able to integrate the work done so far on the `exdf-tools` side under this umbrella.
- This should also simplify integration with EXtra-data, as instead of explicit support by properties and methods it leverages all the existing read machinery just under different names
- For compatibility and to avoid confusion, `DataCollection.all_sources` and other functionality like item access should probably continue to only include `CONTROL` and `INSTRUMENT` sources
- We could go for a similar solution as with aliases and use a custom indexer attribute for access, i.e. `run.reduction[<source>]`
- The same applies to `ERRATA`, which will be used to record anything that went wrong with data saving, e.g. a dropped train.

philsmt:
@takluyver @tmichela

extra_data/file_access.py
```diff
@@ -146,6 +146,8 @@ def __init__(self, filename, _cache_info=None):
             self.train_ids = _cache_info['train_ids']
             self.control_sources = _cache_info['control_sources']
             self.instrument_sources = _cache_info['instrument_sources']
+            self.reduction_data = _cache_info['reduction_data']
```
philsmt, line 149: This should either not be cached at all or use conditional access like `validity_flag` further below. Since it's only auxiliary information and may often be comparably small, it may be fine to always generate it on demand. If this changes in the future, we can add machinery to augment existing indices, or simply re-create them.

extra_data/file_access.py
```diff
@@ -418,6 +427,10 @@ def index_groups(self, source):
             return {''}
         elif source in self.instrument_sources:
             return set(self.file[f'/INDEX/{source}'].keys())
+        elif source in self.reduction_data:
+            return set(self.file[f'/INDEX/{source}'].keys())
+        elif source in self.errata:
```
philsmt, line 432: I understand from our example that `REDUCTION` sources are always prepended by some category (e.g. `REDUCTION/PULSE_SLICING/<XTDF-source>/...`). How would `ERRATA` sources look like? All the reader machinery assumes that source names are unique across all root section.

steffenhauf: There's nothing concrete implemented in the DAQ, but some form of source identification is required. Think of a train that came outside the  buffer range, and couldn't be store. You'd want to know what the data was, so it would be prefixed by the source in the path.

Also technically all this additionally data is FAST data for the DAQ - i.e. it will be stored only if a train contains it.

philsmt: Certainly, my question was indeed more concrete in how you would pick the source name. Say `SA3_XTD10_XGM/XGM/DOOCS:output` has a hiccup and sends a late train, under what _source name_ (i.e. the patch below `ERRATA`) would you save this? It **must** be different from `SA3_XTD10_XGM/XGM/DOOCS:output` by at least a prefix or suffix.

Furthermore I had a look into your example file for data reduction. There are `INDEX` entries for
* PCLAYER_CI/DAQ/DET_DATA_TEST_1:xtdf
* PCLAYER_CI/LFF/DATA_TEST_1:reductionOutput
* PULSE_REDUCTION/PCLAYER_CI/DAQ/DET_DATA_TEST_1:xtdf

with a correspondig entry in `INSTRUMENT` for the first and in `REDUCTION` for the latter two. Is `PCLAYER_CI/LFF/DATA_TEST_1` here a real device used as input for the data aggregator, or also some virtual source used for metadata tracking? Ideally _sources_ in `REDUCTION` or `ERRATA` would **never** be identical to actual sources potentially occuring in `CONTROL` or `INSTRUMENT`.

extra_data/file_access.py
```diff
@@ -320,14 +324,19 @@ def _read_data_sources(self):
...
+            elif category == 'REDUCTION':
+                reduction_info.add(h5_source)
+            elif category == 'ERRATA':
+                errata.add(h5_source)
             elif category == 'Karabo_TimerServer':
                 # Ignore virtual data source used only in file format
                 # version 1.1 / pclayer-1.10.3-2.10.5.
                 pass
             else:
                 raise ValueError("Unknown data category %r" % category)
```
philsmt, line 336: So this bit is probably the big sore point for this proposal - pretty much *all* prior versions of EXtra-data (or even `karabo_data` for that matter) will break with these new files.

Since this only affects the `METADATA/dataSources` entry but not the actual existance of the root groups, one way around this would be not list these sources here and either:
* Not list them at all and instead discover them by iterating over the root group
* List them in a distinct dataset only for these virtual metadata sources
* List both the *real* and *virtual* sources in a new dataset while preserving the old one for compatibility

steffenhauf: Yes, I don't see any way around it though, except if we add below METADATA after all. Generally one should possibly make this less restrictive in my view, as it could hit us again in the future.

extra_data/file_access.py
steffenhauf: So in the reduction case, I prefix all source names with `PULSE_REDUCTION` to make them distinct from the "real" sources, one could do the same for `ERRATA`, e.g. `ERRATA/LATE_TRAINS/source/name/and/data/path/`. 

Hence, for what you observe in the index section:

* PCLAYER_CI/DAQ/DET_DATA_TEST_1:xtdf is the real source
* PCLAYER_CI/LFF/DATA_TEST_1:reductionOutput is the real source
* PULSE_REDUCTION/PCLAYER_CI/DAQ/DET_DATA_TEST_1:xtdf is injected by the pulse dim reducer, i.e. is an artificial source.

takluyver:
Do you have some examples of the structure of what goes in the new sections?

To be honest, this *doesn't* look like a simple solution from my perspective. We need significant changes to EXtra-data just to let it open files with these additions (I don't think the changes here are actually sufficient), we're stuffing an extra layer of info into "source names", and we likely want to introduce some category of hidden sources, since these aren't really like the experimental data. :confused: 

Putting it into metadata made more sense to me, but I can envisage other ways to do it. Could I have a look at some examples and come back with ideas?

extra_data/file_access.py
steffenhauf: Makes sense. I'll remove from caching.

steffenhauf:
@takluyver - there's sample files on Maxwell: /gpfs/exfel/data/scratch/haufs/daq_online_reduction/

steffenhauf:
Btw, putting it in METADATA is fine for me, I would then nevertheless like introduce a substructure as a second level underneath. However, having a convenient way to read the info via extradata is beneficial in my view. Technically, it's like fast data in how the DAQ stores it, so in my view much of the existing machinery there could be reused.

takluyver:
**Main point**: We'd really like these additions to happen in a way that doesn't break old versions of EXtra-data (we know people use old versions and will complain to us), and we think that should be easy with a small adjustment to the structure. We can add the new top-level sections, but rather than adding them to `METADATA/dataSources`, create a new subgroup like `METADATA/reductionSources` to list them in. Existing versions of EXtra-data should just ignore this, and that gives us more time to work out how to expose them.

I appreciate you're trying to get this in very soon, but if it is possible to change this, we'd really appreciate it.

-------

**Secondary points**

1. Having two kinds of sources in REDUCTION - some corresponding to Karabo pipelines, and others which are to be understood as `{reduction_technique}/{source_name}` seems confusing to me. Would it make sense to put the 'real' Karabo sources in INSTRUMENT/CONTROL, so we can assume that everything under REDUCTION fits the same 'reduction technique' structure?
2. If there are multiple reductions for the same source, do we define the order in which they were applied? Probably this can just be the order that they appear in `reductionSources`.

philsmt:
Minor correction after discussing it with Steffen: Let's call the new source listing `METADATA/auxiliarySources` to also account for the `ERRRATA` section.

steffenhauf:
Just also had an offline discussion with @philsmt - I took from this that

- I'll keep the actual data under the high level keys REDUCTION and ERRATA
- The sources will not be added to METADATA/dataSources but rather METADATA/auxiliarySources/reduction and    METADATA/auxiliarySources/errata


Extra data could then provide access to these with a to be developed interface, which closely resembles iterating through INSTRUMENT data like sources.

Would this be within the points you've outline above @takluyver?

takluyver:
Yep, I think that works. I've just copied your example file and modified it (in `/gpfs/exfel/data/scratch/kluyvert/daq_online_reduction`) to do a quick mock-up of this idea, and confirmed that EXtra-data can open the file and read data from it, ignoring the additions.

steffenhauf:
Thanks. I've added a `v2` file to `/gpfs/exfel/data/scratch/haufs/daq_online_reduction` as an example of what the DAQ would output after refactoring to this new schema.

takluyver:
It looks like the `INDEX` sections for the instrument source in that v2 file are all zero-ed out, so EXtra-data sees the instrument data as empty (length 0 on first dimension). Is that a consequence of these changes, or just an artifact of how it was tested?

steffenhauf:
I think it's a bug in my changes. I'll check.

takluyver:
Thanks!

I might close this PR now that we have a design that doesn't require changes in EXtra-data. We'll still want to provide a nice API to get at the reduction information (& errata - I need to better understand what goes there), but it's somewhat lower priority now. We probably won't add the reduction sources to the JSON cache file. But on the other hand we probably should expose them through virtual overview files, as otherwise they'll be somewhat hidden for runs with a virtual overview file.

takluyver:
``

takluyver:
I've opened #477 to track creating APIs for these new sections.

PR #478, opened 2024-01-15
Use logging to show default aliases message instead of print()

### LINKED ISSUE #476: Add option to easily suppress stderr printouts when loading aliases in `open_run`
When opening runs with aliases files, this produces a print to stderr that informs about the loaded aliases (see `extra_data.reader.open_run`). 

e.g. `Loading 278 aliases from: /gpfs/exfel/exp/[...]/usr/extra-data-aliases.yml`

For me, this message turns out to be very annoying particularly when opening many runs. It would be great if this message could be suppressed easily. Therefore, I think one could make it a warning (so people can use warnings.filterwarnings) or have a flag to suppress this.

In the following, a possible work around that can be used right now. However, I think it should not be that complicated to suppress this message:
```python
import sys
import contextlib
import extra_data as xd

# based on reply in https://stackoverflow.com/questions/34904946/how-to-filter-stdout-in-python-logging
class StdFilter(object):
    def __init__(self, strings_to_filter, stream):
        self.stream = stream
        self.strings_to_filter = strings_to_filter
        self.triggered = False

    def __getattr__(self, attr_name):
        return getattr(self.stream, attr_name)

    def write(self, data):
        for string in self.strings_to_filter:
            if string in data:
                self.triggered = True
                self.stream.flush()
                return None
            
        if self.triggered:
            self.triggered = False
            if data=="\n":
                self.stream.flush()
                return None
            
        self.stream.write(data)
        self.stream.flush()

    def flush(self):
        self.stream.flush()
        
stderr_filter = StdFilter([" aliases from: "], sys.stderr)
        
with contextlib.redirect_stderr(stderr_filter):
    r = xd.open_run(4098, 199)

```

takluyver:
Closes #476

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1962,6 +1962,6 @@ def open_run(
 
     if aliases is not None:
         dc = dc.with_aliases(aliases)
-        print(f"Loading {len(dc._aliases)} aliases from: {aliases}", flush=True, file=sys.stderr)
+        log.info("Loading %d aliases from: %s", len(dc._aliases), aliases)
 
     return dc


JamesWrigley (approved):
LGTM!

takluyver:
prnote: Hide the message about proposal aliases when opening a run.

PR #479, opened 2024-01-16
Update RTD Sphinx search-as-you-type plugin

takluyver:
RTD sent me an email about a [security issue](https://github.com/readthedocs/readthedocs-sphinx-search/security/advisories/GHSA-xgfm-fjx6-62mj) with the Sphinx search-as-you-type plugin. I don't think it's a big deal in our static docs pages, but may as well update.

I've also added it to the plugins in config, which will hopefully mean we actually get search-as-you-type. :confused:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -34,6 +34,7 @@ extensions = [
     'sphinx.ext.napoleon',
     'nbsphinx',
     'sphinxcontrib_github_alt',
+    'sphinx_search.extension',
     # Workaround for sphinx-rtd-theme issue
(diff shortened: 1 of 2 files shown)


takluyver:
Search as you type pops up, but always seems to say no results found, even though I'm searching for things I know are there. Maybe it doesn't work correctly in pull requests? I'm not convinced, but I'm going to merge it, let the docs build, and see if it works in master. If it still doesn't work, I'll pull it out again.

takluyver:
Search as you type is working in the docs now.

PR #489, opened 2024-02-23
Catch error calculating run duration from corrupted timestamp

takluyver:
I'm hoping this is rare enough that just not crashing is sufficient. Here's what it displays for the run we found this on (p900438 r1):

```
# of trains:    358
Duration:       OverflowError
First train ID: 1936707599
Last train ID:  9223372040728191546
```


philsmt:
LGTM this fix.

If we're concerned about people being confused, we could add some text that the `OverflowError` is likely a result of bogus train IDs.

takluyver:
Good idea, I've added a note to clarify that.

takluyver:
prnote: Fixed OverflowError in lsxfel & run.info() with some corrupted train IDs

PR #490, opened 2024-02-26
Fix alias loading for old proposals

JamesWrigley:
Old proposals would copy all data from raw into proc so they would always contain the same sources, so when open_run() picked the proc DataCollection to return it wouldn't include any aliases because those are only loaded once for raw data.

Would it be possible to make a patch release for this? Unfortunately without it one can't use aliases in DAMNIT context files for old proposals :frowning_face:


takluyver:
LGTM, thanks!

I'll probably just release master as 1.16 - it looks like all low-risk fixes anyway, and this seems simpler than splitting out this fix as 1.15.2.

Issue #495, opened 2024-03-06
Combine attributes on CONTROL key's parent group with .value attributes


philsmt:
`CONTROL` keys have attributes both on their actual datasets `.value` and `.timestamp`, but also on the parent group of both. On closer inspection, the parent key actually carries most of the attributes, with only a few either copied (like `unitName`) or static at a wrong value (like `daqPolicy`). Most importantly though, attributes like `description`, `displayedName` and others are missing.

We should merge the attributes from the parent group when using `KeyData.attributes()` on `{key}.value`.

PR #498, opened 2024-03-07
Merge attributes of key group and value dataset for CONTROL keys

philsmt:
Unfortunately I noticed that the HDF attributes mapping to Karabo properties at the time of recording are a bit all over the place in the case of `CONTROL` keys. For example, `SA3_XTD10_XGM/XGM/DOOCS.beamposition.ixPos` for a recent run:

```shell
├21 attributes:
│ ├accessMode: [2]
│ ├alias: 'IX.POS'
│ ├assignment: [0]
│ ├daqPolicy: [-1]
│ ├defaultValue: [-1.]
│ ├description: 'Calculated X position [mm]'
│ ├displayedName: 'X'
│ ├frac: [0]
│ ├leafType: [0]
│ ├metricPrefixEnum: [13]
│ ├metricPrefixName: 'milli'
│ ├metricPrefixSymbol: 'm'
│ ├nodeType: [0]
│ ├requiredAccessLevel: [0]
│ ├sec: [0]
│ ├tags: ['doocs' 'poll']
│ ├tid: [0]
│ ├unitEnum: [2]
│ ├unitName: 'meter'
│ ├unitSymbol: 'm'
│ └valueType: 'FLOAT'
├timestamp	[uint64: 500]
│ └4 attributes:
│   ├daqPolicy: [1]
│   ├frac: [0]
│   ├sec: [0]
│   └tid: [0]
└value	[float32: 500]
  └12 attributes:
    ├alias: 'IX.POS'
    ├daqPolicy: [1]
    ├frac: [0]
    ├metricPrefixEnum: [13]
    ├metricPrefixName: 'milli'
    ├metricPrefixSymbol: 'm'
    ├sec: [0]
    ├tags: ['doocs' 'poll']
    ├tid: [0]
    ├unitEnum: [2]
    ├unitName: 'meter'
    └unitSymbol: 'm'
``` 

After scanning through attributes across a large number of sources, I found:
* In all cases except one (`minSize`, not shown here), the keys on `value` are a true subset of the parent
* Duplicate attributes in `value` either have the same value (e.g. `alias` above) or a wrong, likely constant value (e.g. `daqPolicy`)
* `timestamp` always has exactly the four attributes with exactly those values as shown above

Based on that, this PR merges the attributes on the parent group into those returned from `KeyData.attributes()` for the `value` dataset of a key, specifically overwriting any existing key.

### DIFF
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -73,6 +73,16 @@ class KeyData:
                f"for {len(self.train_ids)} trains>"
 
     @property
+    def is_control(self):
+        """Whether this key belongs to a control source."""
+        return self.section == 'CONTROL'
+
+    @property
+    def is_instrument(self):
+        """Whether this key belongs to an instrument source."""
+        return self.section == 'INSTRUMENT'
+
+    @property
     def index_group(self):
         """The part of the key needed to look up index data"""
         if self.section == 'INSTRUMENT':
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -165,6 +165,19 @@ class KeyData:
         from pathlib import Path
         return [Path(p) for p in paths]
 
+    def _find_attributes(self, dset):
+        """Find Karabo attributes belonging to a dataset."""
+        attrs = dict(dset.attrs)
+
+        if self.is_control and self.key.endswith('.value'):
+            # For CONTROL sources, most of the attributes are saved on
+            # the parent group rather than the .value dataset. In the
+            # case of duplicated keys, the parent value appears to be
+            # the correct one.
+            attrs.update(dict(dset.parent.attrs))
+
+        return attrs
+
     def attributes(self):
         """Get a dict of all attributes stored with this data
 
@@ -172,14 +185,14 @@ class KeyData:
         convenient forms.
         """
         dset = self.files[0].file[self.hdf5_data_path]
-        attrs = dict(dset.attrs)
+        attrs = self._find_attributes(dset)
         if (not attrs) and dset.is_virtual:
             # Virtual datasets were initially created without these attributes.
             # Find a source file. Not using source_file_paths as it can give [].
             _, filename, _, _ = dset.virtual_sources()[0]
             # Not using FileAccess: no need for train or source lists.
             with h5py.File(filename, 'r') as f:
-                attrs = dict(f[self.hdf5_data_path].attrs)
+                attrs = self._find_attributes(f[self.hdf5_data_path])
 
         return attrs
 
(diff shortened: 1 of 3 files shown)


philsmt:
Fixes #495

extra_data/keydata.py
```diff
@@ -155,21 +165,32 @@ def source_file_paths(self):
...
         return [Path(p) for p in paths]
 
+    def _find_attributes(self, dset):
+        """Find Karabo attributes belonging to a dataset."""
+        attrs = dict(dset.attrs)
+
+        if self.is_control and self.key.endswith('.value'):
+            # For CONTROL sources, most of the attributes are saved on
+            # the parent group rather than the .value dataset.
+            attrs.update(dict(dset.parent.attrs))
```
takluyver, line 175: Nice, I either didn't know or had forgotten that datasets had a `.parent` attribute. :laughing:

extra_data/keydata.py
```diff
@@ -155,21 +165,32 @@ def source_file_paths(self):
         from pathlib import Path
         return [Path(p) for p in paths]
 
+    def _find_attributes(self, dset):
+        """Find Karabo attributes belonging to a dataset."""
+        attrs = dict(dset.attrs)
+
+        if self.is_control and self.key.endswith('.value'):
+            # For CONTROL sources, most of the attributes are saved on
+            # the parent group rather than the .value dataset.
```
takluyver, line 174: Let's also mention here that at least currently, some overlapping attributes differ, and in that case the group level attribute seems to be the one that's right.

(I have a nasty foreboding that some day we'll come across a case where the group attribute is wrong and the dataset attribute is right, and we'll have to figure out how to wrangle that. But hopefully that's just pessimism. :crossed_fingers: )

takluyver:
Other than that, LGTM. :+1:

extra_data/keydata.py
tmichela: > (I have a nasty foreboding that some day we'll come across a case where the group attribute is wrong and the dataset attribute is right, and we'll have to figure out how to wrangle that. But hopefully that's just pessimism. 🤞 )

I don't think that's a possibility.
The attributes come from the device schema. The HDF5 datasets `.value` and `.timestamp` don't exist on the device. The parent group of those datasets, is actually (on the device) the property leaf that holds the value and attributes on the device we're saving.

So the attributes on these dataset must be copied from the parent group (or have some default). I guess the wrong default could as well be on the parent group, but I don't see how this could be wrong on the parent group and correct on the dataset.

philsmt: Good point, added that clarification.

In general though, I wouldn't worry much. I'll get in touch with the DAQ team and asked them to verify our assumptions here.

We may even be able to remove the dataset attributes entirely again and shift EXtra-data to the ones on the group. Frankly I doubt users would notice.

PR #499, opened 2024-03-07
Support selecting down DataCollection by SourceData objects

philsmt:
Seems like an obvious addition after `SourceData` became a public interface 🤷 

Also adds missing tests for using `KeyData` in this place.


takluyver:
Thanks, LGTM!

philsmt:
Thanks for review!

PR #502, opened 2024-03-20
Return a copy of the array for detector_key.train_id_coordinates()

takluyver:
So that modifying the returned array doesn't affect the internals of the detector data objects.

This also means the `.train_id_coordinates()` method will consistently return a NumPy array - previously for JUNGFRAU it would have been a list. That's technically an API break, but I think the consistency is worth it.

cc @turkot - I spotted this while looking at the code to answer your question.

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -752,7 +752,7 @@ class MultimodKeyData:
         return self.det.train_ids
 
     def train_id_coordinates(self):
-        return self.det.train_ids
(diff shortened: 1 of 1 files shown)


philsmt:
I would say `.train_id_coordinates()` returning a list was a bug, given all other instances of this method in EXtra-data return `ndarray` (or `pd.Series`).

LGTM

takluyver:
Thanks both!

PR #505, opened 2024-03-21
Miscellaneous improvements to buffer_shape() method

takluyver:
While working on #504, I noticed a couple of things to fix about the `.buffer_shape()` method.

- It was only there for XTDF image KeyData objects, but it's useful for other multi-module KeyData.
- It was making unnecessary temporary arrays to calculate the shape (especially since #502)

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -806,8 +806,6 @@ class MultimodKeyData:
         """Get data as a plain NumPy array with no labels"""
         train_ids = np.asarray(self.det.train_ids)
 
-        module_dim = self.det.n_modules if module_gaps else len(self.modno_to_keydata)
-
         out_shape = self.buffer_shape(module_gaps, roi)
 
         if out is None:
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -766,10 +766,22 @@ class MultimodKeyData:
     def ndim(self):
         return self._eg_keydata.ndim + 1
 
+    def buffer_shape(self, module_gaps=False, roi=()):
+        """Get the array shape for this data
+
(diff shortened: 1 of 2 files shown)


extra_data/components.py
```diff
@@ -796,9 +808,7 @@ def ndarray(self, *, fill_value=None, out=None, roi=(), astype=None, module_gaps
 
         module_dim = self.det.n_modules if module_gaps else len(self.modno_to_keydata)
```
philsmt, line 809: This does not seem to be required anymore.

extra_data/components.py
```diff
@@ -871,7 +881,12 @@ def buffer_shape(self, module_gaps=False, roi=()):
         interest on the inner dimensions of the data.
         """
         module_dim = self.det.n_modules if module_gaps else len(self.modno_to_keydata)
```
philsmt, line 883: While it is required here, it almost makes me wonder if it would be useful as standalone method `module_dim(module_gaps=False)`. It's indirectly possible noe with buffer_shape of course, or is there another way to figure out how many modules are there?

philsmt (approved):
Generally looks good to me. Should this be tested for non-XTDF detectors?

extra_data/components.py
takluyver: `det.n_modules` tells you how many modules the detector should have. `det[key].shape[0]` is how many modules we have this data for. I'd rather not make another public way to do this, and I prefer a tiny bit of duplication to the extra complexity of another private method.

extra_data/components.py
takluyver: Thanks, good catch.

philsmt:
Thanks, LGTM

PR #506, opened 2024-03-28
Add pulse_id_coordinates() & train_id_coordinates() for XTDF image data

takluyver:
@turkot I think this was another thing that you requested. I've gone for an API matching the existing `.train_id_coordinates()` method.

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -933,6 +933,14 @@ class XtdfImageMultimodKeyData(MultimodKeyData):
             a = a.copy()  # So you can't accidentally modify the internal array
         return a
 
+    def pulse_id_coordinates(self):
+        """Get an array of pulse IDs per-frame for this data"""
+        return self.det._collect_inner_ids('pulseId')
+
+    def cell_id_coordinates(self):
+        """Get an array of memory cell IDs per-frame for this data"""
+        return self.det._collect_inner_ids('cellId')
+
     @property
     def dimensions(self):
         ndim_inner = self.ndim - 2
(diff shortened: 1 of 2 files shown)


turkot:
Hello @takluyver , thank you, this looks good. Do you plan to add similar API for non-xtdf detectors? (JUNGFRAU :roll_eyes: )

takluyver:
I wondered about that - it seemed a bit confusing to do a very similar API when JUNGFRAU data is so different, with memory cells on a separate axis. I could do it like e.g. `jf.cell_ids()` returning an array of length 1 or length 16, if you like?

takluyver:
I don't quite follow that - what would the 2 dimensions be? The same as `jf['data.memoryCell'].ndarray()`?

I might try implementing my idea for now so we have something concrete to discuss.

turkot:
Sorry, probably my comment was a bit confusing.
What I meant is that for JUNGFRAU 'cellId' has 2 dimensions, e.g.:
`cellId	[uint8: 48656 × 8]`
[but jf.cell_id_coordinates() should better return (8, 48656)]

But 'memoryCell' has the same dimensions as 'trainId' (or Xtdf detector's 'pulseId'), e.g.:
`memoryCell	[uint64: 48656]`

takluyver:
Ah, right, you're looking at the CXI files, and I'm thinking about the source data & the Python API.

What I've implemented just gives you a length 16 array of the cell numbers, which should be the same across all tiles and all trains. So if you want to repeat that to align with an array of frames like in the CXI file, you could do `np.tile(jf.cell_ids(), ntrains)`

turkot:
Yes, sorry, I was approaching from the CXI perspective, which does not make sense here.

LGTM!

takluyver:
Thanks!

PR #509, opened 2024-04-09
Prepare releases notes for 1.17

takluyver:


docs/agipd_lpd_data.rst
```diff
@@ -10,12 +10,9 @@ DSSC and JUNGFRAU, pulling together the separate modules into a single array.
...
 
-   These detectors can record a lot of data. The ``.get_array()`` method
-   loads all of the selected data into memory, which may not be practical for
-   entire runs. You might need to think about iterating over trains, selecting
-   batches of trains from the run, or using `Dask arrays
-   <https://docs.dask.org/en/latest/array.html>`_.
-
+   These detectors can record a lot of data, so loading it all into memory at
+   once may be impossible. You can use the :meth:`~LPD1M.split_trains` method
+   to work on a chunk at a time, or work with Dask arrays.
```
tmichela, line 15: I'd probably still leave the mention to iteration as it is relevant in some use cases, but up to you.

tmichela (approved):
LGTM

docs/agipd_lpd_data.rst
takluyver: It is, and the `.trains()` iterator is still there in the reference docs below. But IMO, if you're considering loading all the data into memory at once and the only obstacle is size, `.split_trains()` is generally a nicer approach than `.trains()`. The advantage of `.trains()` is for doing streaming-style analysis, in a style which is easier to adapt to online processing - but that's not what this note is aimed at.

I don't think this note needs to be exhaustive, so I'll leave it as is for now. We can of course always change it again later. :slightly_smiling_face:

takluyver:
Thanks!

PR #511, opened 2024-04-12
Omit virtual layout for mask if it is missing in the original JUNGFRAU data.

turkot:
Such check is implemented for Xtdf, but was missing for JUNGFRAU.

### DIFF
--- a/extra_data/write_cxi.py
+++ b/extra_data/write_cxi.py
@@ -429,11 +429,13 @@ class JUNGFRAUCXIWriter(VirtualCXIWriterBase):
             self.image_label: VLayout(
                 shape, dtype=image_grp[self.image_label].dtype),
             'gain': VLayout(shape, dtype=image_grp['gain'].dtype),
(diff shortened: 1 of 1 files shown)


extra_data/write_cxi.py
```diff
@@ -429,11 +429,13 @@ def collect_data(self):
             self.image_label: VLayout(
                 shape, dtype=image_grp[self.image_label].dtype),
             'gain': VLayout(shape, dtype=image_grp['gain'].dtype),
```
takluyver, line 431: Do we need to do the same for gain, or does raw JUNGFRAU data include gain?

turkot: Data for the gain in JUNGFRAU raw data is stored as a separate array in the same way as in proc. It is only for raw Xtdf data (at least I'm sure for AGIPD) that the gain stage is stored as an extra dimension of the data array.

I've been able to generate a vds file for the recent JUNGFRAU raw data after applying this small change.

takluyver:
Super, thanks. In that case, this LGTM.

PR #514, opened 2024-04-19
Add source, key & alias completions for IPython

takluyver:
A picture is worth a thousand words:

image

I've tested all three interactively.

Thanks to @JamesWrigley for reminding me of this hook. It was actually [something](https://github.com/ipython/ipython/pull/9289) that I implemented in IPython. :grin:

### DIFF
--- a/extra_data/aliases.py
+++ b/extra_data/aliases.py
@@ -36,6 +36,9 @@ class AliasIndexer:
 
         raise TypeError('expected alias or (source alias, key) tuple')
 
+    def _ipython_key_completions_(self):
+        return list(self.data._aliases.keys())
+
     def __contains__(self, aliased_item):
         try:
             self[aliased_item]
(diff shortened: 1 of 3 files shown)


philsmt:
This is awesome!

I am wondering if we can somehow avoid suggesting a `DataCollection` can only be indexed by sources. Maybe add some notion behind each source string it could optionally be followed by keys, even if we cannot reasonably include them? Even a count of keys seems overly expensive, unless we'd use `sel_keys` if present.

takluyver:
I see what you mean, but I don't see a great way to do that in completions. This `_ipython_key_completions_` hook is a very convenient API, but not a very powerful one - it doesn't give you any ways to insert a comment (which would be annoying anyway) or annotate the results at all.

So I'll argue that we should just keep this simple and live with the limitations.

philsmt:
LGTM

takluyver:
Thanks!

PR #515, opened 2024-04-19
Make aliases case-insensitive, and allow - & _ interchangeably

takluyver:
From the discussion in chat. I'm still open to people saying it's better not to do this, but it was easy enough to implement so we've got a concrete proposal to consider. :slightly_smiling_face: 

The normalised form is lowercase with dashes, e.g. `sa3-xgm`. The tests were already based on this pattern,  so this required fewest changes, and it's also the easiest to type, at least on UK & US keyboards, not requiring the shift key.

### DIFF
--- a/extra_data/aliases.py
+++ b/extra_data/aliases.py
@@ -10,6 +10,7 @@ class AliasIndexer:
         self.data = data
 
     def _resolve_any_alias(self, alias):
+        alias = alias.lower().replace('_', '-')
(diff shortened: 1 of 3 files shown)


philsmt:
Thanks, LGTM

MadsSJakobsen:
looks good to me also.

takluyver:
Thanks!

PR #518, opened 2024-05-03
API to load detector data with mask

takluyver:
This continues the KeyData pattern, so you can do:

```python
jf_det.masked_data().select_trains(np.s_[2000:2500]).ndarray()
```

Which is equivalent to:

```python
mask = jf_det['data.mask'].select_trains(np.s_[2000:2500]).ndarray() != 0
imgs = jf_det['data.adc'].select_trains(np.s_[2000:2500]).ndarray()
imgs[mask] = np.nan
```

- [x] Add tests

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -196,6 +196,7 @@ class MultimodDetectorBase:
           The replacement value to use for masked data. By default this is NaN.
         """
         key = key or self._main_data_key
+        self[self._mask_data_key]  # Check that the mask is there
         return DetectorMaskedKeyData(
             self, key, mask_key=self._mask_data_key,
             mask_bits=mask_bits, masked_value=masked_value
@@ -527,6 +528,7 @@ class XtdfDetectorBase(MultimodDetectorBase):
         """
         key = key or self._main_data_key
         assert key.startswith('image.')
+        self[self._mask_data_key]  # Check that the mask is there
         return XtdfMaskedKeyData(
             self, key, mask_key=self._mask_data_key,
             mask_bits=mask_bits, masked_value=masked_value
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -178,6 +178,9 @@ class MultimodDetectorBase:
     def __getitem__(self, item):
         return MultimodKeyData(self, item)
 
+    def __contains__(self, item):
+        return all(item in self.data[s] for s in self.source_to_modno)
+
     def masked_data(self, key=None, *, mask_bits=None, masked_value=np.nan):
         """Combine corrected data with the mask in the files
 
@@ -192,12 +195,17 @@ class MultimodDetectorBase:
           (e.g. 'data.adc').
         mask_bits: int or list of ints
           Reasons to exclude pixels, as a bitmask or a list of integers.
-          By default, all types of bad pixel are masked out.
+          By default, all types of bad pixel are masked out. See the possible
+          values at: https://extra.readthedocs.io/en/latest/calibration/#extra.calibration.BadPixels
         masked_value: int, float
           The replacement value to use for masked data. By default this is NaN.
         """
         key = key or self._main_data_key
-        self[self._mask_data_key]  # Check that the mask is there
+        if self._mask_data_key not in self:
+            raise RuntimeError(
+                f"This data doesn't include a mask ({self._mask_data_key}). "
+                f"You might be using raw instead of corrected data."
+            )
         if isinstance(mask_bits, Iterable):
             mask_bits = self._combine_bitfield(mask_bits)
         return DetectorMaskedKeyData(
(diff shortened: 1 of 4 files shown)


extra_data/components.py
```diff
@@ -176,6 +177,31 @@ def __init__(self, data: DataCollection, detector_name=None, modules=None,
     def __getitem__(self, item):
         return MultimodKeyData(self, item)
 
+    def masked_data(self, key=None, *, mask_bits=None, masked_value=np.nan):
```
philsmt, line 180: Since bit logic is not generally intuitive for pragmatic programmers, allowing to pass bad pixel bits by iterable may be nice here as well.

takluyver: That makes sense. I guess we should also expose the bad pixel flags somewhere - I don't think we have those at the moment. Maybe they belong in `extra.calibration` ?

philsmt: Definitely, EXtra was the intended target for some time now. `extra.calibration` makes sense given it's used in calibration data.

takluyver: Just to record here, I opened https://github.com/European-XFEL/EXtra/pull/172 to do that.

extra_data/components.py
```diff
@@ -1102,6 +1191,12 @@
 
         return arr
 
+
+class XtdfMaskedKeyData(DetectorMaskedKeyData, XtdfImageMultimodKeyData):
```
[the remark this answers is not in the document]
takluyver: I believe this is a false positive caused by CodeQL not fully understanding `super()` with multiple inheritance. The `__init__` method is inherited from `DetectorMaskedKeyData`, which does call `super().__init__(...)`. That calls the next `__init__` in the method resolution order - not just the parent class - so when it's creating an `XtdfMaskedKeyData`, it will find `XtdfImageMultimodKeyData.__init__()`.

```
In [2]: XtdfMaskedKeyData.__mro__
Out[2]: 
(extra_data.components.XtdfMaskedKeyData,
 extra_data.components.DetectorMaskedKeyData,
 extra_data.components.XtdfImageMultimodKeyData,
 extra_data.components.MultimodKeyData,
 object)
```

I usually avoid multiple inheritance because even if it works, it's a pain to think about stuff like this. In this instance, I started writing the masked KeyData classes as wrappers around the existing classes (has-a relationships rather than is-a), but it was going to be several times more code to provide the same methods and attributes again. So I decided using inheritance was the lesser of two evils.

extra_data/components.py
```diff
@@ -176,6 +178,40 @@ def __init__(self, data: DataCollection, detector_name=None, modules=None,
...
+          The data key to look at, by default the main data key of the detector
+          (e.g. 'data.adc').
+        mask_bits: int or list of ints
+          Reasons to exclude pixels, as a bitmask or a list of integers.
+          By default, all types of bad pixel are masked out.
+        masked_value: int, float
+          The replacement value to use for masked data. By default this is NaN.
+        """
+        key = key or self._main_data_key
+        self[self._mask_data_key]  # Check that the mask is there
```
tmichela, line 200: I would maybe raise a more explicit error message here if that happens.

tmichela (approved):
I would maybe add a link to the [BadPixel](https://extra.readthedocs.io/en/latest/calibration/#extra.calibration.BadPixels) values in the docstrings, and maybe a small usage example.

LGTM

extra_data/components.py
```diff
@@ -1625,6 +1740,7 @@ class JUNGFRAU(MultimodDetectorBase):
         r'(MODULE_|RECEIVER-|JNGFR)(?P<modno>\d+)'
     )
     _main_data_key = 'data.adc'
+    _mask_data_key = 'data.mask'
```
philsmt, line 1743: This is probably not an urgent problem to block this PR, but I think this would not copy the mask's compression and could blow them up quite a bit. I suspect in the mask case, compression is always a performance benefit.

philsmt (approved):
Thanks, this is sorely needed! LGTM.

extra_data/components.py
takluyver: Sorry, I'm a bit confused - where would the compression be copied to?

philsmt: `data.mask` is a product of offline corrections and generally stored compressed. If I'm not mistaken, this `FileWriter` will not maintain the compression and significantly blow up the data size.

takluyver: Ah, is this folded diff deceiving you?

image

This line is in the JUNGFRAU class, not FramesFileWriter. I don't think this MR affects writing files at all.

philsmt: Uhm, what made GitHub skip so many class definitions?

In any case, ignore the comment, sorry.

PR #519, opened 2024-05-08
Prevent iteration over a KeyData object

takluyver:
By adding `__getitem__` we accidentally made KeyData iterable. And in fact it behaved as an infinite iterable, so simply calling `list(kd)` could go into an infinite loop. I ran into this by making a mistake in a DAMNIT context file.

I can imagine we might want it to be iterable at some point, maybe yielding train IDs or data like `.trains()`. But for now, avoiding the infinite loop seems like a useful step.

### DIFF
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -209,6 +209,8 @@ class KeyData:
     def __getitem__(self, item):
         return self.select_trains(item)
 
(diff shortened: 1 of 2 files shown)


tmichela:
LGTM

takluyver:
Thanks :slightly_smiling_face:

PR #520, opened 2024-05-10
Allow kd[trains] for multi-module KeyData objects

takluyver:
Regular KeyData objects allow selecting trains like `xgm_intensity[:100]` . But the objects for multi-module detectors, which generally aim to have a similar API, require something like `lpd_mask.select_trains(np.s_[:100])`. This makes `lpd_mask[:100]` work as well.


philsmt:
LGTM

takluyver:
Thanks!

PR #522, opened 2024-05-17
Allow skipping specific checks in extra-data-validate

takluyver:
This is a partial response to #503.

The idea is that `extra-data-validate` is made up of a series of named checks, and you can pass e.g. `--skip control_timestamps_order` to skip the check with that name. So far only the per-file checks can be controlled like this, but if we like the structure, we could add the per-directory checks (so far only one, checking the JSON cache).

Specifically, the `@file_check` decorator makes a function into a named check. The function name is reused as the check name (so we'll need to remember not to rename these functions too readily).

### DIFF
--- a/extra_data/validation.py
+++ b/extra_data/validation.py
@@ -1,4 +1,4 @@
-from argparse import ArgumentParser
+from argparse import ArgumentParser, Action
 from multiprocessing import Pool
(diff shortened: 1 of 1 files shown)


philsmt:
LGTM

PR #525, opened 2024-06-10
Fix log to stderr when skipping files due to locality

philsmt:
Frankly I'm still not sure how we ended up here, but the fix seems fairly straightforward.


takluyver:
LGTM

PR #526, opened 2024-06-11
Add optional index group filter to SourceData.one_key()

philsmt:
During the offline data reduction trials, calls to `SourceData._get_index_group_sample()` ended up being a significant performance problem, as the [current implementation](https://github.com/European-XFEL/EXtra-data/blob/1.17.0/extra_data/sourcedata.py#L101) triggers a call to the fairly expensive `SourceData.keys()`.

Initially this was fixed by overwriting this method only within [`exdf-tools`](https://git.xfel.eu/dataAnalysis/exdf-tools/-/commit/0509da5e3fb66965d470186cd189cc54cdd13e01), this MR means to upstream these changes now.

Since the improved implementation is based on `SourceData.one_key()`, I decided to augment this method by an optional keyword argument `index_group` and entirely replace the usage of `_get_index_group_sample()`.

Something I noticed in the implementation is that `FileAccess.get_one_key()` (and in extension `SourceData.one_key()`) may technically return `None` if a source has no keys. While the selection mechanism prevents no keys to be selected, it does not seem to prevent files with key-less sources. I would not expect the current DAQ or calibration to create such files, but it's not impossible for someone external to prepare them. For now I decided to not change any behaviour, but preserve the one from `_get_index_group_sample()` and raise an exception if asking for a non-existing index group.

### DIFF
--- a/extra_data/file_access.py
+++ b/extra_data/file_access.py
@@ -467,25 +467,51 @@ class FileAccess(metaclass=MetaFileAccess):
         self._keys_cache[source] = res
         return res
 
-    def get_one_key(self, source):
+    def get_one_key(self, source, index_group=None):
         """Similar to get_keys(), except it returns only a single key for performance"""
+
+        # Use empty prefix if no index group filter is active.
+        prefix = index_group + '.' if index_group is not None else ''
+
         if source in self._keys_cache:
-            return next(iter(self._keys_cache[source]))
-        elif self._known_keys[source]:
-            return next(iter(self._known_keys[source]))
+            for key in self._keys_cache[source]:
+                if key.startswith(prefix):
+                    return key
+
+            # _keys_cache is a complete set, so this point can only be
+            # reached for a key-less source (currently assumed to no
+            # exist) or a non-existing index group was passed.
+            raise ValueError(f'{index_group} not an index group of `{source}`')
+
+        if self._known_keys[source]:
+            for key in self._known_keys[source]:
+                if key.startswith(prefix):
+                    return key
 
         if source in self.control_sources:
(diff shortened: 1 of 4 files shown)


philsmt:
I separated the changes to `FileAccess` and `SourceData` in separate commits for easier review.

extra_data/file_access.py
```diff
@@ -467,25 +467,46 @@ def add_key(key, value):
...
+        # Use empty prefix if no index group filter is active.
+        prefix = index_group + '.' if index_group is not None else ''
+
         if source in self._keys_cache:
-            return next(iter(self._keys_cache[source]))
-        elif self._known_keys[source]:
-            return next(iter(self._known_keys[source]))
+            for key in self._keys_cache[source]:
+                if key.startswith(prefix):
+                    return key
```
takluyver, line 479: `_keys_cache` should be a complete set, so if we get past this loop without finding any match then we can give up already. `_known_keys` below can be partial, by contrast.

takluyver:
Other than that one comment, LGTM.

I agree that it makes sense to throw an error for a non-existent index group. :+1:

extra_data/file_access.py
philsmt: Ah, good point. This only allows us to throw an early exception though in the case of a non-existing index group. I added that after the loop.

philsmt:
I added two additional minor changes for `FileAccess:

* Fixed the branch on line 500 to test for `None` identity rather than boolean `False`
* Fixed the test to try once after asking for keys (to hit cache) rather than after first ask (which does not populate the cache)

Will merge later today unless something comes up.

PR #527, opened 2024-06-12
Add concept of legacy source names

philsmt:
The calibration team wants to discontinue the practice of writing corrected data under the same source name as raw data, but instead switch the _type_ part of the source name from `DET` to `CORR`. For limited backwards compatibility, we aim to [insert soft links](https://git.xfel.eu/calibration/pycalibration/-/merge_requests/872/) under the old source name in corrected files. While this would work transparently in EXtra-data, some more support for this concept seems prudent, which I'd like to start with this MR.

It introduces the concept of _legacy sources_ throughout the `FileAccess`, `DataCollection` and `SourceData` APIs. For now I limited this solely to tracking, i.e. it has no impact on any business logic when accessing files or data:

* When reading sources, `FileAccess` will probe the source's leaf object for being a soft link and record its target. The source is counted as a regular source otherwise. For performance reasons, I limited this to `INSTRUMENT` sources for now (simple benchmarking suggests tens of ms for this operation when cold)
* Legacy sources are part of the `run_files_map` and will invalidate any existing cache.
* `SourceData` objects know whether they're a legacy source through a non-`None` value of `SourceData.legacy`, `DataCollection` has a `legacy_sources: dict` property.
* When accessing the `SourceData` object of a legacy source via a `DataCollection`, a `DeprecationWarning` is emitted.
* `DataCollection.info()` tracks legacy sources separately in their own section alongside their target.

For now it does not yet touch `data='all'`, which will still kick out the raw data. I would add tests once we're happy with the design.

As part of some earlier tests, I also added support for multiple XTDF detectors in a single `DataCollection`. I'm happy to remove this, but it seems useful to have for the future (e.g. AGIPD1M and AGIPD4M in a single run).

@takluyver @tmichela @dgoeries

### DIFF
--- a/extra_data/read_machinery.py
+++ b/extra_data/read_machinery.py
@@ -16,7 +16,7 @@ import numpy as np
 log = logging.getLogger(__name__)
 
 DETECTOR_NAMES = {'AGIPD', 'DSSC', 'LPD'}
-DETECTOR_SOURCE_RE = re.compile(r'(.+(?:DET|CORR))/(\d+)CH')
+DETECTOR_SOURCE_RE = re.compile(r'(.+\/(?:DET|CORR))\/(\d+)CH')
 
 DATA_ROOT_DIR = os.environ.get('EXTRA_DATA_DATA_ROOT', '/gpfs/exfel/exp')
 
--- a/extra_data/read_machinery.py
+++ b/extra_data/read_machinery.py
@@ -16,7 +16,7 @@ import numpy as np
 log = logging.getLogger(__name__)
 
 DETECTOR_NAMES = {'AGIPD', 'DSSC', 'LPD'}
-DETECTOR_SOURCE_RE = re.compile(r'(.+)/DET/(\d+)CH')
+DETECTOR_SOURCE_RE = re.compile(r'(.+(?:DET|CORR))/(\d+)CH')
 
 DATA_ROOT_DIR = os.environ.get('EXTRA_DATA_DATA_ROOT', '/gpfs/exfel/exp')
 
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1394,9 +1394,26 @@ class DataCollection:
         print()
 
         if self.legacy_sources:
+            # Collect legacy souces matching DETECTOR_SOURCE_RE
+            # separately for a condensed view.
+            detector_legacy_sources = defaultdict(set)
+
             print(len(self.legacy_sources), 'legacy source names:')
             for s in sorted(self.legacy_sources.keys()):
-                print(' -', s, '->', self.legacy_sources[s])
+                m = DETECTOR_SOURCE_RE.match(s)
+
+                if m is not None:
+                    detector_legacy_sources[m[1]].add(s)
+                else:
+                    # Only print non-XTDF legacy sources.
+                    print(' -', s, '->', self.legacy_sources[s])
+
+            for legacy_det, legacy_sources in detector_legacy_sources.items():
+                canonical_mod = self.legacy_sources[next(iter(legacy_sources))]
+                canonical_det = DETECTOR_SOURCE_RE.match(canonical_mod)[1]
+
+                print(' -', f'{legacy_det}/*', '->', f'{canonical_det}/*',
+                      f'({len(legacy_sources)})')
             print()
 
     def plot_missing_data(self, min_saved_pct=95, expand_instrument=False):
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1249,30 +1249,30 @@ class DataCollection:
             # Include summary section for multi-module detectors unless
             # source details are enabled.
 
-            detector_modules = {}
+            sources_by_detector = {}
             for source in self.detector_sources:
                 name, modno = DETECTOR_SOURCE_RE.match(source).groups((1, 2))
-                detector_modules[(name, modno)] = source
+                sources_by_detector.setdefault(name, {})[modno] = source
 
-            # A run should only have one detector, but if that changes, don't hide it
-            detector_name = ','.join(sorted(set(k[0] for k in detector_modules)))
+            for detector_name in sorted(sources_by_detector.keys()):
+                detector_modules = sources_by_detector[detector_name]
 
-            print("{} XTDF detector modules ({})".format(
-                len(self.detector_sources), detector_name
-            ))
-            if len(detector_modules) > 0:
-                # Show detail on the first module (the others should be similar)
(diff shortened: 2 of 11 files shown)


takluyver:
Just to note, as part of the name change, we'll also want #468 (or something like it) so that the components layer can pick between corrected & raw data when both are 'visible' under different names.

extra_data/sourcedata.py
```diff
@@ -22,14 +22,15 @@ class SourceData:
...
-            is_single_run, inc_suspect_trains=True
+            self, source, *, sel_keys, train_ids, files, section, legacy,
+            is_single_run, inc_suspect_trains=True,
     ):
         self.source = source
         self.sel_keys = sel_keys
         self.train_ids = train_ids
         self.files: List[FileAccess] = files
         self.section = section
+        self.legacy = legacy
```
takluyver, line 33: Am I right in thinking that the value of `.legacy` is the new, preferred name whereas `.source` is the legacy name? That seems basically guaranteed to cause confusion. :wink: 

Would it work to skip constructing `SourceData` objects for legacy sources, and when you do `dc[legacy_src]`, you just get back the SourceData with the new name, a bit like another kind of alias?

If not that, let's at least come up with a better name for the attribute.

extra_data/read_machinery.py
```diff
@@ -16,7 +16,7 @@
 log = logging.getLogger(__name__)
 
 DETECTOR_NAMES = {'AGIPD', 'DSSC', 'LPD'}
-DETECTOR_SOURCE_RE = re.compile(r'(.+)/DET/(\d+)CH')
+DETECTOR_SOURCE_RE = re.compile(r'(.+(?:DET|CORR))/(\d+)CH')
```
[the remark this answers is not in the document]
philsmt: Yes, thanks, that got lost while tinkering with the regexp.

extra_data/reader.py
```diff
@@ -1377,6 +1393,12 @@ def keys_detail(s, keys, prefix=''):
 
         print()
 
+        if self.legacy_sources:
+            print(len(self.legacy_sources), 'legacy source names:')
+            for s in sorted(self.legacy_sources.keys()):
+                print(' -', s, '->', self.legacy_sources[s])
+            print()
```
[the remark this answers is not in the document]
philsmt: I feel rather strongly to make it visible in `.info()`, it's one of the main motivations for this MR. But I agree with your point about XTDF sources. Would it be fine if XTDF multi-module sources are collated?

extra_data/sourcedata.py
philsmt: > Would it work to skip constructing `SourceData` objects for legacy sources, and when you do `dc[legacy_src]`, you just get back the SourceData with the new name, a bit like another kind of alias?

👀 
I would actually prefer that approach, but was hesitant to be radical. Happy to have a closer look at the needed changes.

extra_data/reader.py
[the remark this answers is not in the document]
philsmt: Alternatively, the target source could be annotated?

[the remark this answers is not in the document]
takluyver: My leaning would still be not to show them, but if you can condense it down to 1-2 lines in the expected case where this applies to an entire big detector, I don't feel strongly about it.

extra_data/sourcedata.py
takluyver: I suspect it will be more work in some ways, because it implies you have to carry a dict of legacy names through all of the DataCollection -> DataCollection operations (select, union, with_aliases...). With the present approach you can reconstruct it from the sources after each of those, and there are fewer SourceData -> SourceData operations to fix.

But it might be neater once that's done.

extra_data/reader.py
```diff
@@ -255,6 +263,14 @@ def _get_source_data(self, source):
         if source not in self._sources_data:
             raise SourceNameError(source)
 
+        sd = self._sources_data[source]
+
+        if sd.is_legacy:
+            warn(f"{source} is a legacy name for {self.legacy_sources[source]}. "
+                 f"Access via this name will be removed at a future data.",
```
kakhahmed, line 270: nitpicking comment
```suggestion
                 "Access via this name will be removed at a future data.",
```

Also, I assume here you don't mean at a future date but for future data.?

philsmt: I did originally mean *date*, but actually referencing future *data* is a good idea. We're not going to remove the legacy name from existing files, just stop adding it. Thanks!

extra_data/sourcedata.py
philsmt: As we also discussed in person, I did not go with a more radical approach given the complexity it introduces with `DataCollection.select()`. We'd either have to keep references to de-selected data around without actually using them in most places, or replace selections automatically by their canonical names. The latter however feels even more instrusive than returning a different `SourceData` object, as it would e.g. result in

```python3
mod0_name not in run.select(mod0_name).all_sources
```

Still, I changed the `SourceData.legacy` property to `SourceData.canonical_name`, which is populated for all `SourceData` objects now. `SourceData.is_legacy` is `True` if and only if this differs from `SourceData.source`.

extra_data/reader.py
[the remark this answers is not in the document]
philsmt: Legacy sources fulfilling `DETECTOR_SOURCE_RE` are now collated into a single line, under the assumption both legacy and canonical name is covered by the regexp.

philsmt:
This PR is now fully ready for review, I added tests for the new APIs.

For this purpose, `extra_data.mockdata.detectors.DetectorModule` can now optionally inject a legacy name and I added a new fixture `mock_spb_modern_proc_run` containing this. This could easily be extended to any device, but I would use this legacy pattern sparingly so we can hold off for now.

takluyver:
Thank-you, LGTM

philsmt:
Just to make sure, we all agree to invalidate existing run files maps?

takluyver:
I'm OK with that. I don't want to invalidate that cache too frequently - that defeats the point of caching - but once in a while I don't think it's a big deal.

If we wanted to get clever, we could say we invalidate it only for proc data, since it doesn't seem likely that the raw data will ever have links like this. But I suspect it's not worth the extra complexity to manage that, as the lower levels of EXtra-data don't have a raw/proc distinction so far.

philsmt:
For the record, I did some benchmarking picking random files across all proposals to avoid caching. Looping over all INSTRUMENT sources, probing a single source whether it's a soft link or not takes between 500 us and 1 ms, 3.5 ms on average per file.

PR #529, opened 2024-06-17
Require Python >= 3.8

takluyver:
This is still fairly conservative, but we're still using 3.8 for offline calibration at present, so I don't want to make extra work for ourselves if updating that gets held up.

### DIFF
--- a/.github/dependabot/constraints.txt
+++ b/.github/dependabot/constraints.txt
@@ -1,5 +1,5 @@
 cloudpickle<=3.0.0
-coverage==6.2
+coverage==6.5
 h5py<3.12.0
 karabo-bridge==0.7.0
 msgpack<=1.0.8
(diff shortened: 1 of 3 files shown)


takluyver:
For the 3.11 tests, pytest crashes basically as soon as it starts. :confused: 

```
   python3 -m pytest -v --nbval-lax --current-env --cov=extra_data --cov-report=xml
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.11.9/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.11.9/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.9/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.9/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.9/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.11.9/x64/lib
    MPLBACKEND: agg
Error: Process completed with exit code 1.
```

philsmt:
Thanks, LGTM!

PR #531, opened 2024-06-18
Numpy 2 compatibility fixes

takluyver:
I've pushed the same commits to #530 to test with Numpy 2.0; I'm opening this PR to verify that I haven't broken things for Numpy 1.x.


takluyver:
Never mind, this was actually redundant. Numpy 2.0 requires Python 3.9 or above, and since we still have 3.8 in the CI, we're still testing on Numpy 1.24 as well.

PR #532, opened 2024-06-18
Fix handling of empty string as index group for control sources

philsmt:
I introduced a bug in https://github.com/European-XFEL/EXtra-data/pull/526 when one calls `fa.get_one_key(<ctrl-src>, '')`, which is actually done by `SourceData.data_counts()`. As the test is only done for `index_group is not None`, an empty string causes `prefix` to be `.`.

I'm still slightly confused why the `test_sourcedata.test_data_counts_values()` does not trigger it, yet I can consistently trigger it by the added test case for `FileAccess`.

### DIFF
--- a/extra_data/file_access.py
+++ b/extra_data/file_access.py
@@ -471,7 +471,7 @@ class FileAccess(metaclass=MetaFileAccess):
         """Similar to get_keys(), except it returns only a single key for performance"""
 
         # Use empty prefix if no index group filter is active.
-        prefix = index_group + '.' if index_group is not None else ''
+        prefix = index_group + '.' if index_group else ''
 
         if source in self._keys_cache:
             for key in self._keys_cache[source]:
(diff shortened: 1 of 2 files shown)


takluyver:
> I'm still slightly confused why the test_sourcedata.test_data_counts_values() does not trigger it

I investigated, it looks like if we don't have the list of keys cached, `get_one_key` ends up returning a key with an extra `.` at the start, which gets translated to an HDF5 path containing a double slash `//`, which HDF5 is OK with. Possibly we should make `SourceData.__getitem__` stricter to catch things like this. :shrug:

philsmt:
Yes of course, caching is the difference. As it happens I found the bug only in the cached branch, because `prefix = '.'` will no longer match any key. Thanks.

takluyver:
Good catch :+1:

takluyver:
(Didn't mean to close it, hit the wrong button)

PR #538, opened 2024-07-23
Return train timestamps in local (i.e., EuXFEL) time not just UTC

fadybishara:
Added a non-positional boolean argument `local_time=False/True` that returns the time in the `Europe/Berlin` timezone.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1605,13 +1605,16 @@ class DataCollection:
         print('\tControls')
         [print('\t-', d) for d in sorted(ctrl)] or print('\t-')
 
-    def train_timestamps(self, labelled=False):
+    def train_timestamps(self, local_time=True, labelled=False):
         """Get approximate timestamps for each train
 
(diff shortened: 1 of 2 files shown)


philsmt:
The idea of a boolean argument returning "EuXFEL-time" is on me - the idea was that since all data is still taken at European XFEL (even if copied otherwise), I would expect essentially everyone to either care for UTC or for local time when data was acquired. It seems far-fetched to consider any other time would be relevant, i.e. someone creating EXDF data independently of data acquired at EuXFEL.

fadybishara:
The "use-case" for me arose when I was trying to query the DOOCS DAQ via `pydaq` -- and I needed to do this in specific time windows corresponding to certain runs. So for me, it was also the case that I needed EuXFEL local time.

I agree with @takluyver that naming the option `local_time` could be confusing -- I can name it, e.g., `exfel_local_time` or `exfel_time`

takluyver:
Done in #550, thanks!

PR #547, opened 2024-08-28
Ignore incomplete entries in JSON 'run files map' caches

takluyver:
There was a bug in `FileAccess` where if we read a valid run files map without `legacy_sources`, it wouldn't look at the files to identify legacy_sources.

We have tried to be clever, by using the available information from the cache in such cases and getting the new information from the files. This has some performance benefit, because you read less from the files, but it means more complexity, creating bugs like this one. And we still have to open every file to read something from it.

So I think we should go with the simpler option, ignoring incomplete cache entries entirely, even if that is a bit slower.

[We should also extend it to allow caching runs from the `red` folder into scratch, but that's a separate question]

cc @turkot @philsmt

### DIFF
--- a/extra_data/run_files_map.py
+++ b/extra_data/run_files_map.py
@@ -136,14 +136,13 @@ class RunFilesMap:
             dt = time.monotonic() - t0
             log.debug("Loaded cached files map in %.2g s", dt)
 
-    @staticmethod
-    def _cache_info_valid(info, file_stat: os.stat_result):
+    @classmethod
+    def _cache_info_valid(cls, info, file_stat: os.stat_result):
         # Ignore the cached info if the file size or mtime have changed, or
-        # if it is missing expected keys (likely
-        added_keys = {'suspect_train_indices', 'legacy_sources'}
+        # if it is missing expected keys (likely keys added more recently).
         return ((file_stat.st_mtime == info['mtime'])
                 and (file_stat.st_size == info['size'])
-                and added_keys.issubset(info.keys()))
+                and cls.expected_cache_keys.issubset(info.keys()))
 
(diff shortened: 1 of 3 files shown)


extra_data/run_files_map.py
```diff
@@ -121,13 +129,22 @@ def load(self):
...
                 self.files_data[filename] = info
 
         if loaded_data:
             dt = time.monotonic() - t0
             log.debug("Loaded cached files map in %.2g s", dt)
 
+    @staticmethod
+    def _cache_info_valid(info, file_stat: os.stat_result):
+        # Ignore the cached info if the file size or mtime have changed, or
+        # if it is missing expected keys (likely
```
philsmt, line 142:
```suggestion
        # if it is missing expected keys (likely)
```

extra_data/run_files_map.py
```diff
@@ -121,13 +129,22 @@ def load(self):
...
+        added_keys = {'suspect_train_indices', 'legacy_sources'}
+        return ((file_stat.st_mtime == info['mtime'])
+                and (file_stat.st_size == info['size'])
+                and added_keys.issubset(info.keys()))
```
philsmt, line 146: Hmmm, I'm wondering if checking `cls.expected_cache_keys.issubset(info.keys())` wouldn't be simpler.

philsmt (approved):
Caching and impure functions don't mix 😒 

Thanks for the fix, and I agree the performance benefit is likely negligible. LGTM!

extra_data/run_files_map.py
takluyver: It would, good idea.

PR #548, opened 2024-08-28
Allow caching 'red' run folders to scratch

takluyver:
At present, 'run files map' caches can only be created inside the folder itself (which largely means by `xred`).

cc @turkot


turkot:
lgtm!

PR #549, opened 2024-08-29
allow caching "open" run folders to scratch

tmichela:
@philsmt pointed out that `open` should be there as well.


takluyver:
LGTM, thanks

PR #550, opened 2024-09-09
Return timezone-aware values from train_timestamps() where possible.

takluyver:
This is an alternative to #538. The idea is that where possible, we return timezone-aware UTC timestamps, making it easier to convert them to local time after calling this method, if desired. E.g. with pandas you would do:

```python
run.train_timestamps(labelled=True).dt.tz_convert('Europe/Berlin')
```

Timezone-aware timestamps are AFAIK always better where possible. Unfortunately, Numpy datetime64 has no notion of timezones. I've also added an option to get timezone-aware Python `datetime` objects, which may be more familiar.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1608,7 +1608,7 @@ class DataCollection:
     def train_timestamps(self, labelled=False, *, pydatetime=False, euxfel_local_time=False):
         """Get approximate timestamps for each train
 
-        Timestamps are stored and returned in UTC (not local time).
+        Timestamps are stored and returned in UTC by default.
         Older files (before format version 1.0) do not have timestamp data,
         and the returned data in those cases will have the special value NaT
         (Not a Time).
@@ -1618,6 +1618,9 @@ class DataCollection:
         objects (truncated to microseconds) is returned, the same length as
         data.train_ids. Otherwise (by default), timestamps are returned as a
         NumPy array with datetime64 dtype.
+
+        *euxfel_local_time* can be True when either *labelled* or *pydatetime* is True.
+        In this case, timestamps are converted to the `Europe/Berlin` timezone.
         """
         arr = np.zeros(len(self.train_ids), dtype=np.uint64)
         id_to_ix = {tid: i for (i, tid) in enumerate(self.train_ids)}
@@ -1661,7 +1664,7 @@ class DataCollection:
         elif euxfel_local_time:
             raise ValueError(
                 'The euxfel_local_time option '
-                + 'can only be invoked if either labelled or pydatetime '
+                + 'can only be used if either labelled or pydatetime '
                 + 'are set to True'
             )
         return arr
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1642,7 +1642,7 @@ class DataCollection:
         arr[arr == epoch] = 'NaT'  # Not a Time
         if labelled:
             import pandas as pd
-            return pd.Series(arr, index=self.train_ids)
+            return pd.Series(arr, index=self.train_ids).dt.tz_localize('UTC')
         return arr
 
     def run_metadata(self) -> dict:
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1605,7 +1605,7 @@ class DataCollection:
         print('\tControls')
         [print('\t-', d) for d in sorted(ctrl)] or print('\t-')
 
-    def train_timestamps(self, labelled=False):
+    def train_timestamps(self, labelled=False, *, pydatetime=False):
         """Get approximate timestamps for each train
 
         Timestamps are stored and returned in UTC (not local time).
@@ -1614,8 +1614,10 @@ class DataCollection:
         (Not a Time).
 
         If *labelled* is True, they are returned in a pandas series, labelled
-        with train IDs. If False (default), they are returned in a NumPy array
-        of the same length as data.train_ids.
+        with train IDs. If *pydatetime* is True, a list of Python datetime
+        objects (truncated to microseconds) is returned, the same length as
+        data.train_ids. Otherwise (by default), timestamps are returned as a
+        NumPy array with datetime64 dtype.
         """
         arr = np.zeros(len(self.train_ids), dtype=np.uint64)
         id_to_ix = {tid: i for (i, tid) in enumerate(self.train_ids)}
@@ -1643,6 +1645,15 @@ class DataCollection:
         if labelled:
             import pandas as pd
             return pd.Series(arr, index=self.train_ids).dt.tz_localize('UTC')
+        elif pydatetime:
+            from datetime import datetime, timezone
+            res = []
+            for npdt in arr:
+                pydt = npdt.astype('datetime64[ms]').item()
+                if pydt is not None:  # Numpy NaT becomes None
+                    pydt = pydt.replace(tzinfo=timezone.utc)
+                res.append(pydt)
+            return res
         return arr
 
     def run_metadata(self) -> dict:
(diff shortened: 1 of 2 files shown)


fadybishara:
Thanks a lot @takluyver and @philsmt for the guidance and help! I added the functionality associated with the `euxfel_local_time` flag as well as a test `test_train_timestamps_local_time` for it.

However, there is a little problem, when Philipp and I were thinking about the timezone conversion in the case of `pydatetime=True`; this can be done with the `.astimezone` method which takes an instance of `datetime.tzinfo` as an input.

The options are for such an object are:
1. `zoneinfo.ZoneInfo('Europe/Berlin')`
2. `pytz.timezone('Europe/Berlin')`

We (Philipp and I) opted for option 1 because `zoneinfo` is part of the standard library while `pytz` is not. The caveat is... `zoneinfo` was added in 3.9 so the 3.8 test are failing :face_with_diagonal_mouth: and on top of that, it's not required in `setup.py` so we'd have to add it just for this...

Note that his is only a problem for the combination `pydatetime=True, euxfel_local_time=True` but not for `labelled=True, euxfel_local_time=True` since the conversion in latter case goes via pandas and is trivial in comparison.

There are, of course, less elegant solutions but before resorting to those, I thought I'd put this up for a brief discussion first.

extra_data/reader.py
```diff
@@ -1642,7 +1644,26 @@ def train_timestamps(self, labelled=False):
...
+            return series.dt.tz_convert('Europe/Berlin') if euxfel_local_time else series
+        elif pydatetime:
+            from datetime import datetime, timezone
+            res = []
+            for npdt in arr:
+                pydt = npdt.astype('datetime64[ms]').item()
+                if pydt is not None:  # Numpy NaT becomes None
+                    pydt = pydt.replace(tzinfo=timezone.utc)
+                    if euxfel_local_time:
+                        from zoneinfo import ZoneInfo
```
philsmt, line 1657: Oops, looks like I missed that `zoneinfo` was only added in 3.9. The test for 3.8 fails expectedly at this point. Shall we use `pytz` instead, which is likely already installed as a dependency of something?

takluyver: I think we could also drop 3.8 support - IIRC we were keeping it for calibration, but calibration has moved to Python 3.11. @philsmt do you know of any other reason we need to keep 3.8?

fadybishara: Dropping 3.8 and sticking with the standard library would be my preference if that is possible.

extra_data/tests/test_reader_mockdata.py
```diff
@@ -743,15 +743,49 @@ def test_train_timestamps(mock_scs_run):
...
+      
+    # First, the pydatetime case
+    tss = run.train_timestamps(pydatetime=True)
+    tss_berlin = run.train_timestamps(pydatetime=True, euxfel_local_time=True)
+
+    # The time difference between UTC and Europe/Berlin can only be
+    # one or two hours depending on daylight savings
+    assert all(
+        t2.replace(tzinfo=None) - t1.replace(tzinfo=None) == del1h or
+        t2.replace(tzinfo=None) - t1.replace(tzinfo=None) == del2h
```
takluyver, line 773: The `datetime` object has a `.utcoffset()` method you can use instead of these subtractions - it also gives a timedelta.

extra_data/tests/test_reader_mockdata.py
```diff
@@ -743,15 +743,49 @@ def test_train_timestamps(mock_scs_run):
...
+    tss_berlin = run.train_timestamps(labelled=True, euxfel_local_time=True)
+  
+    dtss = tss_berlin.dt.tz_localize(None) - tss.dt.tz_localize(None) 
+    assert all(dtss == del1h) or all(dtss == del2h)
+    
+    # Finally, check that ValueError is raised if euxfel_local_time is used
+    # on its own
+    with pytest.raises(ValueError):
+        run.train_timestamps(pydatetime=False, euxfel_local_time=True)
+        run.train_timestamps(labelled=False, euxfel_local_time=True)
```
takluyver, line 788: Just to note, putting two lines inside the `with` block here checks that *either* of them raises the error, not that both do separately.

It doesn't matter here since they're logically the same, though. Let's do it just once:

```suggestion
    with pytest.raises(ValueError):
        run.train_timestamps(pydatetime=False, labelled=False, euxfel_local_time=True)
```

fadybishara: Ah, thanks for pointing this out -- yes, since the condition is triggered by a fall-through an if/elif checking it once checks it for both, that's what you mean, right?

extra_data/reader.py
```diff
@@ -1642,7 +1644,26 @@ def train_timestamps(self, labelled=False):
...
+                        pydt = pydt.astimezone(ZoneInfo('Europe/Berlin'))
+                res.append(pydt)
+            return res
+        elif euxfel_local_time:
+            raise ValueError(
+                'The euxfel_local_time option '
+                + 'can only be invoked if either labelled or pydatetime '
```
takluyver, line 1664: Nitpick: you invoke (or call) a method, not a parameter.

```suggestion
                + 'can only be used if either labelled or pydatetime '
```

extra_data/reader.py
```diff
@@ -1605,7 +1605,7 @@ def train_info(self, train_id):
         print('\tControls')
         [print('\t-', d) for d in sorted(ctrl)] or print('\t-')
 
-    def train_timestamps(self, labelled=False):
+    def train_timestamps(self, labelled=False, *, pydatetime=False, euxfel_local_time=False):
         """Get approximate timestamps for each train
 
         Timestamps are stored and returned in UTC (not local time).
```
takluyver, line 1611:
```suggestion
        Timestamps are stored and returned in UTC by default.
```

extra_data/reader.py
```diff
@@ -1614,8 +1614,10 @@ def train_timestamps(self, labelled=False):
         (Not a Time).
 
         If *labelled* is True, they are returned in a pandas series, labelled
-        with train IDs. If False (default), they are returned in a NumPy array
-        of the same length as data.train_ids.
+        with train IDs. If *pydatetime* is True, a list of Python datetime
+        objects (truncated to microseconds) is returned, the same length as
+        data.train_ids. Otherwise (by default), timestamps are returned as a
+        NumPy array with datetime64 dtype.
         """
```
takluyver, line 1621:
```suggestion

        *euxfel_local_time* can be True when either *labelled* or *pydatetime* is True.
        In this case, timestamps are converted to the `Europe/Berlin` timezone.
        """
```

takluyver (commented):
We should add the new option to the docstring as well.

extra_data/tests/test_reader_mockdata.py
takluyver: Yup - because we've made `False` the default, passing `pydatetime=False` is the same as not passing `pydatetime` at all.

extra_data/tests/test_reader_mockdata.py
fadybishara: Good point, the case with `euxfel_local_time=False` is tested separately so it's safe to assume that `t1` in this case is always correct and therefore sufficient to check the UTC offset. Will consolidate all your suggestions into a commit and fix this at the same time. Thanks!

extra_data/tests/test_reader_mockdata.py
```diff
@@ -743,15 +743,47 @@ def test_train_timestamps(mock_scs_run):
...
+    assert tss_ser.dt.tz is timezone.utc
+
+def test_train_timestamps_local_time(mock_scs_run):
+    run = RunDirectory(mock_scs_run)
+
+    del1h = timedelta(hours=1)
+    del2h = timedelta(hours=2)
+
+    # First, the pydatetime case
+    tss = run.train_timestamps(pydatetime=True)
```
takluyver, line 766:
```suggestion
```

PR #559, opened 2024-10-04
Selecting train IDs in DataCollection and SourceData

tmichela:
Allow using the `object[]` syntax to select train IDs in addition to source or keys for `DataCollection` and `SourceData`

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -274,7 +274,11 @@ class DataCollection:
         return self._sources_data[source]
 
     def __getitem__(self, item):
-        if isinstance(item, tuple) and len(item) == 2:
+        if (
+            isinstance(item, tuple) and
+            len(item) == 2 and
+            all(isinstance(e, str) for e in item)
+        ):
             return self._get_key_data(*item)
         elif isinstance(item, str):
             return self._get_source_data(item)
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -33,17 +33,17 @@ import h5py
 import numpy as np
 
 from . import locality, voview
+from .aliases import AliasIndexer
 from .exceptions import (MultiRunError, PropertyNameError, SourceNameError,
                          TrainIDError)
 from .file_access import FileAccess
 from .keydata import KeyData
-from .read_machinery import (DETECTOR_SOURCE_RE, FilenameInfo, by_id, by_index,
-                             find_proposal, glob_wildcards_re, same_run,
-                             select_train_ids, split_trains)
+from .read_machinery import (DETECTOR_SOURCE_RE, by_id, by_index,
+                             find_proposal, glob_wildcards_re, is_int_like,
+                             same_run, select_train_ids)
 from .run_files_map import RunFilesMap
 from .sourcedata import SourceData
 from .utils import available_cpu_cores
-from .aliases import AliasIndexer
 
 __all__ = [
     'H5File',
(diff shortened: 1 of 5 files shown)


extra_data/reader.py
```diff
@@ -278,8 +278,13 @@ def __getitem__(self, item):
...
         elif isinstance(item, str):
             return self._get_source_data(item)
+        elif (
+            isinstance(item, (by_id, by_index, list, np.ndarray, slice)) or
+            is_int_like(item)
+        ):
+            return self.select_trains(item)
 
-        raise TypeError("Expected data[source] or data[source, key]")
+        raise TypeError("Expected data[source], data[source, key] or data[train_selection]")
```
fadybishara, line 287: While trying to trigger this error, I ran across a surprising (to me) behavior, the following does not raise this error:
```python
run['SA3_XTD10_XGM/XGM/DOOCS:output', 2]
```
rather, it returns a very sensible object:
```python
<extra_data.SourceData source='SA3_XTD10_XGM/XGM/DOOCS:output' for 1 trains>
```

I suppose this is part of `train_selection`  but I couldn't easily find it documented anywhere.

extra_data/tests/test_sourcedata.py
```diff
@@ -114,6 +114,19 @@ def test_select_trains(mock_spb_raw_run):
...
     assert sel.keys() == xgm.keys()
 
+    sel = xgm[by_id[10020:10040]]
+    assert sel.train_ids == list(range(10020, 10040))
+
+    sel = xgm[by_index[:10]]
+    assert sel.train_ids == list(range(10000, 10010))
+
+    sel = xgm[10]
+    assert sel.train_ids == [10010]
```
fadybishara, line 124: Okay, I see that this is the same behavior I flagged above :heavy_check_mark: so it's only a matter of documentation *if* it's indeed missing (I could have simply missed it)

fadybishara:
I have a minor comment (see review), otherwise everything LGTM! The RTD should also be updated, right?

fadybishara (approved):
Modulo the minor comment, LGTM!

extra_data/reader.py
tmichela: huh, interesting side effect. but yes, since this essentially does `run[source][key]` and now that `SourceData[key]` can be used for filtering trains, it makes sense.
I'm fine keeping that behavior, unless others object.

tmichela:
I added a few lines in the docs, will merge in a few days (when Thomas is back) if no-one objects the hybrid behavior.

extra_data/reader.py
philsmt: Mhh, I *think* I would error on the side of caution here and raise an exception:
* Supporting `DataCollection[source, train_sel]` begs the question why not `DataCollection[source, key, train_sel]`
* It would hide (and make it really hard to debug) type errors in your code, say you call `run[source, key]` and due to some other bug `key` becomes an `int` - a wrong `str` would raise an exception, but a random `int` doesn't?

fadybishara: Well, as long as the random `int` is larger than zero and less than `len(run.train_ids)` but that's a large range, typically, and that's what you mean?

takluyver:
Thanks, the changes LGTM too (and I lean towards agreeing with Philipp - it seems worth blocking a `run[source, index]` selection to make clearer errors on simple mistakes)

tmichela:
Thanks all for the review!

PR #566, opened 2024-10-21
fix: file_filter result is not taken into account when opening a run

tmichela:


RobertRosca (approved):
LGTM :+1:

takluyver:
Thanks, good catch, LGTM.

PR #569, opened 2024-11-01
Combine raw & proc in open_run() by default

takluyver:
This adds a new `data='default'` option (which becomes the default) to combine raw & proc data, preferring raw where names clash. For new data, this means you can access corrected detector data with the new source names (e.g. `*/CORR/*`), alongside raw data with names like `*/DET/*`. For older data, there will be no difference from `data='raw'`.

- This & #468 together aren't 100% backwards compatible. Code that calls `open_run()` without passing `data=` and then uses a detector component like `AGIPD1M()` previously got raw data, but would now get corrected data when reading runs with the new source names. But hopefully this isn't common, and it's easily fixed.
- I'm not completely sure about when to error or warn on missing folders. At present, this errors if neither raw or proc run folders exist, and warns if only proc exists (somehow). It won't complain about a missing proc folder if run exists.

A large part of the diff is just moving some existing tests to a new file, so it may be easier to review by commit.

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -2011,7 +2011,7 @@ def open_run(
                 origin_dc = open_run(
                     proposal, run, data=origin, include=include,
                     file_filter=file_filter, inc_suspect_trains=inc_suspect_trains,
-                    parallelize=parallelize, aliases=aliases,
+                    parallelize=parallelize, aliases=aliases, _use_voview=_use_voview,
                 )
             except FileNotFoundError:
                 if origin not in absence_ok:
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -1946,7 +1946,7 @@ RunHandler = RunDirectory
 DEFAULT_ALIASES_FILE = "{}/usr/extra-data-aliases.yml"
 
 def open_run(
-        proposal, run, data='raw', include='*', file_filter=locality.lc_any, *,
+        proposal, run, data='default', include='*', file_filter=locality.lc_any, *,
         inc_suspect_trains=True, parallelize=True, aliases=DEFAULT_ALIASES_FILE,
         _use_voview=True,
 ):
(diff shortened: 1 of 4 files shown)


extra_data/reader.py
```diff
@@ -1993,26 +1994,35 @@ def open_run(
         the proposal directory path. By default it looks for a file named
         ``{}/usr/extra-data-aliases.yml``.
     """
+    absence_ok = set()
+    if data == 'default':
+        data = ['proc', 'raw']
+        absence_ok = {'proc'}
     if data == 'all':
```
philsmt, line 2001: It's somewhat nitpicking, but an `elif` makes the flow clearer.

extra_data/reader.py
```diff
@@ -1993,26 +1994,35 @@ def open_run(
...
+                    file_filter=file_filter, inc_suspect_trains=inc_suspect_trains,
+                    parallelize=parallelize, aliases=aliases,
+                )
             except FileNotFoundError:
-                warn(f'No data available for this run at origin {origin}')
+                if origin not in absence_ok:
+                    if base_dc is None:
+                        raise
+                    warn(f'No data available for this run at origin {origin}')
             else:
```
philsmt, line 2021: I would probably go for a `continue` in the `except` branch, and move the entire `else` branch to after the complete `try-except` to avoid a level of indentation for so much code. I realize this has been like this before (written by me 🙃), but the logic seems complicated enough now that it seems useful.

philsmt (approved):
I gave the tests only a brief look, as I suspect the non-`default` parts are copied verbatim.

The main logic looks good. LGTM, apart from some minor code flow recommendations.

extra_data/reader.py
takluyver: Thanks, done!

extra_data/reader.py
takluyver: Good point, done

PR #571, opened 2024-11-20
Fix det.masked_data().select_pulses() in XTDF detector components

takluyver:
We were loading the data with the pulse selection, but loading all pulses for the mask, and then failing to combine them.

@JunCEEE @philsmt


philsmt:
LGTM

takluyver:
Thanks!

PR #572, opened 2024-11-22
Update architecture page in docs

takluyver:
This was written before `SourceData`.

While I was looking at this, I also moved `open_run()` before `RunDirectory` in the reference docs, as `open_run()` is the main starting point for most use cases.


tmichela (approved):

takluyver:
Thanks!

PR #576, opened 2024-11-26
Avoid giant memory allocations when reading xtdf data with pulse selection

### LINKED ISSUE #573: Reading selected pulses from xtdf detector with virtual overview file runs out of memory
Hazem mentioned to me that he hit a memory error reading from one memory cell of one module in p6936 r50, even though the actual data to be read is only about 10 GB.

```python
run = open_run(6936, 50, data='raw')
lpd8 = LPD1M(run, modules=[8], parallel_gain=True)
lpd8['image.data'].select_pulses([1]).ndarray()
```

```text
MemoryError                               Traceback (most recent call last)
Cell In[12], line 1
----> 1 lpd8['image.data'].select_pulses([1]).ndarray()

File /gpfs/exfel/sw/software/mambaforge/22.11/envs/202402/lib/python3.10/site-packages/extra_data/components.py:1180, in XtdfImageMultimodKeyData.ndarray(self, fill_value, out, roi, astype, module_gaps)
   1178     mod_ix = (modno - self.det._modnos_start_at) if module_gaps else i
   1179     for chunk in kd._data_chunks:
-> 1180         self._read_chunk(chunk, reading_view[mod_ix], roi)
   1182 return out

File /gpfs/exfel/sw/software/mambaforge/22.11/envs/202402/lib/python3.10/site-packages/extra_data/components.py:1141, in XtdfImageMultimodKeyData._read_chunk(self, chunk, mod_out, roi)
   1125     continue
   1127 # Read a subset of pulses from the chunk:
   1128 
   1129 # Reading a non-contiguous selection in HDF5 seems to be slow:
   (...)
   1139 # zeros() uses calloc, so the OS can do virtual memory tricks.
   1140 # Don't change this to zeros_like() !
-> 1141 tmp = np.zeros(chunk.dataset.shape, chunk.dataset.dtype)
   1142 pulse_sel = np.nonzero(inc_pulses_chunk)[0] + chunk_slice.start
   1143 sel_region = (pulse_sel,) + roi

MemoryError: Unable to allocate 1.66 TiB for an array with shape (13947990, 1, 256, 256) and data type uint16
```

This looks like a combination of [this workaround for an HDF5 issue](https://github.com/European-XFEL/EXtra-data/blob/c2153c67e030b8a522b61ab934625ee69cbff9ad/extra_data/components.py#L1172-L1203) plus virtual overview files. We allocate an array the size of the data in 1 file, ask HDF5 to read into the relevant portions of it (which is faster than reading into a denser array) and then `numpy.compress()` it into the output array. But with virtual overview files, this allocates just one temporary array for the full length of the run.

Up to a point, Linux 'overcommits' memory, allowing you to allocate more than the system has, because it's common not to use it all. It looks like Maxwell nodes are set to allow 50% overcommit. This allocation is larger than that.

takluyver:
Fixes #573

To work around an HDF5 performance issue, we read pulse-selected data into a temporary array which remains mostly empty. This was being allocated per file, but then we introduced virtual overview files, which meant it was effectively allocated per run. Even though we don't use most of the space, there's still a limit on how much you can allocate.

The fix is in two parts:

- Allocate the temporary array to the same shape as the matched chunk, instead of the shape of the entire dataset. This means the indices for the temp array differ from those in the file by a constant offset, but HDF5 still considers the selections the same shape, which is important for good read performance.
- When finding chunks to read into the temp array (or direct to the output, without pulse selection), limit them to a maximum size. I've chosen 5 GB fairly arbitrarily.
  - This does lead me into thinking about loading chunks in parallel, but I wanted to solve the immediate issue before thinking about that.

With these changes, I can load a 10 GB single pulse selection from data that's 1.6 TB in total (the example in #573).

I don't think there's an easy way to test this; it's not meant to produce any visible change in behaviour except when memory allocation would fail.

### DIFF
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -1176,18 +1176,21 @@ class XtdfImageMultimodKeyData(MultimodKeyData):
             # Except it's fast if you read the data to a matching selection in
             # memory (one weird trick).
             # So as a workaround, this allocates a temporary array of the same
-            # shape as the dataset, reads into it, and then copies the selected
+            # shape as the full chunk, reads into it, and then copies the selected
             # data to the output array. The extra memory copy is not optimal,
             # but it's better than the HDF5 performance issue, at least in some
             # realistic cases.
             # N.B. tmp should only use memory for the data it contains -
             # zeros() uses calloc, so the OS can do virtual memory tricks.
             # Don't change this to zeros_like() !
-            tmp = np.zeros(chunk.dataset.shape, chunk.dataset.dtype)
-            pulse_sel = np.nonzero(inc_pulses_chunk)[0] + chunk_slice.start
-            sel_region = (pulse_sel,) + roi
+            tmp = np.zeros(
+                shape=inc_pulses_chunk.shape + chunk.dataset.shape[1:],
+                dtype=chunk.dataset.dtype
+            )
+            tmp_sel = np.nonzero(inc_pulses_chunk)[0]
+            dataset_sel = tmp_sel + chunk_slice.start
             chunk.dataset.read_direct(
-                tmp, source_sel=sel_region, dest_sel=sel_region,
+                tmp, source_sel=(dataset_sel,) + roi, dest_sel=(tmp_sel,) + roi,
             )
             # Where does this data go in the target array?
             tgt_start_ix = self._sel_frames[:tgt_slice.start].sum()
@@ -1195,10 +1198,8 @@ class XtdfImageMultimodKeyData(MultimodKeyData):
                 tgt_start_ix, tgt_start_ix + inc_pulses_chunk.sum()
             )
             # Copy data from temp array to output array
-            tmp_frames_mask = np.zeros(len(tmp), dtype=np.bool_)
-            tmp_frames_mask[pulse_sel] = True
             np.compress(
-                tmp_frames_mask, tmp[np.index_exp[:] + roi],
+                inc_pulses_chunk, tmp[np.index_exp[:] + roi],
                 axis=0, out=mod_out[tgt_pulse_sel]
             )
 
--- a/extra_data/components.py
+++ b/extra_data/components.py
@@ -1,6 +1,7 @@
 """Interfaces to data from specific instruments
 """
 import logging
+import math
 import re
 from collections.abc import Iterable
 from copy import copy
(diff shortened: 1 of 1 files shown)


philsmt:
Looks good, LGTM.

The word *chunk* is a bit unfortunate when working with HDF.

Do you have some rough performance numbers for different values of `length_limit`? I wonder if one would want to expose it as an optional argument if there's a significant impact, though I doubt it.

takluyver:
I tried before and after the second commit (which adds length_limit) loading 1 pulse per train for 1000 trains; the limit I set would break this into 13 chunks. As usual for reading data, timings are all over the place, but in both versions I got a best time a bit under 400 ms. It was similar with `_use_voview=False`, which (before `length_limit`) would split this data into 2-3 pieces (500 trains per file).

So it looks like the limit I've set doesn't make a big difference, and intuitively that makes sense - 5 GB should be big enough that the overhead of Python & HDF5 calls is swamped by reading (even if we're only actually reading a small fraction of that data).

I guess I should also try skipping the temporary array and reading frames one by one, to see how that compares.

philsmt:
I would recommend to not spend too much time trying to tune the parameters for the current serial loop. I recently made the observation again that almost *any* form of process parallelization trumps even the most carefully crafted sequential loop. So we can go with this PR as is, and rather spend the time parallelizing it, e.g. via pasha.

Context: I made a branch to parallelize [`AdqRawChannel.pulse_data()`](https://github.com/European-XFEL/EXtra/blob/master/src/extra/components/adq.py#L1088), but just trivially throwing `pasha` at it and working by train was hardly if at all slower than cleverly trying to distribute the chunked loop over workers.

takluyver:
OK, I just tested reading frame by frame with code like this (inside `_read_chunk`):

```python
# Read a subset of pulses from the chunk:
dataset_idxs = np.nonzero(inc_pulses_chunk)[0] + chunk_slice.start
# Where does this data go in the target array?
tgt_start_ix = self._sel_frames[:tgt_slice.start].sum()
for tgt_ix, read_ix in enumerate(dataset_idxs, start=tgt_start_ix):
    chunk.dataset.read_direct(
        mod_out, source_sel=(read_ix,) + roi, dest_sel=(tgt_ix,) + roi
    )
```

This is still several times slower - best time about 1.9 seconds compared to 0.4 s. And I think my test case, reading only 1 pulse out of every 510, is about the most favourable for this strategy. So I think the temporary arrays can stay.

takluyver:
Thanks for the review. :smiley:

PR #577, opened 2024-12-05
Move pyyaml to the default requirements.

turkot:


turkot:
Running this code:
```python
hdf5_data = open_run(6936, 220)
```
with default EXtra-data installation was throwing an error:
```python
File /gpfs/exfel/sw/software/extra-xwiz-alpha_env/lib/python3.10/site-packages/extra_data/reader.py:688, in DataCollection._load_aliases_from_file(self, aliases_path)
    685         data = json.load(f)
    687 elif aliases_path.suffix in ['.yaml', '.yml']:
--> 688     import yaml
    690     with open(aliases_path, 'r') as f:
    691         data = yaml.safe_load(f)

ModuleNotFoundError: No module named 'yaml'
```
Since opening run data is the base functionality of EXtra-data, I think it makes sense to move 'pyyaml' to the default requirements.

takluyver:
Just to record, the original logic of making this optional was that it's only needed if you specify a yaml file for aliases - but then we added default aliases from a yaml file in the proposal folder. The alternative would be to skip the aliases if yaml fails, but this approach seems better.

LGTM!

PR #581, opened 2024-12-17
Option to get run values without timestamps

takluyver:
To parallel `src.keys(inc_timestamps=False)`, this allows `src.run_values(inc_timestamps=False)`.

cc @philsmt


philsmt:
Thanks, I've been filtering this manually a couple of times already.

How about the API in `DataCollection` using `SourceData`?

philsmt:
Fair enough!

LGTM

PR #582, opened 2024-12-19
Support more magic methods

philsmt:
By chance I stumbled upon weird exceptions when you try to use the `in` operator on a `DataCollection`. Due to interactions in the [Python data model](https://docs.python.org/3/reference/datamodel.html#object.__contains__), the absence of a custom `__contains__` causes it to try treating `DataCollection` as a sequence...

This PR adds proper support for `__contains__` in `DataCollection` via a custom magic method. While at it, I also added support for the OR operator for merging (both binary and infix) and excplicitly disabled direct iteration to avoid the issue above.

Open questions:
- Should `__contains__` allow train ID checks? It is somewhat weird to use, but without it we would not support everything one can pass to `__getitem__`  
- Any other magic methods we could add? What I deliberated and rejected:
  - `__eq__`: somewhat straightforward to add semantically, but probably not useful outside of tests
  - `__len__`: which axis to use, trains or sources/keys?
  - `__iter__`: Rather iterate explicitly via the provided methods, and avoid weird data model interactions

### DIFF
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -290,6 +290,18 @@ class DataCollection:
 
         raise TypeError("Expected data[source], data[source, key] or data[train_selection]")
 
+    def __contains__(self, item):
+        if (
+            isinstance(item, tuple) and
+            len(item) == 2 and
+            all(isinstance(e, str) for e in item)
+        ):
+            return item[1] in self._get_source_data(item[0])
+        elif isinstance(item, str):
+            return item in self.all_sources
+
+        return False
+
     def _ipython_key_completions_(self):
         return list(self.all_sources)
 
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -662,6 +662,12 @@ class DataCollection:
             is_single_run=same_run(self, *others),
         )
 
+    def __or__(self, other):
+        return self.union(other)
+
+    def __ior__(self, other):
+        return self.union(other)
+
     def _parse_aliases(self, alias_defs):
         """Parse alias definitions into alias dictionaries."""
 
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -302,6 +302,8 @@ class DataCollection:
 
         return False
 
+    __iter__ = None  # Disable iteration
+
     def _ipython_key_completions_(self):
         return list(self.all_sources)
 
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -296,7 +296,8 @@ class DataCollection:
             len(item) == 2 and
             all(isinstance(e, str) for e in item)
         ):
-            return item[1] in self._get_source_data(item[0])
+            return item[0] in self.all_sources and \
+                item[1] in self._get_source_data(item[0])
         elif isinstance(item, str):
             return item in self.all_sources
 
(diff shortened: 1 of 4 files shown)


extra_data/reader.py
```diff
@@ -290,6 +290,20 @@ def __getitem__(self, item):
 
         raise TypeError("Expected data[source], data[source, key] or data[train_selection]")
 
+    def __contains__(self, item):
+        if (
+            isinstance(item, tuple) and
+            len(item) == 2 and
+            all(isinstance(e, str) for e in item)
+        ):
+            return item[1] in self._get_source_data(item[0])
```
takluyver, line 299: We still need to check the source name (`item[0]`) as well in this case - at present, it would raise `SourceNameError` instead of returning False, I think.

takluyver:
> Should `__contains__` allow train ID checks? 

Good question. I lean towards no, at least for now. It's easy enough to write `tid in foo.train_ids`.

> Any other magic methods we could add?

Let's not for the moment. I could see an argument for making `DataCollection` behave more like a mapping of source names to `SourceData` objects, i.e. supporting iter & len over source names.

extra_data/reader.py
philsmt: Good catch, also forgot testing it.

extra_data/tests/test_reader_mockdata.py
```diff
@@ -1050,3 +1070,14 @@ def test_proc_legacy_sources(mock_modern_spb_proc_run):
...
 
     assert run.select(det_mod0).all_sources == {det_mod0}
+
+
+def test_datacollection_contains(mock_fxe_control_data):
+    run = H5File(mock_fxe_control_data)
+
+    assert 'FXE_XAD_GEC/CAM/CAMERA:daqOutput' in run
+    assert 'MY/LITTLE/PONY' not in run
+    assert 'MY/LITTLE/PONY', 'actualPosition' not in run
```
takluyver, line 1080:
```suggestion
    assert ('MY/LITTLE/PONY', 'actualPosition') not in run
```

It may have been doing the right thing before, but parentheses are easier than thinking about operator precedence. :wink:

extra_data/tests/test_reader_mockdata.py
```diff
@@ -1050,3 +1070,14 @@ def test_proc_legacy_sources(mock_modern_spb_proc_run):
...
+    assert 'SPB_XTD9_XGM/DOOCS/MAIN', 'beamPosition.ixPos' in run
+    assert 'SPB_XTD9/XGM/DOOCS/MAIN', '42' not in run
```
takluyver, line 1082:
```suggestion
    assert ('SPB_XTD9_XGM/DOOCS/MAIN', 'beamPosition.ixPos') in run
    assert ('SPB_XTD9/XGM/DOOCS/MAIN', '42') not in run
```

extra_data/tests/test_reader_mockdata.py
philsmt: Indeed I checked back then, but good point for a test.

Essentially the worded operators come after all sign operators: https://docs.python.org/3/reference/expressions.html#operator-precedence

PR #592, opened 2025-02-04
Add units to xarray's attributes

tmichela:


extra_data/keydata.py
```diff
@@ -435,12 +435,14 @@ def xarray(self, extra_dims=None, roi=(), name=None):
 
         # Train ID index
         coords = {'trainId': self.train_id_coordinates()}
+        # xarray attributes
+        attrs = {'units': self.units}
```
takluyver, line 439: If the key in the file doesn't have units, do we want to have a `'units': None` attr, or just leave it out entirely? I lean towards leaving it out, but not strongly.

```suggestion
        attrs = {}
        if (units := self.units):
            attrs['units'] = units
```
```

tmichela: yes, this is better, thanks

PR #599, opened 2025-02-19
Allow passing a detector name to extra-data-make-virtual-cxi

JamesWrigley:
This is necessary for runs with multiple multi-module detectors.

### DIFF
--- a/extra_data/cli/make_virtual_cxi.py
+++ b/extra_data/cli/make_virtual_cxi.py
@@ -61,6 +61,11 @@ def main(argv=None):
              ' incorrect train IDs in the data, but may mean less data is'
              ' available.'
     )
+    ap.add_argument(
+        '--detector-name', default=None,
+        help='Name of the detector to make a VDS for. This can be omitted if'
+             ' only one multi-module detector is being used.'
+    )
     args = ap.parse_args(argv)
     out_file = args.output
     fill_values = None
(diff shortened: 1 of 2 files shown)


extra_data/cli/make_virtual_cxi.py
```diff
@@ -100,7 +105,7 @@ def main(argv=None):
     inc_suspect = not args.exc_suspect_trains
     run = RunDirectory(run_dir, inc_suspect_trains=inc_suspect)
 
-    _, det_class = identify_multimod_detectors(run, single=True)
+    _, det_class = identify_multimod_detectors(run, detector_name=args.detector_name, single=True)
```
turkot, line 108: Hi James,
I'm afraid this would not work in the current implementation, because this argument, 'detector_name', is not currently used by the 'identify_multimod_detectors' function at all :sweat:

extra_data/components.py
```diff
@@ -2013,7 +2013,10 @@ def identify_multimod_detectors(
     res = set()
     for cls in clses:
         for name in cls._find_detector_names(data):
-            res.add((name, cls))
+            if (detector_name is None
+                or detector_name in name
+            ):
+                res.add((name, cls))
```
turkot, line 2019: I suggest this change to 'identify_multimod_detectors' function to solve my previous comment.

extra_data/components.py
```diff
@@ -2013,7 +2013,10 @@ def identify_multimod_detectors(
     res = set()
     for cls in clses:
         for name in cls._find_detector_names(data):
-            res.add((name, cls))
+            if (detector_name is None
+                or detector_name in name
```
takluyver, line 2017: I think if you're specifying a name, it should be exact; this matches how  the`detector_name` is [used when instatiating a component](https://github.com/European-XFEL/EXtra-data/blob/c6307c5fad46aa8a751461745085f81adba3237d/extra_data/components.py#L269).

```suggestion
                or detector_name == name
```

turkot: Thanks Thomas, looks good to me. James, would you like to commit this?

PR #600, opened 2025-02-24
Exclude length 0 chunks from VDSs in virtual overview files

takluyver:
This should fix/workaround an issue getting attributes when a virtual overview file is used, which resulted in an error coming from HDF5 (`RuntimeError: Can't get selection bounds (can't get selection bounds)`).


JamesWrigley (approved):
LGTM! Tested in production :sunglasses:

takluyver:
Thanks!

PR #601, opened 2025-02-24
No longer throw NoDataError when there is no source_file_path

philsmt:
https://github.com/European-XFEL/EXtra-data/pull/399 introduced properties for filename traits like `aggregator`. To transparently work with voview files, it uses the `.source_file_paths` property rather than `.files`.

Back then I introduced the behaviour of throwing `NoDataError` when `.source_file_paths` is empty, which at the time I assumed to only happen when no trains are selected. Unfortunately what happens in reality is that the voview file generally contains true datasets for emtpy sources, and `.source_file_paths` actually contains the voview file, causing a return of `None` for the `.aggregator` property.

Therefore I see no benefit of throwing this exception, as the worst case is to return `None` (which happens anyway already) or more likely we're able to actually return a proper result (if no voview file is used, `.files` still contains at least one usable path).

Additionally, this fixes an issue I had with corrupted files in p2654 containing only zero train IDs.

### DIFF
--- a/extra_data/sourcedata.py
+++ b/extra_data/sourcedata.py
@@ -106,7 +106,7 @@ class SourceData:
             # This property is an empty list if no trains are selected.
             sample_path = first_kd.source_file_paths[0]
         except IndexError:
-            raise NoDataError(self.source) from None
+            sample_path = first_kd.files[0].filename
 
         return FileAccess(sample_path)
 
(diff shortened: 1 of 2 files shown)


philsmt:
As discussed with Thomas, ideally we obtain some way to identify a voview file without finding any of its mapping. This would allow to properly fail in all cases currently returning `None`, while still returning a usable result as often as possible.

takluyver:
LGTM

philsmt:
Thanks for review!

I'll keep this pending for a day or two until we resolved https://github.com/European-XFEL/EXtra-data/pull/602.

philsmt:
Merging this now to handle the legacy case better.

PR #604, opened 2025-03-11
Fix getting pulse/cell_id_coordinates for XTDF data with pulse selection

takluyver:
I discovered that e.g. `lpd['image.data'].select_pulses(np.s_[:5]).pulse_id_coordinates()` ignores the pulse selection and gives you all the pulse IDs in the underlying data. This is a bug; these methods are meant to work as labels for what `.ndarray()` would return.


philsmt:
LGTM

takluyver:
Thanks!

PR #606, opened 2025-03-13
Fix default value of `skip` when calling validate from cli

RobertRosca:
I noticed the following when calling `extra-data-validate .`

```python
Checking run directory: .

  File "/gpfs/exfel/sw/software/mambaforge/22.11/envs/202501/bin/extra-data-validate", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/gpfs/exfel/sw/software/mambaforge/22.11/envs/202501/lib/python3.11/site-packages/extra_data/validation.py", line 401, in main
    validator = RunValidator(path, term_progress=True, skip_checks=args.skip)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfs/exfel/sw/software/mambaforge/22.11/envs/202501/lib/python3.11/site-packages/extra_data/validation.py", line 266, in __init__
    self.skip_checks = set(skip_checks)
                       ^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not iterable
```

Seems to happen due to `args.skip` being `None` if not used.

PR sets the default value to an empty list.

### DIFF
--- a/extra_data/validation.py
+++ b/extra_data/validation.py
@@ -382,14 +382,14 @@ def main(argv=None):
     ap.add_argument('path', help="HDF5 file or run directory of HDF5 files.")
     ap.add_argument('-l', '--list', action=ListAction, nargs=0,
                     help="List available checks (options for --skip)")
-    ap.add_argument('--skip', action='append',
+    ap.add_argument('--skip', action='append', default=[],
                     help="Skip a named check (may be used several times)")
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, good catch, LGTM

PR #607, opened 2025-03-17
Index groups according to metadata

egorsobolev:
`EXtra-data` reconstructs the index groups of instrument sources by stored index structure. But index groups are also listed in `METADATA/dataSources` and this list is ignored.

In contrast, `EXtra-data` respects the list of sources in METADATA and does not give an access to sources are not listed there even if they really stored in files.

Usually there is no difference between the stored index groups and exposed in METADATA. But recently it changed. In the legacy sources may be less indexed groups then in new sources. But legacy source is a link to the new source and structurally they are identical.

We have an example when we reduce the original data and the `exdf-tools` change the metadata by adding index groups of new source to the legacy source, although they were not there in initial files.

see discussion in https://git.xfel.eu/dataAnalysis/exdf-tools/-/merge_requests/25

This reads the index groups from metadata.

@takluyver @philsmt

### DIFF
--- a/extra_data/file_access.py
+++ b/extra_data/file_access.py
@@ -515,8 +515,15 @@ class FileAccess(metaclass=MetaFileAccess):
             root = 'CONTROL'
         elif source in self.instrument_sources:
             root = 'INSTRUMENT'
-            if index_group not in self._index_groups[source]:
-                raise ValueError(f'{index_group} not an index group of `{source}`')
+            index_groups = self._index_groups[source]
+            if index_group:
+                if index_group not in index_groups:
+                    raise ValueError(f'{index_group} not an index group of `{source}`')
+            else:
+                if not index_groups:
+                    raise ValueError(f'`{source}` has no keys')
+                index_group = next(iter(index_groups))
+                prefix = index_group + '.'
         else:
             raise SourceNameError(source)
 
--- a/extra_data/file_access.py
+++ b/extra_data/file_access.py
@@ -147,6 +147,7 @@ class FileAccess(metaclass=MetaFileAccess):
             self.control_sources = _cache_info['control_sources']
             self.instrument_sources = _cache_info['instrument_sources']
             self.legacy_sources = _cache_info['legacy_sources']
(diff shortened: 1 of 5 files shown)


egorsobolev:
Can I ignore this coverage check?

egorsobolev:
When it was just the reading issue, I didn't have any objection, althouth it is not consistent with metadata in the files (and formally brakes the format).

But when we copy files with `exdf-tools` and supposed to reproduce the structure we really change it. It happen because the `extra-data` provides data structure differed from the initial in the files. 

I don't consider this as critical issue. But it really upset me. Looks disgusting although doesn't have much impact.

I see that we naturally brake the statement of 'just an alias' from the beginning. It is disaster if we would demand that all new code support back all deprecated names. I think this is how it happens with `litpixel counter`.

philsmt:
Looking at the files you linked in the `exdf-tools` MR, I understand the issue now. Frankly from my POV we wrote out broken files. A legacy sources was always meant to be an alias to an existing source, not some subset we happen to care about at the time. I didn't spot it back then, and/or the `DataFile` API should've caught it.

In any case, no change will be needed in EXtra-data.

egorsobolev:
But let's leave aesthetics aside for a moment.

This MR doesn't implements anything specific to the legacy sources and doesn't change itself the 'just an alias' concept.

This change the handling of indexing groups, which are already exists. Now we discover them from the h5 tree, but this MR reads them from metadata instead.

That is a bit of design change which seems to me positive.

The first point we make design consistent, because we read data sources from metadata. Why then we ignore the index groups, which exposed in the same metadata as data sources itself?

The second we make the files and internal representation consistent. And don't tolerate misinformation. If someone improperly fill metadata he cannot properly read.

And in general it gives more control in different cases including the issue that we started with.

egorsobolev:
I don't think we have written broken files. They work fine and even implicitly strongly support an idea of deprecation, althought it wasn't intendet initially. Just change a point of view and will see nice advantages.

takluyver:
Following discussion on Zulip yesterday, we agreed not to change how this currently works.

PR #612, opened 2025-03-25
Feat: trainid selection by bool

tmichela:
Allows selecting train IDs of a DataCollection, Source/KeyData with a boolean (x)array.
This lets you do things like:

```
xgm = run.alias['xgm-photonFlux'].xarray()
sel = run[xgm > 1000]
```

### DIFF
--- a/extra_data/read_machinery.py
+++ b/extra_data/read_machinery.py
@@ -152,6 +152,8 @@ def select_train_ids(train_ids, sel):
         # Select a list of trains by index in this collection
         return sorted(np.asarray(train_ids)[sel])
     elif isinstance_no_import(sel, 'xarray', 'DataArray'):
+        if sel.dtype != bool:
+            raise TypeError(f"xarray selector must be boolean, got {sel.dtype}")
         if 'trainId' not in sel.coords:
             raise ValueError("xarray selector must have a 'trainId' coordinate")
 
--- a/extra_data/read_machinery.py
+++ b/extra_data/read_machinery.py
@@ -152,8 +152,6 @@ def select_train_ids(train_ids, sel):
         # Select a list of trains by index in this collection
         return sorted(np.asarray(train_ids)[sel])
     elif isinstance_no_import(sel, 'xarray', 'DataArray'):
-        if sel.dtype != bool:
-            raise TypeError(f"xarray selector must be boolean, got {sel.dtype}")
         if 'trainId' not in sel.coords:
             raise ValueError("xarray selector must have a 'trainId' coordinate")
 
--- a/extra_data/read_machinery.py
+++ b/extra_data/read_machinery.py
@@ -15,6 +15,8 @@ from warnings import warn
 
 import numpy as np
 
+from .utils import isinstance_no_import
+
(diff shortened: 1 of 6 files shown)


takluyver:
Neat, LGTM!

extra_data/read_machinery.py
```diff
@@ -140,9 +142,23 @@ def select_train_ids(train_ids, sel):
...
+        # Handle boolean numpy array
+        if len(sel) != len(train_ids):
+            raise ValueError(f"Boolean selector length ({len(sel)}) must match "
+                             f"train_ids length ({len(train_ids)})")
+        return [tid for tid, keep in zip(train_ids, sel) if keep]
     elif isinstance(sel, (list, np.ndarray)):
         # Select a list of trains by index in this collection
         return sorted(np.asarray(train_ids)[sel])
+    elif isinstance_no_import(sel, 'xarray', 'DataArray'):
+        if sel.dtype != bool:
```
philsmt, line 155: I wonder whether we need this restriction? We only care about a `trainId` coordinate, and it also could enable an implicit alignment pattern like:

```python
traces = run.alias['some-adc-channel'].xarray()
aligned_run = run[traces]
```

takluyver: That's a bit too implicit for my taste - I'd prefer to force people to type something a bit longer that makes the intention clearer.

tmichela: Oh, I just saw Thomas' reply, while making the change. I'm fine with any, so I let you guys decide. 🙃 

Though my original thinking was akin to Thomas' and not be too permissive to avoid hidden mistake. I was also weighting only allowing 1D xarrays, thinking higher orders could maybe lead to accidental mistakes.

takluyver: I was assuming that `run[traces]` was just a shorthand for `run[by_id[traces.trainId.values]]`, but from the tests I see you've interpreted it to mean 'select trains where the value is non-zero'. I think that supports my theory that the intended meaning is not clear. :wink: 

I'd be fine with allowing 1D integer xarrays doing the same thing as numpy arrays, so you can do `run[by_id[traces.trainId]]` without needing `.values`. But let's not try to get too clever.

philsmt: > I was assuming that run[traces] was just a shorthand for run[by_id[traces.trainId.values]], but from the tests I see you've interpreted it to mean 'select trains where the value is non-zero'

Actually I intended the former and did not check the implementation much. In that sense, the `bool` check is relevant insofar as the value is ignored for non-`bool` types.

To me the intent of this expression seems as clear as passing a boolean array (only for labeled data, mind you), but I won't fight for it.

takluyver: If you encounter the more general 'align to xarray using trainId dimension' behaviour first, though, the special case for boolean xarrays is liable to be surprising.

tmichela: aha, I did not think it this way. Then I would agree with Thomas and prefer the explicit selection for this case.

takluyver:
LGTM

tmichela:
thanks

PR #616, opened 2025-04-01
Exclude object arrays for streaming karabo-bridge data

takluyver:
VectorString properties such as `availableScenes` are now saved, but streaming them goes wrong. Fixing it properly is a bit messy, but for now we don't really need these to be streamed, so it's simplest just to drop them.

Tested with:

```shell
python -m extra_data.cli.serve_run --include 'FXE_XTD9_MONO/MDL/ACCM_PITCH' 7517 35
```

### DIFF
--- a/extra_data/export.py
+++ b/extra_data/export.py
@@ -121,6 +121,8 @@ def _drop_object_arrays(data):
         for k, v in d.items():
             if isinstance(v, np.ndarray) and v.dtype == np.dtype(object):
                 to_delete.append(k)
+                if k.endswith('.value') and (tsk := k[:-6] + '.timestamp') in d:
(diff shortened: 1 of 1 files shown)


philsmt:
This is fine in principle, though I wonder whether we shouldn't fix it [here](https://github.com/European-XFEL/karabo-bridge-py/blob/master/karabo_bridge/serializer.py#L130)?

takluyver:
I don't have a strong preference on that. But the 'send this data' code probably shouldn't silently part of the data passed in, so it would be a question of either raising an error (in which case a fix is still needed in EXtra-data) or converting arrays of strings to something that can be sent, like Python lists.

PR #619, opened 2025-04-10
Add support for custom error messages in SourceNameError

JamesWrigley:
Now that DAMNIT can recognize `SourceNameError` I'd like to try using it a bit more extensively in EXtra to be more consistent. Currently the components tend to return either `RuntimeError` or `ValueError` when they can't find a source, which leads to this situation:
image

Focusing on the bottom half of the runs, the attenuator boxes are gray because their variables just do `run.alias[...]` and that raises a `SourceNameError`, but the Pulses/Scan type/XGM boxes are orange because they raise different exceptions even though it's the same underlying problem.

Components often have specific, more informative error messages so this adds an option to override the default message, and then I think we can start using it in EXtra.

### DIFF
--- a/extra_data/exceptions.py
+++ b/extra_data/exceptions.py
@@ -6,11 +6,12 @@ class FileStructureError(Exception):
 
 
 class SourceNameError(KeyError):
-    def __init__(self, source):
+    def __init__(self, source=None, custom_message=None):
         self.source = source
+        self.custom_message = custom_message
 
     def __str__(self):
-        return (
+        return self.custom_message if self.custom_message is not None else (
             "This data has no source named {!r}.\n"
             "See data.all_sources for available sources.".format(self.source)
         )


philsmt:
So you'd like to raise `SourceNameError` for all the various cases during automatic source detection in components?

JamesWrigley:
Yeah exactly.

extra_data/exceptions.py
```diff
@@ -6,11 +6,13 @@ class FileStructureError(Exception):
 
 
 class SourceNameError(KeyError):
-    def __init__(self, source):
+    # If `custom_message=True`, then `source` is treated as the full error message
+    def __init__(self, source, custom_message=False):
```
philsmt, line 10: This falls under nitpicking, but I'd argue `None` is the canonical way to indicate absence of value in Python

philsmt:
Fair enough. I'd use `None` instead of `False` as the default value, but LGTM.

extra_data/exceptions.py
JamesWrigley: Hmm, would you then call it as e.g. `SourceNameError(None, "foo")`?

philsmt: Oh, not quite, sorry! I was thinking of:

```python
class SourceNameError(KeyError):
    def __init__(self, source, custom_message=None):
        ...
```

and check for identity with `None` further below.

Concerning how to call it, actually I would be fine to make both arguments optional and use keywords for new invocation. That would keep old code compatible.

extra_data/exceptions.py
```diff
@@ -6,11 +6,13 @@ class FileStructureError(Exception):
...
         self.source = source
+        self.custom_message = custom_message
 
     def __str__(self):
-        return (
+        return self.source if self.custom_message else (
```
philsmt, line 15: Shouldn't this be `return self.custom_message if self.custom_message else ...` 🤔

extra_data/exceptions.py
JamesWrigley: Alrighty, I think I like the keyword argument style more so I went with that in 2996de4.

extra_data/exceptions.py
JamesWrigley: Nope, `custom_message` was originally just a boolean flag so in this case `self.source` would be the custom message. Possibly better named as `use_custom_message` :sweat_smile:

philsmt: Oh... so `self.source` will be any text and `custom_message` is just a flag to display it instead?

I wonder if I would rename it then to avoid confusion. I doubt it was ever called with argument names, so it should still continue to work the old way. Sorry, the above comment concerning calling by keyword is then meant in a different way.

PR #620, opened 2025-04-14
Add reference labels for each release in changelog

takluyver:
Headings starting with a number appear to get an automatic `#id1` type target for links. This adds stable IDs based on the version number.

The alternative would be to make headings like 'Version 1.21'. :shrug:


JamesWrigley (approved):
LGTM! Thanks :heart:

PR #621, opened 2025-04-14
Make zlib_into an optional dependency

takluyver:
So that EXtra-data can still easily be installed on other platforms, but we don't have to build `zlib_into` for those platforms.


tmichela:
LGTM!

takluyver:
Thanks!

PR #623, opened 2025-04-29
Support string keys in KeyData.as_single_value

philsmt:
For a short while now, `CONTROL` data also contains string properties. The current `KeyData.ndarray()` reads this as an array of `bytes` objects right now. This breaks the current implementation of `KeyData.as_single_value()` as it was written with the assumption of numerical types.

This PR is a minimal attempt of fixing this by introducing a new method `KeyData.as_unique_value()` and having `KeyData.as_single_value()` to behave identically for string valued keys, essentially forcing `np.unique` reduction behaviour for such types.

Some tidbits or questions related to this:
* The HDF dataset does not easily convert to an actual *string* type like `U<N>` beforehand with `h5py` throwing `no conversation path` exceptions even if I know the right one, I got only `data.astype(str)` to work after the fact.
* `np.unique()` only works on `object` arrays if the `axis` keyword argument is omitted
* Assuming all object arrays are strings, which seems to hold at the moment, should we transparently convert such keys in all read methods to, well, `str`?

### DIFF
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -340,7 +340,13 @@ class KeyData:
         the first value encountered. By default, 'median' is used.
 
         If within tolerances, the reduced value is returned.
+
+        For non-numerical keys like strings, the method behaves
+        identically to .as_unique_value() and ignores all arguments.
         """
+
+        if not np.issubdtype(self.dtype, np.number):
+            return self.as_unique_value()
 
         data = self.ndarray()
 
@@ -365,6 +371,26 @@ class KeyData:
                              f'(absolute: {adev:.3g}, relative: {rdev:.3g})')
 
         return value
+
+    def as_unique_value(self):
+        """Retrieve a single value if unique.
+
+        Returns the single unique of this key for all trains provided
+        there is a single one.
+        """
+
+        data = self.ndarray()
+
+        if len(data) == 0:
+            raise NoDataError(self.source, self.key)
+
+        axis = 0 if np.issubdtype(self.dtype, np.number) else None
+        unique_values = np.unique(data, axis=axis)
+
+        if len(unique_values) > 1:
+            raise ValueError(f'data values are not unique: {unique_values}')
+
+        return unique_values[0]
 
     # Getting data as different kinds of array: -------------------------------
 


takluyver:
> The HDF dataset does not easily convert to an actual string type like U<N>

Numpy's `U<N>` dtypes represent a fixed number of unicode code points, which there's no equivalent of in HDF5, so h5py deliberately [doesn't try to support that](https://docs.h5py.org/en/stable/strings.html#what-about-numpy-s-u-type). Fixed-length strings in HDF5 are fixed in terms of bytes, like Numpy's `S<N>` dtypes.

I believe our HDF5 files use variable-length strings, which h5py currently converts to Python `bytes` objects on read, or `str` with `dset.asstr()`. There's work underway to support a new variable-length string dtype added in Numpy 2.

> np.unique() only works on object arrays if the axis keyword argument is omitted

That's a shame. We do already have some vectors of strings that can get saved, although the ones like availableScenes are perhaps unlikely to be used for analysis.

> Assuming all object arrays are strings, which seems to hold at the moment, should we transparently convert such keys in all read methods to, well, str?

That would technically be a breaking change, but I think we can get away with it since strings in control data are still pretty new. The code reading RUN already returns str objects:

https://github.com/European-XFEL/EXtra-data/blob/d7fcb1528877774e44b9164a359f0b0a37200656/extra_data/sourcedata.py#L441-L443

We can check if an object array comes from HDF5 variable-length strings. h5py attaches some metadata to the object dtypes it constructs, so you can use [h5py.check_string_dtype(dt)](https://docs.h5py.org/en/stable/special.html#h5py.check_string_dtype), either on the dataset's dtype or that from the returned array.

extra_data/keydata.py
```diff
@@ -366,6 +372,26 @@ def as_single_value(self, rtol=1e-5, atol=0.0, reduce_by='median'):
...
+        there is a single one.
+        """
+
+        data = self.ndarray()
+
+        if len(data) == 0:
+            raise NoDataError(self.source, self.key)
+
+        axis = 0 if np.issubdtype(self.dtype, np.number) else None
+        unique_values = np.unique(data, axis=axis)
```
takluyver, line 388: I'm OK with not supporting multidimensional object data for now, but I think we should throw an error (`NotImplementedError`?) rather than just doing something different with it.

philsmt:
Splitting it into two was an attempt to easier explain why `as_single_value` behaves so differently. As a side effect, it offer the uniqueness functionality for other data types. `as_single_value(reduce_func='unique')` essentially doesn't work for anything that actually changes value.

But I do not mind at all folding it back into `as_single_value` for the string case.

philsmt:
@takluyver Before I add tests, something like this?

extra_data/keydata.py
```diff
@@ -340,13 +340,38 @@ def as_single_value(self, rtol=1e-5, atol=0.0, reduce_by='median'):
...
+
+            unique_values = np.unique(data, axis=None)
+
+            if len(unique_values) > 1:
+                raise ValueError(f'str values are not unique: {unique_values}')
+
+            value = unique_values[0]
+
+            if isinstance(value, bytes):
+                value = value.decode('utf-8', 'surrogateescape')
```
takluyver, line 368: I'd lean towards skipping this so that `kd.as_single_value()` gives us the same type as `kd.ndarray()[0]`. I don't feel strongly about it, though - I accept people are probably more likely to want strings, I just like consistency.

philsmt: Ah, as mentioned above I would quite like to expose strings as strings. As there's been no type from the DAQ before that could end up in `bytes` in files, it seems safe to convert. I agree with your point of consistency, so how about also converting it in the other getters?

takluyver: Right, sorry, I'd forgotten our discussion about that. That seems reasonable. If we can get numpy 2.x in our environments we could use the new numpy variable length string type, which is meant to be more efficient. But the new 202502 environment still has Numpy 1.x, so maybe that's overambitious. Hopefully string data should all be relatively small anyway.

philsmt: Ah right, that is tricky for arrays. So we just go with fixed-length strings conversion? What should `KeyData.dtype` report in such a case? 🤔

takluyver: I'd guess it might be more efficient to do the conversion in one go rather than making an array of fixed width strings as an intermediate. This is what h5py does:

https://github.com/h5py/h5py/blob/6f0d8f0d411eacf8ac6da90302996e2fe9a95619/h5py/_hl/dataset.py#L292-L294

I believe most of the performance advantage of numpy over Python operations goes out the window once object arrays are involved. I haven't really tested this, though.

extra_data/keydata.py
```diff
@@ -396,6 +416,14 @@ def ndarray(self, roi=(), out=None):
             )
             dest_cursor = dest_chunk_end
 
+        if out.dtype.hasobject:
+            # Can current only occur for string properties, convert from
+            # object array of bytes to to object array of strings.
+            out = np.array(
+                [x.decode('utf8', 'surrogateescape') for x in out.flat],
+                dtype=object
+            ).reshape(out.shape)
```
takluyver, line 425: This will fail if we ever have a compound dtype containing a string field. But that seems unlikely and it's not trivial to do the right thing in that case, so let's leave it as is and cross that bridge if we come to it.

philsmt: Hu, well spotted. Better safe than sorry, I will check `dtype.fields` for this branch as well. In this weird case we then simply opt out of string treatment.

extra_data/keydata.py
```diff
@@ -396,6 +416,16 @@ def ndarray(self, roi=(), out=None):
             )
             dest_cursor = dest_chunk_end
 
+        if out.dtype.hasobject
```
takluyver, line 419:
```suggestion
        if out.dtype.hasobject:
```

extra_data/keydata.py
```diff
@@ -340,13 +340,33 @@ def as_single_value(self, rtol=1e-5, atol=0.0, reduce_by='median'):
         the first value encountered. By default, 'median' is used.
 
         If within tolerances, the reduced value is returned.
+
+        For non-numerical keys like strings, the method instead always
```
tmichela, line 344:
```suggestion
        For non-numerical data like strings, the method instead always
```

or 

```suggestion
        For non-numerical type like strings, the method instead always
```

extra_data/keydata.py
```diff
@@ -340,13 +340,33 @@ def as_single_value(self, rtol=1e-5, atol=0.0, reduce_by='median'):
...
+            # Handle non-numeric types first.
+
+            if reduce_by is not None:
+                raise TypeError('custom reduce method not supported for '
+                                'non-numeric type')
+
+            unique_values = np.unique(data, axis=None)
+
+            if len(unique_values) > 1:
+                raise ValueError(f'str values are not unique: {unique_values}')
```
tmichela, line 363:
```suggestion
                raise ValueError(f'Values are not unique: {unique_values}')
```

tmichela: The output can potentially be very large if we display all unique values. we should maybe trim the array or not display it at all?

tmichela (commented):
> I would quite like to expose strings as strings. As there's been no type from the DAQ before that could end up in bytes in files, it seems safe to convert.

is this still true in Karabo3?

I think I would prefer to keep the bytes strings, because I know this will break code in few place at HED. But if you both want it I won't object, though we need to communicate this well.

takluyver:
I think we were working on the assumption that no-one would have used saved string data yet because it wasn't stored until quite recently. Do you have an example of how it's used?

philsmt:
Thanks for the example. I would still like to use the LIMP to not carry around this baggage for any significant amount of time. We are generally very cautious in breaking API compatibility, but there need to be times to clean up technical debts.

I was already thinking of a "Brekaing changes at DA from the LIMP" document independently a few days ago, and this is something we should definitely do, coupled with a deadline for such things (end of Summer?).

If there are now strong objections to just follow through with this one:
- Revert to `bytes` being the default and add an optional (default `False`) argument to decode to strings for all getters
- The same as above, but defaulting to `True` 
- To avoid an argument for all getters, make it a static flag on `KeyData` defaulting to either `True` or `False` and corresponding methods to switch it, e.g. `run['...', 'state'].as_pystring()`

philsmt:
Thanks for the review and discussion!

PR #624, opened 2025-04-29
Close newly written virtual overview file before renaming it

takluyver:
This will hopefully prevent a rare race condition where something tries to read a virtual overview file before it's fully written.


philsmt:
LGTM

PR #625, opened 2025-05-09
Optionally include aggregators in lsxfel and run.info()

philsmt:
It's sometimes useful to quickly find in which aggregator a given source has been saved, e.g. for debugging purposes. This has recently become more pressing due to the introduction of load balancing.

This PR adds a new flag to `DataCollection.info()` to include the aggregator of a given source, and a correspondig flag to `lsxfel`.

Calling `SourceData.aggregator` is somewhat expensive, as a key has to be accessed to differentiate between real and virtual datasets. If one were to use `SourceData.files` directly, this is actually free.

### DIFF
--- a/extra_data/lsxfel.py
+++ b/extra_data/lsxfel.py
@@ -11,13 +11,13 @@ from .read_machinery import FilenameInfo
 from .reader import H5File, RunDirectory
 
 
-def describe_file(path, details_for_sources=()):
+def describe_file(path, details_for_sources=(), with_aggregators=False):
     """Describe a single HDF5 data file"""
     basename = os.path.basename(path)
     print(basename, ": Data file")
 
     h5file = H5File(path)
-    h5file.info(details_for_sources)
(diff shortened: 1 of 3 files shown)


takluyver:
One question: we might conceivably one day want an option to show *attributes* such as units in `lsxfel` output. So we should give a brief thought to whether the `-a` short option would be better for that (or something else entirely). That said, we've encountered a need for this first, and it's probably a decent rule of thumb to use the short option for the first thing you want it for.

LGTM

philsmt:
I don't particularly mind another name, attributes seems to be somewhat more user-facing. How about `--aggregators, -g`?

takluyver:
I don't think I can see anything else that `-g` might be - though saying that probably guarantees there will be something in the near future. Go for it: we can't spend too much time anticipating the future.

philsmt:
Thanks, merging with `-g`

PR #626, opened 2025-05-20
Change DataCollection.alias to a static property

philsmt:
I got a complaint that aliases do not support tab completion in Jupyter, even though it actually contains an `_ipython_key_completions_` method. Some testing revealed this is due to it being a `@property`, which IPython apparently checks for explicitly.

Given `AliasIndexer` is actually stateless we could just make it a attribute, but it turned out `@cached_property` works just as well and doesn't require any logic changes.

### DIFF
--- a/docs/reading_files.rst
+++ b/docs/reading_files.rst
@@ -72,7 +72,9 @@ to refer to all data associated with that 0.1 second window.
 
    .. automethod:: info
 
-   .. autoattribute:: alias
+   .. attribute:: alias
+
+      Enables item access via source and key aliases.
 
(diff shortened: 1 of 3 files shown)


takluyver:
Unfortunately autodoc doesn't pick up the docstring from the AliasIndexer class, so we'll need to tweak the docs.

image

philsmt:
Changed the docs to a manual entry, looks like it works.

takluyver:
LGTM, thanks!

philsmt:
Thanks!

PR #628, opened 2025-05-22
Allow substring matches for lsxfel --detail option

takluyver:
This allows e.g. `--detail XGM`, saving a few keystrokes in many common situations.

The previous behaviour is kept if the value looks like a glob pattern (has `*`, `?` or `[`), so if you want to wildcard part of the pattern, you may also need an explicit wildcard at the start and/or end.

### DIFF
--- a/extra_data/cli/lsxfel.py
+++ b/extra_data/cli/lsxfel.py
@@ -7,8 +7,8 @@ import os.path as osp
 import re
 import sys
 
-from .read_machinery import FilenameInfo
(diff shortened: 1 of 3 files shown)


philsmt:
LGTM, seems like a nice QoL improvement. Speaking of saving keystrokes, how about `-d` while we're at it? 😉

takluyver:
Good idea :+1:

philsmt (approved):

PR #629, opened 2025-05-23
Set thresholds for coverage CI statuses

takluyver:
This should reduce PRs being marked as 'failing' based on code coverage.

(Of course we can always choose to ignore the status from codecov and merge anyway, but it's nice to have the status be more informative.)


philsmt:
Frankly I don't have a good measure of what numbers we want to have here, but let's start with these ones. Thanks, LGTM.

PR #633, opened 2025-06-11
Use zlib_into.unshuffle to improve performance on decompression

takluyver:
We have been relying on Numpy to implement the shuffle filter, which rearranges the bytes in the data to improve compression. But because Numpy is quite generic, the rearrangement is not actually that fast. Some specific code can be faster. In my tests, I saw speedups of roughly:

- 79% on unshuffling alone (890 µs -> 180 µs on 65536 4-byte values)
- 58% on decompression & unshuffling (580 µs -> 245 µs on 65536 2-byte values)
- 15% on loading & decompressing AGIPD data (testing with 1500 AGIPD-1M frames: 1.46 s -> 1.25) from GPFS with the default parallel setting (16 threads)

I was hoping for a bigger overall speedup, but I guess we're pushing it more into I/O being the limiting factor again. I was also testing on a shared node because the cluster is busy right now; the results might look different on a dedicated node.

### DIFF
--- a/extra_data/compression.py
+++ b/extra_data/compression.py
@@ -4,7 +4,7 @@ from multiprocessing.pool import ThreadPool
 
 import h5py
 import numpy as np
-from zlib_into import decompress_into
+from zlib_into import decompress_into, unshuffle
 
 
 def filter_ids(dset: h5py.Dataset):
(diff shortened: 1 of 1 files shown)


JamesWrigley (approved):
LGTM!

PR #640, opened 2025-07-07
Add support for generating Jupyterhub links to alias files

JamesWrigley:
This required tracking any alias files used through the `DataCollection` object.

The result is:
image

### DIFF
--- a/extra_data/aliases.py
+++ b/extra_data/aliases.py
@@ -93,8 +93,21 @@ class AliasIndexer:
         if len(source_key_aliases) == 0:
             return "No aliases have been loaded."
 
+        # Print links to the alias files
+        output_lines = []
+        n_files = len(self.file_paths())
+        links = self.jhub_links()
+        if n_files == 1:
+            path = self.file_paths()[0]
+            output_lines.append(f"Alias file: {links.get(path, path)}\n")
+        elif n_files > 1:
+            output_lines.append(f"Alias files:")
+            for i, path in enumerate(self.file_paths()):
+                line_end = "\n" if i == n_files - 1 else ""
+                output_lines.append(f"- {links.get(path, path)}{line_end}")
+
         # Print the aliases
-        output_lines = ["Loaded aliases:"]
+        output_lines.append("Loaded aliases:")
         for source, alias_keys in source_key_aliases.items():
             if len(alias_keys) == 0:
                 # If there are no keys then this is a plain source alias
@@ -166,6 +179,23 @@ class AliasIndexer:
                 res.append(item)
 
         return res
+
+    def file_paths(self):
+        """Return any file paths that were used to add aliases."""
+        return self.data._alias_files
+
+    def jhub_links(self):
+        """Return a dict of alias file paths to clickable Jupyterhub links.
+
+        Note that a link will only be generated if the file path is under ``/gpfs``.
+        """
+        links = { }
+        for p in self.file_paths():
+            path_from_home = str(p).replace("/gpfs", "GPFS")
+            if path_from_home.startswith("GPFS"):
+                links[p] = f"https://max-jhub.desy.de/hub/user-redirect/lab/tree/{path_from_home}"
+
+        return links
 
     def select(self, seln_or_alias, key_glob='*', require_all=False,
                require_any=False):
--- a/extra_data/reader.py
+++ b/extra_data/reader.py
@@ -22,7 +22,7 @@ import tempfile
 import time
 from collections import defaultdict
 from collections.abc import Iterable, Mapping, Sequence
-from itertools import groupby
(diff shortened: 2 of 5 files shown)


extra_data/reader.py
```diff
@@ -629,6 +631,10 @@ def _merge_aliases(self, alias_dicts):
 
         return new_aliases
 
+    def _merge_alias_files(self, *alias_files):
+        all_files = chain.from_iterable(alias_files)
+        return list(set(all_files))
```
takluyver, line 636:
```suggestion
        return sorted(set(all_files))
```

Sets give a random order, which doesn't really matter, but it's always nice to be deterministic if possible.

extra_data/reader.py
```diff
@@ -693,10 +702,11 @@ def is_valid_alias(k, v):
                 alias_dicts.append(alias_def)
             elif isinstance(alias_def, (str, os.PathLike)):
                 # From a file.
+                alias_files.append(alias_def)
```
takluyver, line 705: Do we want to normalise the type here?

```suggestion
                alias_files.append(os.fspath(alias_def))
```

This would convert Path objects to strings.

setup.py
```diff
@@ -84,6 +84,7 @@ def find_version(*parts):
               'nbval',
               'pytest',
               'pytest-cov',
+              'pyyaml',
```
takluyver, line 87: This is already in the regular requirements, so it doesn't need to be here.

```suggestion
```

takluyver (commented):
This broadly looks good, I have a few specific suggestions.

extra_data/reader.py
philsmt: I was about to comment, I would prefer it we always have `str` or `Path` internally (as well as externally via the property), and not return combination of types was passed.

extra_data/aliases.py
```diff
@@ -167,6 +167,15 @@ def _resolve_aliased_selection(self, selection):
 
         return res
 
+    def file_paths(self):
+        """Return any file paths that were used to add aliases."""
+        return self.data._alias_files
+
+    def jhub_links(self):
+        """Return clickable Jupyterhub links to any alias files that were used."""
+        paths_from_home = [str(p).replace("/gpfs", "GPFS") for p in self.file_paths()]
```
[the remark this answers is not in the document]
philsmt: Actually, shouldn't we specifically exclude (and/or print a warning) for anything not under `/gpfs`? Someone else will hardly be able to open a jhub link pointing to someone's home directory beyond `~/GPFS`.

extra_data/reader.py
```diff
@@ -117,6 +118,7 @@ def __init__(
             }
         self._sources_data = sources_data
 
+        self._alias_files = [] if alias_files is None else alias_files
```
philsmt, line 121: It makes sense in the context of this PR, but could you add a comment this is only for tracking purposes? The actual aliases are still carried via the `aliases` dict after all.

philsmt:
Generally good, also got some comments about details.

Interestingly it's a situation I have not yet encountered, but I trust it's relevant enough for you.

extra_data/aliases.py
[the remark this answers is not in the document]
takluyver: I'd assume the links are primarily for the person running the code, and if they can use an alias file in their home directory, they can follow the link for it as well. It's not ideal if they share the notebook with someone else, but using things in your home directory harms that anyway.

extra_data/reader.py
JamesWrigley: Fixed in d92ccd1.

extra_data/tests/conftest.py
```diff
@@ -5,11 +5,19 @@
...
 
 from . import make_examples
 
 
-@pytest.fixture(scope='session', params=['0.5', '1.0', '1.2'])
+# Make Pyyaml write tuples as regular lists, otherwise it adds a tag to the
+# element that makes it impossible to deserialize without a custom constructor.
+def tuple_representer(dumper, data):
+    return dumper.represent_list(list(data))
+yaml.add_representer(tuple, tuple_representer)
```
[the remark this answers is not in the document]
JamesWrigley: Yep that works, fixed in d92ccd1.

setup.py
JamesWrigley: Whoops I missed that, fixed in d92ccd1.

JamesWrigley:
Yeah I added it because people frequently want to add more aliases and have trouble finding the file they're defined in. To that end, I made another change in d92ccd1 to show the links in `repr()` as well to be even more obvious:
image

I also changed the link to use Jupyterlab instead of Ye Olde Notebooks since that's what most people seem to use these days.

takluyver:
Thanks, LGTM!

PR #648, opened 2025-08-05
Fix one_key() not finding a key for CONTROL sources with a selection

philsmt:
`SourceData.one_key(index_group=None)` may be called with `''` for `index_group` in the case of `CONTROL` sources internally, which is meant to be supported throughout the stack. This fails however in the case of a key selection, as this enters a different branch not handling it properly. I suspect this fell through given the lack of test for it specifically.

Oddly enough I found this bug through `exdf-tools` reductions, where one would have assumed it should've come up sooner 🤷

### DIFF
--- a/extra_data/sourcedata.py
+++ b/extra_data/sourcedata.py
@@ -195,7 +195,7 @@ class SourceData:
         any index group.
         """
         if self.sel_keys is not None:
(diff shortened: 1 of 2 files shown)


takluyver:
LGTM

PR #652, opened 2025-08-06
Add support for voview files for FileAccess property detection

philsmt:
A side product of https://github.com/European-XFEL/EXtra-data/pull/651, but no longer used by it.

This adds support for voview files to `FileAccess` when extracting metadata from the file path, and adds some properties that were always part of the regexp, but not exposed.

### DIFF
--- a/extra_data/file_access.py
+++ b/extra_data/file_access.py
@@ -130,14 +130,14 @@ class FileAccess(metaclass=MetaFileAccess):
     # Regular expressions to extract path and filename information for HDF5
     # files saved on the EuXFEL computing infrastructure.
     euxfel_path_pattern = re.compile(
-        # A path may have three different prefixes depending on the storage location.
-        r'\/(gpfs\/exfel\/exp|gpfs\/exfel\/d|pnfs\/xfel.eu\/exfel\/archive\/XFEL)'
+        # A path may have four different prefixes depending on the storage location.
+        r'\/(gpfs\/exfel\/exp|gpfs\/exfel\/d|gpfs\/exfel\/u|pnfs\/xfel.eu\/exfel\/archive\/XFEL)'
 
         # The first prefix above uses the pattern <instrument>/<cycle>/<proposal>/<class>/<run>,
-        # the other two use <class>/<instrument>/<cycle>/<proposal>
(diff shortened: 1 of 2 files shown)


extra_data/file_access.py
```diff
@@ -213,19 +213,19 @@ def _evaluate_path_infos(self):
...
         else:
-            self._path_infos = (None, None, None)
+            self._path_infos = (None, None, None, None)
 
     def _evaluate_filename_infos(self):
         m = self.euxfel_filename_pattern.match(osp.basename(self.filename))
 
         if m:
-            self._filename_infos = (m[1], m[2], int(m[3]))
+            self._filename_infos = (m[1], int(m[2][1:]), m[3], int(m[4]) if m[3] != 'OVERVIEW' else 0)
```
takluyver:
```suggestion
            self._filename_infos = (m[1], int(m[2]), m[3], int(m[4]) if m[3] != 'OVERVIEW' else 0)
```

I think this was to slice off the `R` for run, but that's not in your capturing group, so it slices off the first digit instead, and will go wrong for run numbers over 999.

extra_data/tests/test_file_access.py
```diff
@@ -87,24 +87,31 @@ def test_euxfel_path_infos(mock_sa3_control_data, monkeypatch):
...
+        '/pnfs/xfel.eu/exfel/archive/XFEL/raw/SA3/202301/p001234/r0100/foo.h5',
+        '/gpfs/exfel/exp/SA3/202301/p001234/usr/.extra_data/RAW-R0100-OVERVIEW.h5'
     ]:
         fa = FileAccess(filename, _cache_info=_empty_cache_info)
 
-        assert fa.storage_class == 'raw'
+        assert fa.storage_class in {'raw', 'usr'}
         assert fa.instrument == 'SA3'
         assert fa.cycle == '202301'
+        assert fa.proposal == 1234
```
takluyver:
```suggestion
        assert fa.proposal == 1234
        assert fa.run == 5100
```

extra_data/file_access.py
philsmt: Well caught!

extra_data/tests/test_file_access.py
philsmt: That actually doesn't work, as `run` was only inferred from filename, not path. But we can actually, so I implemented this test by having run be a two-step check.

PR #672, opened 2025-10-20
fix: xarray roi selection

tmichela:
This fixes a few bugs found after #662 was deployed:

For example, with Jungfrau data, shape (1,512,1024) per trainId 
- Ellipsis
  - np.s_[:1,30:40,...]: Returns dim_1 with coords arange(30,40) only. dim_2 has no coords
  - np.s_[...,30:40]: CoordinateValidationError
- Single valued dimensions are omitted from coords/dimension names
  - np.s_[0,30:40,:]: Returns array with dim_0 arange(30,40), which should be dim_1, retaining dim_0=0 as collapsed dimension
  - np.s_[0, [30, 33], :]: CoordinateValidationError: conflicting sizes for dimension 'dim_0'
- Related to single valued dimensions, this no longer works:
  - roi = np.s_[0]: returns TypeError: 'int' object is not iterable
  - np.s_[0, 10:20,30:40] with extra_dims=['a', 'b', 'c']: ValueError: different number of dimensions on data and dims: 3 vs 4

Enhancement:
Provides a new `extra_coords` argument that takes either a dict for user provided coordinates, or a bool to generate default coordinates.

### DIFF
--- a/extra_data/keydata.py
+++ b/extra_data/keydata.py
@@ -37,9 +37,9 @@ def expand_indexing(shape, indexing):
     # expanding Ellipsis
     if any(x is Ellipsis for x in indexing):
         ellipsis_idx = indexing.index(Ellipsis)
-        # Count non-Ellipsis, non-newaxis dimensions
-        n_before = sum(1 for x in indexing[:ellipsis_idx] if x is not None)
-        n_after = sum(1 for x in indexing[ellipsis_idx + 1:] if x is not None)
+        # Count non-Ellipsis
+        n_before = len(indexing[:ellipsis_idx])
+        n_after = len(indexing[ellipsis_idx + 1:])
         n_ellipsis = max(0, ndim - n_before - n_after)
 
         # Replace Ellipsis with appropriate number of colons
@@ -553,7 +553,10 @@ class KeyData:
             Name the array itself. The default is the source and key joined by a
             dot. Ignored for structured data when a dataset is returned.
         extra_coords: bool or dict
-            Add coordinates to the returned DataArray.
+            Add coordinates to the returned DataArray. If roi is used, the
+            coordinates will match the selected region of interest. If a dict is
+            given, it should map dimension names to coordinate arrays. If True,
+            default coordinate arrays will be generated.
         """
         import xarray
 
@@ -562,7 +565,6 @@ class KeyData:
         # Train ID index
         coords = {'trainId': self.train_id_coordinates()}
         dims = ['trainId']
-        drop_dim = []
 
         def _dim_name(idx):
             if extra_dims is not None:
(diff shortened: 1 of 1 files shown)


extra_data/keydata.py
```diff
@@ -10,6 +10,119 @@
...
+
+        elif isinstance(idx, np.ndarray):
+            # Handle integer array
+            arr = idx.copy()
+            # Normalize negative indices
+            arr = np.where(arr < 0, arr + size, arr)
+            result.append(arr)
+
+        else:
+            # Default: all indices
```
takluyver: Do we know what remaining cases this `else` covers? It seems odd to assume anything that's not covered above is equivalent to `slice(None)`.

extra_data/keydata.py
```diff
@@ -467,33 +580,28 @@ def xarray(self, extra_dims=None, roi=(), name=None):
...
+        # Dimension labels after the train dimension
+        _extra_dims = []
+        if roi != ():
+            for idx, coord in enumerate(expand_indexing(self.entry_shape, roi, expand_full=False)):
+                dim = _get_dim_name(idx)
+
+                if coord is None or not isinstance(coord, int):
+                    _extra_dims.append(dim)
+                if coord is not None and not isinstance(coord, int):
+                    coords[dim] = coord
```
takluyver: I thought from your description that you did want to keep the scalar coordinate label when a dimension is indexed with an integer, whereas this discards it. Did I misunderstand?

tmichela: Is it possible to add coordinates to a non existing dimension?

extra_data/keydata.py
```diff
@@ -10,6 +10,119 @@
...
+                   (slice(None),) * n_ellipsis + 
+                   indexing[ellipsis_idx + 1:])
+
+    # Separate newaxis from array dimensions
+    array_indices = []
+    new_axes_indices = []
+    dim_idx = 0
+
+    for position, idx in enumerate(indexing):
+        if idx is None:
```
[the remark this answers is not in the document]
tmichela: I noticed it after implementing this. I though it would not hurt to keep it in case h5py supports this in future, but I can remove if you prefer.

extra_data/keydata.py
takluyver: Yup! Xarray keeps the coordinate when you select a scalar on that dimension, even though it's no longer in `dims`, so  `data.sel(x=10).coords['x']` gives you a 0D array containing the number 10.

tmichela: oh OK, it works but only with a scalar :bulb:

extra_data/keydata.py
```diff
@@ -10,6 +10,93 @@
...
+        indexing = (indexing,)
+
+    ndim = len(shape)
+
+    # expanding Ellipsis
+    if any(x is Ellipsis for x in indexing):
+        ellipsis_idx = indexing.index(Ellipsis)
+        # Count non-Ellipsis, non-newaxis dimensions
+        n_before = sum(1 for x in indexing[:ellipsis_idx] if x is not None)
+        n_after = sum(1 for x in indexing[ellipsis_idx + 1:] if x is not None)
```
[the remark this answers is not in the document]
tmichela: updated in edc75545cb501882ee67630f9dad0d34b7030431

PR #675, opened 2025-10-30
CLI to check whether files are readable

philsmt:
There have been some issues with dCache recently again, which typically results in complaints that `open_run` hangs or is throwing `PermissionError`. The latter is the optimal case, as here dCache cleanly indicates a file is not available on disk, but unfortunately the former case is also quite common with processes hanging forever in uninterruptible sleep. That can cause shells to be unrecoverable without a hard kill, making troubleshooting these issues rather annoying.

This PR is an attempt at a CLI that tries to perform readability check in a robust way without leaving zombies around or locking up shells. It spawns a single process per file to perform a few checks and verifies via a finite timeout whether the process returned. The process is killed in case of timeout, which on the new RHEL seems to not leave `D` state zombies around.

I have experimented a bit with the `spawn` method of `multiprocessing` and directly using `subprocess.run` and decided now for latter. I like `ThreadPool` due to its `imap_unordered` method that works nicely for feedback. A `tqdm` progressbar might also be nice, but I wasn't sure if we want that as dependency...

The checks that are run are

- [dCache only] read locality
- stat the file
- create a `FileAccess` object, which reads the source structure

Example:

```
(extra) schmidtp@max-exfl461:EXtra-data% python3 -m extra_data.cli.check_readable `findxfel 3090 235` --timeout 10
.............................................T....T..T.................
average access time: 2.42508s, max access time: 4.1392s
/gpfs/exfel/exp/SQS/202301/p003090/raw/r0235/RAW-R0235-DA02-S00006.h5          dCache reports on disk but read attempt hangs
/gpfs/exfel/exp/SQS/202301/p003090/raw/r0235/RAW-R0235-DA04-S00001.h5          dCache reports on disk but read attempt hangs
/gpfs/exfel/exp/SQS/202301/p003090/raw/r0235/RAW-R0235-DIGICTRL00-S00001.h5    dCache reports on disk but read attempt hangs
```

### DIFF
--- a/extra_data/cli/check_readable.py
+++ b/extra_data/cli/check_readable.py
@@ -1,0 +1,161 @@
+
+import subprocess
+import sys
+from argparse import ArgumentParser
+from multiprocessing.pool import ThreadPool
+from os import cpu_count
+from pathlib import Path
+from time import monotonic
+
+from extra_data import FileAccess
+from extra_data.locality import get_locality
+
+
+def check_access(path):
+    print('any access hangs', flush=True)
+
+    start = monotonic()
+
+    try:
+        _, locality = get_locality(path)
+    except Exception as e:
+        print(f'locality check raises {e.__class__.__name__}')
+        return 1
+
+    if locality == 'UNAVAILABLE':
+        print('dCache reports unavailable')
+        return 0
+    elif locality == 'NEARLINE':
+        print('dCache reports tape only')
+        return 0
+    elif locality == 'NOT_ON_DCACHE':
+        prefix = 'file on GPFS but '
+    else:
+        prefix = 'dCache reports on disk but '
+
+    print(f'{prefix}stat hangs', flush=True)
+
+    try:
+        path.stat()
+    except Exception as e:
+        print(f'{prefix}stat raises {e.__class__.__name__}')
+        return 1
+    else:
+        print(f'{prefix}read attempt hangs', flush=True)
+
+    try:
+        fa = FileAccess(path)
+    except Exception as e:
+        print(f'{prefix}read attempt raises {e.__class__.__name__}')
+    else:
+        fa.file.close()
+        print(f'readable_{monotonic() - start}')
+
+    return 0
+
+
+def main(argv=None):
+    ap = ArgumentParser(description='Check whether European XFEL data files '
+                                    'are readable')
+
+    ap.add_argument(
+        'input', metavar='INPUT', nargs='+', type=Path,
+        help='folder of input data to check')
+
+    ap.add_argument(
+        '--all', '-a', action='store_true',
+        help='whether to show the result for all files rather than only '
+             'unreadable ones')
+
+    ap.add_argument(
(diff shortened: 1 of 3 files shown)


extra_data/cli/check_readable.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def check_access(path):
+    print('any access hangs', flush=True)
+
+    start = monotonic()
+    locality = get_locality(path)
+
+    if locality == 'UNAVAILABLE':
+        print('dCache reports unavailable', flush=True)
```
JamesWrigley: Is the explicit flush necessary here? Libc's `exit()` guarantees that stdio streams are flushed, which I believe is what Python will call upon a normal exit.

extra_data/cli/check_readable.py
```diff
@@ -0,0 +1,141 @@
...
+
+    # Map of the result for each path.
+    path_states = {}
+
+    def monitor_access_check(path):
+        try:
+            p = subprocess.run(
+                [sys.executable, '-c', check_access_runtime.format(path)],
+                timeout=args.timeout, capture_output=True,
+                env={'OMP_NUM_THREADS': '1'})
```
JamesWrigley: Why set this?

extra_data/cli/check_readable.py
philsmt: Happy to add it for symmetry, but at that point the process actually will exit gracefully 🙃 

The `flush=True` is essentially needed as long as you're still doing calls that may hang.

extra_data/cli/check_readable.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # Map of the result for each path.
+    path_states = {}
+
+    # Bootstrap code to run access check
+    check_access_runtime = '''
+import sys
+from pathlib import Path
+from extra_data.cli.check_readable import check_access
+sys.exit(check_access(Path('{}')))
```
takluyver:
```suggestion
sys.exit(check_access(Path({!r})))
```

extra_data/cli/check_readable.py
```diff
@@ -0,0 +1,150 @@
...
+    '''
+
+    def monitor_access_check(path):
+        try:
+            # OMP_NUM_THREADS=1 minimizes creation of thread pools when
+            # numpy is imported in the subprocess. As the subprocesses
+            # often live less than 1s for successful checks, this can
+            # cause significant load on cluster-sized nodes.
+            p = subprocess.run(
+                [sys.executable, '-c', check_access_runtime.format(path)],
```
takluyver:
```suggestion
                [sys.executable, '-c', check_access_runtime.format(str(path))],
```

I'm nitpicking now, but these two changes together should correctly handle paths with `'` quotes inside them.

JamesWrigley (approved):
LGTM!

philsmt:
Thank you both for review.

I tried to make a test that fails via mocking, but the subprocess part makes this tricky. The best I could think of is unit-testing `check_access` directly.

PR #747, opened 2026-06-10
Miscellaneous minor cleanups

takluyver:
Modernising type hints, removing unused imports and unnecessary character escapes. This cuts down on the warnings I see in Pycharm.

I'll merge this so long as the tests pass, I don't think it's worth review.

### DIFF
--- a/extra_data/file_access.py
+++ b/extra_data/file_access.py
@@ -134,12 +134,12 @@ class FileAccess(metaclass=MetaFileAccess):
     # Paths starting with /gpfs/exfel/exp are either the canonical link
     # on Maxwell or an ONC path.
     euxfel_exp_path_pattern = re.compile(
-        r'\/gpfs\/exfel\/exp\/(\w+)\/(\d{6})\/p(\d{6})\/([a-z]+)\/(?:r(\d{4})|.extra_data)')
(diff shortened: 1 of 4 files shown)


extra_data/file_access.py
```diff
@@ -134,12 +134,12 @@ class FileAccess(metaclass=MetaFileAccess):
     # Paths starting with /gpfs/exfel/exp are either the canonical link
     # on Maxwell or an ONC path.
     euxfel_exp_path_pattern = re.compile(
-        r'\/gpfs\/exfel\/exp\/(\w+)\/(\d{6})\/p(\d{6})\/([a-z]+)\/(?:r(\d{4})|.extra_data)')
```
philsmt, line 137: Are you sure these are not necessary? The site I am regularly using to test regexps claims they are 🤷

takluyver: In Javascript and some other languages, `/` marks regex literals, i.e. `const re = /ab+c/;`. I'd guess this is what the site is referring to. But Python doesn't have regex literals, and `/` has no special meaning in Python style regexes. I've just double checked this interactively. `re.escape("/")` also doesn't add a backslash.

philsmt: Thanks for the clarification!

PR #749, opened 2026-06-22
Write proposal and run numbers to virtual CXI files.

turkot:
Although proposal and run numbers are not part of the CXI format, I find them useful additional information.

### DIFF
--- a/extra_data/write_cxi.py
+++ b/extra_data/write_cxi.py
@@ -240,26 +240,26 @@ class VirtualCXIWriterBase:
         # be read by HDF5 1.10.
         with h5py.File(filename, 'w', libver=('v110', 'v110')) as f:
             f.create_dataset('cxi_version', data=[150])
-            d = f.create_dataset('entry_1/experiment_identifier',
-                                 shape=experiment_ids.shape,
-                                 dtype=h5py.special_dtype(vlen=str))
+            expgrp = f.create_group('entry_1')
+            d = expgrp.create_dataset(
+                'experiment_identifier',
+                shape=experiment_ids.shape,
+                dtype=h5py.special_dtype(vlen=str)
+            )
             d[:] = experiment_ids
 
             # proposal, run, pulseId, trainId, cellId are not part of
             # the CXI standard, but it allows extra data.
             run_metadata = self.detdata.data.run_metadata()
             if 'proposalNumber' in run_metadata and 'runNumber' in run_metadata:
-                f.create_dataset(f'entry_1/proposalNumber',
-                                 data=run_metadata['proposalNumber'])
-                f.create_dataset(f'entry_1/runNumber',
-                                 data=run_metadata['runNumber'])
-            f.create_dataset(f'entry_1/{self.pulse_id_label}', data=pulse_ids)
-            f.create_dataset('entry_1/trainId', data=self.train_ids_perframe)
-            cellids = f.create_virtual_dataset('entry_1/cellId',
-                                               layouts[self.cell_id_label])
+                expgrp.attrs['proposalNumber'] = run_metadata['proposalNumber']
+                expgrp.attrs['runNumber'] = run_metadata['runNumber']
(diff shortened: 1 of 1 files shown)


extra_data/write_cxi.py
```diff
@@ -245,8 +245,14 @@ def write(self, filename, fillvalues=None):
                                  dtype=h5py.special_dtype(vlen=str))
             d[:] = experiment_ids
 
-            # pulseId, trainId, cellId are not part of the CXI standard,
-            # but it allows extra data.
+            # proposal, run, pulseId, trainId, cellId are not part of
+            # the CXI standard, but it allows extra data.
+            run_metadata = self.detdata.data.run_metadata()
+            if 'proposalNumber' in run_metadata and 'runNumber' in run_metadata:
+                f.create_dataset(f'entry_1/proposalNumber',
```
daniloefl: I can see you are just following a previous standard, but is there a reason for `entry_1`?

extra_data/write_cxi.py
```diff
@@ -245,8 +245,14 @@ def write(self, filename, fillvalues=None):
                                  dtype=h5py.special_dtype(vlen=str))
             d[:] = experiment_ids
 
-            # pulseId, trainId, cellId are not part of the CXI standard,
-            # but it allows extra data.
+            # proposal, run, pulseId, trainId, cellId are not part of
+            # the CXI standard, but it allows extra data.
+            run_metadata = self.detdata.data.run_metadata()
+            if 'proposalNumber' in run_metadata and 'runNumber' in run_metadata:
```
daniloefl, line 254: It might be more dependable for software reading this to always have this data, even if it is not in the metadata. One idea could be to always add this information in the file and mark it as unknown somehow. This is really not needed, but I thought it could be a nicer way for the downstream software to recognize this then checking if the data exists.

extra_data/write_cxi.py
```diff
@@ -245,8 +245,14 @@ def write(self, filename, fillvalues=None):
...
+                f.create_dataset(f'entry_1/proposalNumber',
+                                 data=run_metadata['proposalNumber'])
+                f.create_dataset(f'entry_1/runNumber',
```
daniloefl: Finally, why is this a dataset? I would have expected it to be an attribute in HDF5. Or is that not compatible with CXI files for some reason?

daniloefl (approved):
Looks great! Feel free ignore all my comments. They are mostly only ideas for alternative implementations, in case you think it is worth going in those directions.

extra_data/write_cxi.py
turkot: According to CXI format convention, `entry_1` should correspond to the experiment.

extra_data/write_cxi.py
turkot: After general discussion, we've decided to omit this data when not available.

extra_data/write_cxi.py
turkot: Changed to attributes with https://github.com/European-XFEL/EXtra-data/pull/749/commits/f948bfd1d08075fed63b6928706a1fa18d22fa41

extra_data/write_cxi.py
```diff
@@ -240,20 +240,28 @@ def write(self, filename, fillvalues=None):
...
+                expgrp.attrs['runNumber'] = run_metadata['runNumber']
+            expgrp.create_dataset(f'{self.pulse_id_label}', data=pulse_ids)
+            expgrp.create_dataset('trainId', data=self.train_ids_perframe)
+            cellids = expgrp.create_virtual_dataset('cellId', layouts[self.cell_id_label])
             cellids.attrs['axes'] = 'experiment_identifier:module_identifier'
 
-            dgrp = f.create_group('entry_1/instrument_1/detector_1')
+            dgrp = expgrp.create_group('instrument_1/detector_1')
+            # Store detector name as an attribute
+            dgrp.attrs['name'] = self.detdata.detector_name
```
turkot, line 266: I'm adding name of the detector as 'detector_1' attribute.

